PR #3, opened 2022-03-10
example minor issues and possible fixes

AmineChikhaoui:
Hello, first of all I'm really excited about this project, really nice to see declarative package builds coming up to Alpine :-)

I know the project is still in the early phases so just reporting few minor issues I encountered while playing with the example locally and also while at it just pushing some local fixes I did. In case they make sense, otherwise feel free to ignore if there are upcoming changes to fix them anyway :)

* add missing wget, help2man packages to the build environment. Not sure how it worked before though, as the build runs in a chroot?
* use a tmp directory for the build directory
* bind-mount /etc/resolv.conf instead of copy but maybe useless change ?
* run build sandbox in PID namespace, but likely other namespaces could be used (user, mount, ..) to lock down the build sandbox a bit more ? 
* unset PATH otherwise it inherits the host $PATH and source /etc/profile to get the right $PATH in the sandbox. Not sure but there might be other environment variables that need to be unset just to make sure not much gets leaked from the host ?

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -297,6 +297,7 @@ func (ctx *Context) WorkspaceCmd(args ...string) (*exec.Cmd, error) {
 		"--bind", ctx.GuestDir, "/",
 		"--bind", ctx.WorkspaceDir, "/home/build",
 		"--bind", "/etc/resolv.conf", "/etc/resolv.conf",
+		"--unshare-pid",
 		"--proc", "/proc",
 		"--chdir", "/home/build",
 	}
--- a/pkg/build/pipeline.go
+++ b/pkg/build/pipeline.go
@@ -140,7 +140,8 @@ func monitorPipe(pipe io.ReadCloser) {
 func (p *Pipeline) evalRun(ctx *PipelineContext) error {
 	replacer := replacerFromMap(p.With)
 	fragment := replacer.Replace(p.Runs)
-	script := fmt.Sprintf("#!/bin/sh\nset -e\n%s\nexit 0\n", fragment)
+	sys_path := "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+	script := fmt.Sprintf("#!/bin/sh\nset -e\nexport PATH=%s\n%s\nexit 0\n", sys_path, fragment)
 	command := []string{"/bin/sh", "-c", script}
 
 	cmd, err := ctx.Context.WorkspaceCmd(command...)


kaniini:
Why is `--workspace-dir` removed?  This is needed for some cases.

kaniini:
(to answer your question: wget is included in busybox)

pkg/build/package.go
```diff
@@ -114,7 +114,8 @@ func (pc *PackageContext) EmitPackage() error {
 		tarball.WithOverrideUIDGID(0, 0),
 		tarball.WithOverrideUname("root"),
 		tarball.WithOverrideGname("root"),
-		tarball.WithUseChecksums(true),
+		//FIXME: not used anymore ??
```
kaniini, line 117: This feature needs to be enabled, it attaches the *expected* checksum of each file in the tarball as a PAX record.  apk-tools *absolutely* uses this when using APKv2 format packages.

AmineChikhaoui:
> (to answer your question: wget is included in busybox)

Oh right, I think my addition of help2man/wget were before doing `unset PATH; source /etc/profile` so probably the issues I had are related to $PATH being wrong for me. Will remove them then.

pkg/build/package.go
AmineChikhaoui: It's just that the function doesn't exist https://github.com/chainguard-dev/melange/search?q=WithUseChecksums. Maybe it's still in a local diff ?

pkg/build/build.go
```diff
@@ -297,6 +282,8 @@ func (ctx *Context) WorkspaceCmd(args ...string) (*exec.Cmd, error) {
 	baseargs := []string{
 		"--bind", ctx.GuestDir, "/",
 		"--bind", ctx.WorkspaceDir, "/home/build",
+		"--bind", "/etc/resolv.conf", "/etc/resolv.conf",
```
kaniini, line 285: ok

pkg/build/pipeline.go
```diff
@@ -140,7 +140,7 @@ func monitorPipe(pipe io.ReadCloser) {
 func (p *Pipeline) evalRun(ctx *PipelineContext) error {
 	replacer := replacerFromMap(p.With)
 	fragment := replacer.Replace(p.Runs)
-	script := fmt.Sprintf("#!/bin/sh\nset -e\n%s\nexit 0\n", fragment)
+	script := fmt.Sprintf("#!/bin/sh\nset -e\nunset PATH\nsource /etc/profile\n%s\nexit 0\n", fragment)
```
kaniini, line 143: `/etc/profile` is not guaranteed to exist, so we cannot do this.

kaniini (changes requested):

pkg/build/package.go
kaniini: https://github.com/chainguard-dev/apko/blob/main/pkg/tarball/tarball.go#L100-L107

AmineChikhaoui: Oh sorry 🤦🏼

pkg/build/pipeline.go
AmineChikhaoui: perhaps, $PATH should be set to `/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin` initially?
I realize I'm in a somehow unusual setup since my host distro is NixOS, so for instance my $PATH is `/run/wrappers/bin:/home/amine/.nix-profile/bin:/etc/profiles/per-user/amine/bin:/nix/var/nix/profiles/default/bin:/run/current-system/sw/bin` which wouldn't work. But I suspect relying on PATH from the host can bring problems anyway and better set it to what the default is for Alpine ?

pkg/build/build.go
```diff
@@ -79,9 +78,13 @@ type Dependencies struct {
 }
 
 func New(opts ...Option) (*Context, error) {
+	build_dir, err := os.MkdirTemp("", "melange-build-*")
```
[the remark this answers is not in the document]
AmineChikhaoui: Got it 👍🏼

kaniini:
As an aside, this tool provides declarative builds for the APK ecosystem, which is inclusive of Alpine, but not exclusive to Alpine.  Accordingly, we should avoid to hardcode dependency on Alpine-specific behavior, as there are other distributions which use apk, such as Adelie and OpenWrt, which we want to support with this tool.

pkg/build/build.go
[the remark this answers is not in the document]
AmineChikhaoui: Actually not even worth another option as I could have done `--workspace-dir=$(mktemp -d "/tmp/melange-build-XXXX")` already. So I just reverted that change.

AmineChikhaoui:
Alright, now the PR just contains the $PATH change and `--unshare PID`. Any opinions about using more namespaces for the build sandbox btw ? I think all of `--unshare-pid`, `--unshare-user-try` (I vaguely remember some distros disable user namespaces), `--unshare-ipc`, `--unshare-uts` could be used, so I can add them in this PR if that makes sense.

pkg/build/build.go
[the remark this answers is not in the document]
kaniini: Yep, that was the idea.

kaniini (approved):

PR #7, opened 2022-03-11
refactor: bump apko and use multitar api

kruskall:


kruskall:
This has the side effect of chaning the order of the content of multi tar: eg. signature is written last.

kaniini:
> This has the side effect of changing the order of the content of multi tar: eg. signature is written last.

This probably won't work, the APKv2 format needs Signature, then Control, then Data streams.

kruskall:
> > This has the side effect of changing the order of the content of multi tar: eg. signature is written last.
> 
> This probably won't work, the APKv2 format needs Signature, then Control, then Data streams.

Should be fixed.

kaniini:
I updated the apko package, which should fix the multitar E2E.  Can you update apko version and run the tests?

kruskall:
> I updated the apko package, which should fix the multitar E2E. Can you update apko version and run the tests?

So... I think I misunderstood the package format. I'll do a bit of research and update this PR once I have something ready/better.

kaniini:
The APKv2 format is basically a multi-part TAR with each part being its own tar stream.

There is the signature part, and then the control part, and then, for packages, the data part.  Indices only have the first two parts.

PR #12, opened 2022-03-15
extend e2e test with published melange image

kaniini:
apko needs a new entrypoint type to support bare commands for this to be useful, but progress!

### DIFF
--- a/.apko.yaml
+++ b/.apko.yaml
@@ -1,0 +1,14 @@
+contents:
+  repositories:
+    - http://dl-cdn.alpinelinux.org/alpine/edge/main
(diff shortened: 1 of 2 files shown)


.apko.yaml
```diff
@@ -0,0 +1,17 @@
+contents:
+  repositories:
+    - http://dl-cdn.alpinelinux.org/alpine/edge/main
+    - /__w/melange/melange/repo
```
kaniini, line 4: Should add `--repositories-extra` to match apko 0.2's `--keyring-extra` flag.

PR #13, opened 2022-03-23
upgrade to apko 0.2 and use apko 0.2 APIs

kaniini:


puerco (approved):
Nice to see the apko FS at work! :tada:

PR #23, opened 2022-04-06
Docker env for running melange without alpine

amouat:
Created a script based on apko's apko-inception for running melange.

It has a few issues currently:

 - requires a modified version of the apko-inception image with bubblewrap.
 - Should we host the image somewhere? GHCR?
 - As part of a run script, it installs a go module and makes melange. I guess the next step would be to see if this can be done with apko?
 - Currently requires "--priviliged" for bubblewrap; i'll look into the exact caps required.

Anyway, I thought I'd make it available now for people to play with, but I'm interested in what direction people think it should go in.

### DIFF
--- a/hack/run-devenv.sh
+++ b/hack/run-devenv.sh
@@ -16,7 +16,7 @@ function run() {
 }
 
 function docker_run() {
-    docker run --rm -w /melange -v $(pwd):/melange --entrypoint /melange/hack/run-devenv.sh apko-inception:latest run $@
+    docker run --privileged --rm -w /melange -v $(pwd):/melange --entrypoint /melange/hack/run-devenv.sh apko-inception:latest run $@
 }
 
 case "$1" in
--- a/hack/run-devenv.sh
+++ b/hack/run-devenv.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+# Copyright 2022 Chainguard, Inc.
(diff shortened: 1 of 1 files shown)


kaniini:
(is this ready to go in?)

amouat:
I think we should get it going using the official melange image rather than the apko dev environment, if that makes sense.

hack/run-devenv.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
```
cpanato, line 1: maybe to be similar to the other scripts we have we can use

```
#!/usr/bin/env bash
```

just a comment non blocking

amouat: Note it gets called from within a container, so bash won't work. I can't remember if there was a problem `/usr/bin/env sh` or not.

cpanato: ahh so ignore my comment 🤦‍♂️

kaniini: `/bin/sh` is required to exist due to POSIX anyway :p

jdolitsky:
Using this script, the build fails with:
```
2022/07/01 00:14:59 melange (hello/s390x): bwrap: Creating new namespace failed: Invalid argument
Error: failed to build package: unable to run pipeline: exit status 1
```

jdolitsky:
Going to close this one in relation to #76

PR #54, opened 2022-06-02
support copying source artifacts to the build workspace

### LINKED ISSUE #38: Bring back the ability to do hermetic builds
When multiarch support was merged, it cost us the ability to do hermetic builds.  Restore this ability by copying the directory the `.melange.yaml` file is in to the workspace.  I guess we will want something like `.melangeignore` (as Docker has `.dockerignore`).

kaniini:
Implements support for copying source artifacts to the build workspace, thus restoring the ability to do hermetic builds.

Closes #38.


pkg/build/build.go
```diff
@@ -317,6 +352,132 @@ func (ctx *Context) BuildWorkspace(workspaceDir string) error {
...
+	destPath := filepath.Join(dest, src)
+	destDir := filepath.Dir(destPath)
+
+	inF, err := os.Open(basePath)
+	if err != nil {
+		return err
+	}
+	defer inF.Close()
+
+	if err := os.MkdirAll(destDir, 0o755); err != nil {
```
puerco, line 366: Nit, for later:
If we're creating the directory, perhaps we should check access on the destination and also carry the same directory owner too?

pkg/build/build.go
```diff
@@ -26,6 +29,8 @@ import (
 
 	apko_build "chainguard.dev/apko/pkg/build"
 	apko_types "chainguard.dev/apko/pkg/build/types"
+	apkofs "chainguard.dev/apko/pkg/fs"
+	"github.com/zealic/xignore"
```
puerco, line 33: For future reference, the go-git package is actively maintained and it has support for gitignore patterns too:

https://pkg.go.dev/gopkg.in/src-d/go-git.v4/plumbing/format/gitignore

puerco (approved):
Thanks @kaniini ! I left two comments that we can address later on :)

pkg/build/build.go
kaniini: Cool I made an issue for that too.

PR #61, opened 2022-06-17
Try removing LDFLAGS from ko

amouat:
Not essential, but means we can build the image on MacOS easily.


amouat:
Lol, those changes better get squashed 😳

amouat:
Lol, how did that make this worse. I'm starting again if the tests pass :joy:

PR #75, opened 2022-06-29
Updated README.

amouat:
I mainly moved things about and extended the example. Plus some installation instructions (not enough really, I'll extend when we have a development image).

I was considering removing this section:

>"Secure software factories are the evolution of DevOps, allowing a
>user to prove the provenance of all artifacts incorporated
>into a software appliance.  By building and capturing software
>artifacts into packages, DevOps teams can manage their software
>artifacts as if they were any other component of an image."

I like the writing, but it feels like the beginning of a blog rather than a README. Anyway, I've left it for the moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ the form of OCI container images with [apko][apko].
 
 ## Install
 
-Melange has a dependency on [Alpine Linux](https://www.alpinelinux.org/)
+Melange has a dependency on [apk-tools](https://gitlab.alpinelinux.org/alpine/apk-tools).
 tooling. Currently the easiest way to run melange is inside an Alpine VM or
 container. If you're on MacOS, you can use a Lima VM, as [documented for
 apko](https://github.com/chainguard-dev/apko/blob/main/mac/README.md).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,16 +26,67 @@ the form of OCI container images with [apko][apko].
 
    [apko]: https://github.com/chainguard-dev/apko
 
-## How
-
-To build an unsigned APK, use the `melange build` command:
+## Install
+
+Melange has a dependency on [Alpine Linux](https://www.alpinelinux.org/)
```
kaniini, line 31:
```suggestion
Melange has a dependency on [apk-tools](https://gitlab.alpinelinux.org/alpine/apk-tools).
```

README.md
```diff
@@ -15,16 +26,67 @@ the form of OCI container images with [apko][apko].
...
+tooling. Currently the easiest way to run melange is inside an Alpine VM or
```
kaniini, line 32:
```suggestion
Currently the easiest way to run melange is inside an Alpine VM or
```

kaniini (changes requested):
Melange can technically be used on any distribution with apk-tools available, like apko.

PR #78, opened 2022-07-01
build: use a temporary directory for workspace by default

### LINKED ISSUE #66: Clean workspace by default
It can be confusing to run melange twice and get different results, because there are already files in the workspace directory. This will happen whether or not `--empty-workspace` is used.

I think the best option is to have each run in a clean tmp directory unless otherwise specified. Another option would be to wipe the contents of workspace before a new run.

### LINKED ISSUE #73: file name too long
I got the following error for the simple-hello example while trying to build an apk pkg for it:

```shell
$ docker container run --rm -ti --privileged golang:1.18-alpine sh
$ apk add git make bubblewrap libc-dev 
$ git clone https://github.com/chainguard-dev/melange.git
$ make melange
$ mv melange /usr/local/bin
$ cd examples/simple-hello
$ mkdir -p /usr/share/melange/pipelines
$ cp -R ../../pipelines /usr/share/melange
$ melange build melange.yaml 
```

Screen Shot 2022-06-27 at 7 35 33 PM

kaniini:
In the very early days of melange, we defaulted to use the current
directory as the workspace directory.  This had a lot of problems,
and evolved to ./workspace instead.

When copying from the source directory to the workspace was
implemented, this resulted in workspaces being copied over.

Using a temporary directory also ensures that the build begins
from a clean state every time.

Fixes #66.
Fixes #73.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -112,7 +112,6 @@ type Dependencies struct {
 
 func New(opts ...Option) (*Context, error) {
 	ctx := Context{
-		WorkspaceDir:    "./workspace",
 		WorkspaceIgnore: ".melangeignore",
 		PipelineDir:     "/usr/share/melange/pipelines",
 		SourceDir:       ".",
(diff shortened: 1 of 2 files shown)


pkg/build/build.go
```diff
@@ -127,6 +126,19 @@ func New(opts ...Option) (*Context, error) {
 		}
 	}
 
+	// If no workspace directory is explicitly requested, create a
+	// temporary directory for it.  Otherwise, ensure we are in a
+	// subdir for this specific build context.
+	if ctx.WorkspaceDir != "" {
+		ctx.WorkspaceDir = filepath.Join(ctx.WorkspaceDir, ctx.Arch.ToAPK())
+	} else {
+		tmpdir, err := os.MkdirTemp("", "melange-workspace-*")
```
cpanato, line 135: should we clean this temp directory at the end of the process? or it is ok to keep it?

kaniini: It gets cleaned up later on.  If we do a `defer` here, it will get cleaned up too early.

cpanato: yep, not here.

ah if that cleans up later that is cool

PR #85, opened 2022-07-06
build: ensure that parallel builds dont race

kaniini:


pkg/cli/build.go
```diff
@@ -105,17 +105,26 @@ func BuildCmd(ctx context.Context, archs []apko_types.Architecture, base_opts ..
...
 
-			bc, err := build.New(opts...)
-			if err != nil {
-				return err
-			}
+		bcs = append(bcs, bc)
+	}
 
+	for _, bc := range bcs {
+		errg.Go(func() error {
```
priyawadhwa, line 127: Think this needs a 

```go
bc := bc
```

kaniini: yeah it does

priyawadhwa (approved):

PR #90, opened 2022-07-11
Add --template flag to melange build

priyawadhwa:
I'm thinking this will make it easier to build and release multiple versions of the same package  (e.g. nginx:1.20.3, nginx:1.20). 

My plan is that in CI, we can use the Github actions [matrix strategy](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#using-a-matrix-strategy) to run the melange/apko jobs for different versions of the same package. The specific version can passed in via the flag, and will be applied to `melange.yaml` in code: 

So, melange.yaml could look like this:

```
package:
  name: nginx
  version: {{ .Version }}
```

and we'd pass in 
```
melange build --template '{"Version": "1.20.3"}'
```

Signed-off-by: Priya Wadhwa <priya@chainguard.dev>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.18
 
 require (
 	chainguard.dev/apko v0.4.1-0.20220701165036-c5450771a18e
+	github.com/google/go-cmp v0.5.8
 	github.com/psanford/memfs v0.0.0-20210214183328-a001468d78ef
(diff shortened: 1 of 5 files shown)


kaniini (approved):

kaniini:
this is super cool, i like how it composes similar to kubectl

go ahead and merge it!

PR #92, opened 2022-07-12
Fix bug in template, clashing with pipeline substitutions

priyawadhwa:
the pipeline provides some vars which we replace in code:

```
"${{package.name}}"
"${{package.version}}"
"${{package.epoch}}"
"${{targets.destdir}}"
```
Any config file with these is panicking out w/ template bc it doesn't know how to parse them. Unfortunately go templates don't allow ignoring certain values (TIL a proposal was denied https://github.com/golang/go/issues/31147) 

There are a couple options:
1. change the delimiter from `{{ }}` to `[[ ]]` so go templates can differentiate.
2. do an ugly string substitution before applying the go template and then reset it

I started with (1) but figured that would be really inconvenient for users so switched to (2)

Signed-off-by: Priya Wadhwa <priya@chainguard.dev>

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -27,6 +27,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strconv"
+	"strings"
 	"text/template"
 	"time"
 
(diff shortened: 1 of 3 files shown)


amouat:
Yeah, I see what you mean in the nginx build.

I'm not sure what to suggest either. There's probably something to do with quoting, but that might get complex. Using a different delimiter is wise anyway as melange is intended to work well with actions.

amouat (approved):

amouat:
Hmm, this is a very GitHub action specific fix. We might want to think about a more general solution?

PR #99, opened 2022-07-27
add support for overlaid /bin/sh

kaniini:
This can be used to speed up QEMU builds by overlaying a native `/bin/sh`.  As the overlay is strictly for /bin/sh, it has no noticeable effects on the build environment itself.

The overlay `/bin/sh` must be staticly linked, as it is running inside the build environment.

For example, it is possible to do: `melange build --overlay-binsh /bin/busybox.static` on Alpine to speed up QEMU builds.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -527,6 +527,40 @@ func (ctx *Context) matchesIgnorePattern(path string) bool {
 	return false
 }
 
+func (ctx *Context) OverlayBinSh() error {
+	if ctx.BinShOverlay == "" {
+		return nil
+	}
+
+	targetPath := filepath.Join(ctx.GuestDir, "bin", "sh")
+
+	inF, err := os.Open(ctx.BinShOverlay)
+	if err != nil {
+		return fmt.Errorf("copying overlay /bin/sh: %w", err)
+	}
+	defer inF.Close()
+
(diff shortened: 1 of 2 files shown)


jdolitsky (approved):

amouat:
Interesting. What sort of builds does it speed up? I guess stuff that's doing a shell calls to copy files or similar?

I'm wondering if this could break things but I can't think of anything.

pkg/build/build.go
```diff
@@ -322,6 +323,15 @@ func WithDependencyLog(logFile string) Option {
 	}
 }
 
+// WithBinShOverlay sets a filename to copy from when installing /bin/sh
+// into a build environment.
+func WithBinShOverlay(binShOverlay string) Option {
+	return func(ctx *Context) error {
+		ctx.BinShOverlay = binShOverlay
```
amouat, line 330: If this is a symlink does it still work?

amouat (approved):
Looks good, I'm interested in how much difference it makes.

If you've not already tested, it might be worth checking it all works if the host shell is a symlink.

pkg/build/build.go
kaniini: It does, because the data will get copied.

kaniini:
> What sort of builds does it speed up?

Anything involving autoconf.

PR #119, opened 2022-09-23
Adding an origin header to the generated PKGINFO

bureado:
By adding an `origin` header to the control template that will eventually make it into the `PKGINFO`, packages generated by `melange` and installed in Alpine-based systems such as Wolfi will register an `origin` attribute in the APK index, which is an attribute that at least one popular open source scanner relies on for vulnerability assessment.

I have tested this patch by building `melange` and running `melange build  --arch x86_64 examples/minimal.yaml`.

The resulting `apk` has an `origin` attribute in `.PKGINFO`, and installs cleanly with `apk add` on a `wolfi-base:latest` image.

The final state of `/lib/apk/db` shows the registered attribute, and the target scanner (Trivy 0.32.0) serializes and outputs the `origin` field correctly.


pkg/build/package.go
```diff
@@ -97,6 +97,7 @@ pkgname = {{.PackageName}}
 pkgver = {{.Origin.Version}}-r{{.Origin.Epoch}}
 arch = {{.Arch}}
 size = {{.InstalledSize}}
+origin = {{.PackageName}}
```
kaniini, line 100: This should be `.Origin.Name` to match `abuild`:

```suggestion
origin = {{.Origin.Name}}
```

kaniini (changes requested):

bureado:
Thanks @kaniini, I will rebuild and test again in the next few minutes and force push with a gitsign'ed commit.

PR #120, opened 2022-09-23
fetch: support expected-sha512

imjasonh:
This came up while trying to build ICU, which only provides sha512s of its source, e.g., https://github.com/unicode-org/icu/releases/tag/release-71-1

cc @jdolitsky

### DIFF
--- a/pipelines/fetch.yaml
+++ b/pipelines/fetch.yaml
@@ -18,10 +18,12 @@ inputs:
   expected-sha256:
     description: |
       The expected SHA256 of the downloaded artifact.
+    default: ""
 
   expected-sha512:
     description: |
       The expected SHA512 of the downloaded artifact.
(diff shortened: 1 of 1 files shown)


imjasonh:
@jdolitsky tried this out and with the params unset we got a bad bash substitution trying to check for them in the script. This now does a terrible hack where the default value is `"NOTSET"` as a sentinel value, but we should either fix fetch.yaml (or maybe melange?) to handle unset params so we don't have to do this.

jdolitsky:
Yes, I think desired behavior here would be that if an input is missing that is not required, it should result in empty string substitution

jdolitsky:
so the issue was that the hyphen character (`-`) was not supported by the regex. Opened #121 in place of this PR. Closing

PR #121, opened 2022-09-23
fetch: support expected-sha512

jdolitsky:
This came up while trying to build ICU, which only provides sha512s of its source, e.g., https://github.com/unicode-org/icu/releases/tag/release-71-1

Replaces #120

cc @imjasonh


dlorenc (approved):
Approving on behalf of @kaniini !

PR #126, opened 2022-09-26
Add dir input for autoconf pipelines

jdolitsky:
On using the fetch pipeline, some tarballs contain nested locations for `configure` or `Makefile`. This allows using the autoconf pipelines in a different directory.

### DIFF
--- a/pipelines/autoconf/make-install.yaml
+++ b/pipelines/autoconf/make-install.yaml
@@ -1,8 +1,16 @@
 name: Run autoconf make install
 
+inputs:
+  dir:
+    description: |
+      The directory containing the Makefile.
+    default: .
+
 needs:
   packages:
     - make
 
 pipeline:
-  - runs: make install DESTDIR="${{targets.destdir}}"
+  - runs: |
+      cd ${{inputs.dir}}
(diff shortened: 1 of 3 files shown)


pipelines/autoconf/make-install.yaml
```diff
@@ -1,8 +1,16 @@
...
+
 needs:
   packages:
     - make
 
 pipeline:
-  - runs: make install DESTDIR="${{targets.destdir}}"
+  - runs: |
+      cd ${{inputs.dir}}
+      make install DESTDIR="${{targets.destdir}}"
```
kaniini, line 16: use `make -C ${{inputs.dir}}`.

pipelines/autoconf/make.yaml
```diff
@@ -1,8 +1,16 @@
...
+
 needs:
   packages:
     - make
 
 pipeline:
-  - runs: make -j$(nproc)
+  - runs: |
+      cd ${{inputs.dir}}
+      make -j$(nproc)
```
kaniini, line 16: Likewise.

kaniini (changes requested):

PR #143, opened 2022-10-21
fetch: check a cache directory for files matching expected SHAs

imjasonh:
This adds a `--cache-dir` flag (default `"/var/cache/melange"`), which is bound into the build environment at `/var/cache/melange` (constant), and updates the fetch.yaml pipeline to look for files matching the `expected-sha{256,512}` in that directory.

The intention is that builds could pre-fetch external dependencies into
the cache directory, then execute builds without hitting the network.

If --cache-dir doesn't exist, it's not bound and fetch will always wget.

Also adds CI checks to build all examples.

Fixes https://github.com/chainguard-dev/melange/issues/144

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -5,7 +5,39 @@ on:
   pull_request:
 
 jobs:
-  build:
+  examples:
+    name: build examples
+    runs-on: ubuntu-latest
+    container:
+      image: alpine:latest
+      options: |
+        --cap-add NET_ADMIN --cap-add SYS_ADMIN --security-opt seccomp=unconfined --security-opt apparmor:unconfined
+
+    strategy:
+      matrix:
+        example:
+        - gnu-hello.yaml
+        - mbedtls.yaml
+        - minimal.yaml
+        - sshfs.yaml
+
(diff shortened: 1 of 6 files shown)


imjasonh:
This passes tests, but it doesn't actually work. `/var/cache/melange` is never actually populated inside the build environment, so we never get a cache hit. I'm not sure what I'm missing, I'll take another look in the morning.

imjasonh:
This should work now. I can try to add an e2e test that populates and exercises the cache, but it may be tough to prove it doesn't hit the network in GitHub Actions.

kaniini:
> This should work now. I can try to add an e2e test that populates and exercises the cache, but it may be tough to prove it doesn't hit the network in GitHub Actions.

We can ask bubblewrap to unshare the network namespace.

imjasonh:
> > This should work now. I can try to add an e2e test that populates and exercises the cache, but it may be tough to prove it doesn't hit the network in GitHub Actions.
> 
> We can ask bubblewrap to unshare the network namespace.

Ooh! A new flag for melange that passes `--unshare-net` to bwrap?

kaniini (approved):

PR #145, opened 2022-10-21
add flag to disable networking in build environment

imjasonh:
Intended to be used alongside https://github.com/chainguard-dev/melange/pull/143 where dependencies are pre-fetched.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -139,6 +139,7 @@ type Context struct {
 	DependencyLog     string
 	BinShOverlay      string
 	ignorePatterns    []*xignore.Pattern
+	DisableNetwork    bool
 }
 
 type Dependencies struct {
@@ -372,6 +373,12 @@ func WithBinShOverlay(binShOverlay string) Option {
 		ctx.BinShOverlay = binShOverlay
 		return nil
 	}
+}
+
+// WithDisableNetwork disables networking in the build environment.
+var WithDisableNetwork Option = func(ctx *Context) error {
+	ctx.DisableNetwork = true
+	return nil
 }
 
 // Load the configuration data from the build context configuration file.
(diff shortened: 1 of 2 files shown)


imjasonh:
Worth noting that this will still `apk add` any of the pipeline's dependencies, which might undermine some assumptions about how the flag works.

Should `--disable-network` also skip the `apk add` steps, and assume they're present in the apk cache location? 🤔

kaniini:
I would rather have the melange.yaml file declare the desired capabilities, rather than have a CLI option for this.  We can sync on it if you want.

imjasonh:
> I would rather have the melange.yaml file declare the desired capabilities, rather than have a CLI option for this. We can sync on it if you want.

I like that idea too, especially as we have more capabilities than just networking.

But even if we had those we should have some way to invoke `melange` explicitly without those capabilities, and rely on pre-caching the things `fetch` steps require. `fetch` is sort of an odd case in that way, it "requires" networking except when it doesn't. I'd like to think of a more holistic and generalized way to express that.

In the meantime, solving this isn't a huge blocker -- there's plenty to do to improve the build system, and even without disabling networking explicitly, I think we can attest we're building hermetically if we pre-cache. We can also disable networking at the Pod/K8s layer, instead of asking `melange` to do it, until we figure out the best way to do that.

kaniini:
What about something like:

**pipelines/fetch.yaml**:

```yaml
wants:
  packages:
    - wget
  capabilities:
    - net

...
```

**package.yaml**:

```yaml
package: ...

wants:
  capabilities:
    - net

pipeline:
  - wants:
       capabilities:
          - something-else
     runs: ...
```

And then having a `--deny-capabilities=net` commandline option?

imjasonh:
> What about something like:

That sounds good. The current direction we're going (not too late to change though) would be that `fetch` would still _run_ without the capability, we'd just depend on its cache lookup + digest checking logic to avoid actually calling `wget`.

If that's messy we can investigate some kind of `--replace-pipeline=fetch=fetch-from-cache` flag that swaps the pipeline step definition at call-time, for one that doesn't require the capability.

kaniini:
Of note, by reusing the `wants` struct here, we could have pipeline elements declare their package requirements at the specific pipeline element requiring it, which could be interesting.

PR #146, opened 2022-10-23
add publish command

imjasonh:
This adds a `publish` command, e.g., `melange publish gs://my-bucket/path/to/stuff`

For Wolfi builds running in GKE (or EKS, or another cloud-aligned Kubernetes) it's useful to have an ambient-auth-aware method for publishing built packages to same-cloud-aligned blob storage. Doing this with `gsutil` or `aws s3` adds another set of tools we have to maintain. Using `gocloud` lets us add support for other cloud-storages without installing any new tools.

If the feeling is that this doesn't belong in melange I can make it part of [`dag`](https://github.com/wolfi-dev/dag) instead, or a separate tool there, or make it an optional arg on `build`.


kaniini:
I think this should be part of `dag` instead, but I'm open to syncing on this if you think otherwise.

imjasonh:
> I think this should be part of `dag` instead, but I'm open to syncing on this if you think otherwise.

That sounds good. I can have `dag` be responsible for pre-fetching all the remote deps before starting the build. I'll close this and move this logic over there.

PR #164, opened 2022-11-08
Embed built-in pipelines

### LINKED ISSUE #162: Embed built-in pipelines
Melange's built-in pipelines (e.g., `fetch`, `patch`, etc.) are defined [here](https://github.com/chainguard-dev/melange/tree/main/pipelines), and copied into `/usr/share/melange/pipelines` when you run `make install`, or available in that location when you use the Melange container image.

To make it easier to use Melange without considering shared state on a filesystem, and to avoid hacks where one user updates `/usr/share/melange/pipelines/fetch.yaml` to mine crypto or whatever, we should embed these pipeline definitions at build-time using [`go:embed`](https://pkg.go.dev/embed).

We might want to have a release or two where both the shared dir and compiled-in embedded definitions are considered to avoid breaking, then remove support for `/usr/share/melange/pipelines` in a subsequent release.

imjasonh:
Fixes #162 

With this change, if a pipeline definition can't be found in the user's specified `--pipeline-dir`, or `/usr/share/melange/pipelines`, then we'll also look for a pipeline definition embedded in the binary at build-time.

The pipelines installed into `/usr/share/melange/pipelines` should be the same as those we embed, and in a future release we should stop installing into that shared location.

This moves the builtin pipeline definitions from `pipelines/` to `pkg/build/pipelines/` in our source tree, since `go:embed` requires that the embedded files live in a directory at or below the location where they're embedded from. Symlinks don't seem to satisfy in this case. 😕 


luhring (approved):
Awesome! 👏

PR #167, opened 2022-11-09
Add GNU and Rust triplet support to Melange

kaniini:
Building on the work in apko, we expose the triplets to Melange.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -868,3 +868,26 @@ func (ctx *Context) WorkspaceCmd(args ...string) (*exec.Cmd, error) {
 
 	return cmd, nil
 }
+
+// BuildFlavor determines if a build context uses glibc or musl, it returns
+// "gnu" for GNU systems, and "musl" for musl systems.
+func (ctx *Context) BuildFlavor() string {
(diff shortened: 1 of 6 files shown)


pkg/build/build.go
```diff
@@ -866,3 +868,26 @@ func (ctx *Context) WorkspaceCmd(args ...string) (*exec.Cmd, error) {
 
 	return cmd, nil
 }
+
+// BuildFlavor determines if a build context uses glibc or musl, it returns
+// "gnu" for GNU systems, and "musl" for musl systems.
```
imjasonh, line 873: Not blocking: Might be worth a hacky Go enum for this eventually, if there are only a few strings we expect this to ever be.

pkg/build/pipeline.go
```diff
@@ -339,3 +340,6 @@ func (p *Pipeline) ApplyNeeds(ctx *PipelineContext) error {
 
 	return nil
 }
+
+//go:embed pipelines/*
+var f embed.FS
```
imjasonh, line 345: Was this moved on purpose?

imjasonh (approved):

pkg/build/pipeline.go
kaniini: Yes, it was causing my editor to mishighlight due to the `/*`

PR #173, opened 2022-11-17
Fix input check that triggers tar extraction on fetch pipeline

erikaheidi:
While trying to use the fetch pipeline to download `composer.phar`, I was getting some errors related to trying to unpack the file, even when setting the `extract` input to `false`. Then I found this comparison at the end of the pipeline missing a `=`.


erikaheidi:
Ohh I see that's not the issue, I got confused by the shell script comparison operators. Sorry, closing!

Issue #161, opened 2022-11-07
Consider alternatives to document-wide templating


imjasonh:
Melange supports [templating using Go's template syntax](https://github.com/chainguard-dev/melange#build-file-templating), which I think is really good. I was surprised however to find that templating was applied _before_ parsing a melange.yaml file as YAML.

On the plus side, this means Melange supports meta-programming like [this](https://github.com/wolfi-dev/os/blob/e9cd19b6975a1975011ef716597be0e96ab16975/glibc.yaml#L228), which creates a subpackage per locale/language-name combo.

This also means that it's possible in theory to template anywhere in a Melange config, including keys:

```
package:
  name: description
  ${{package.name}}: lolwut
```

The downside of this is that a melange.yaml file on disk cannot be guaranteed to be parseable as YAML, which would let users leverage existing YAML tooling ([yq](https://mikefarah.gitbook.io/yq), [yamlfmt](https://github.com/google/yamlfmt), schema validators, other [YAML templating engines](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/) maybe, etc.), without having this built in to or otherwise supported by Melange.

I wonder if we should consider reducing the scope of the templating engine, and only have templates applied on (certain?) config values, and not document-wide.

The locale example above could be supported by relying on a shared pipeline (hard in this case because it's iteratively generating `subpackages` and not just pipeline steps), or a separate YAML-generation step that lives in wolfi-dev/os that gets checked in CI.

This would guarantee that Melange YAMLs are valid YAMLs, and let us leverage existing YAML tooling instead of having to invest it in it Melange itself.

If there's a broader thinking behind Melange's templating being done this way, that's fine, and that context would be helpful too -- it's totally possible that such powerful templates are more valuable than having parseable YAML.

I thought I'd bring this up pre-0.2 in case we decide to change course, since post-0.2 it becomes harder to put the genie back in the bottle.

kaniini:
I think we should keep it as it is, because of the metaprogramming capabilities enabled by it.

imjasonh:
> I think we should keep it as it is, because of the metaprogramming capabilities enabled by it.

That's completely reasonable. Can you give me an idea of what kinds of metaprogramming you have in mind, so I can understand how it outweighs the YAML benefits? I'm definitely willing to be convinced.

kaniini:
The locales are an example.  It is also used already in this way inside chainguard images, such as to build multiple versions of nginx from a single melange yaml file.

[...]

PR #175, opened 2022-11-17
Support range for subpackages; drop templates

imjasonh:

This expands on an idea @luhring had in #161 to rely on YAML-defined key/value pairs instead of Go templates.

It's not yet a fully complete idea, just a starting point to discuss how we might like this to work.

As an example:

```
package:
  name: hello
  version: world

pipeline:
- name: hello
  runs: world

data:
  - name: animals
    items:
      dogs: loyal
      cats: angry
      turtles: slow

subpackages:
  - range: animals
    name: ${{range.key}}
    pipeline:
    - runs: ${{range.key}} are ${{range.value}}
```

When a `subpackage` specifies a `range`, key/value pairs for that range must be specified in `data`, where each item in the k/v pair gets expanded in the `subpackages` array before proceeding to the next one.

Multiple named `data`s can be provided, and the order according to `subpackages` is retained.

`range`s are only (currently) supported in `subpackages`, since that's the use case described in #161 and wolfi's glibc locales use case.

With the change as it stands, Go templates are dropped, along with the `--template` flag, but we could consider putting that back, and (I would suggest) only applying templates to YAML string values, and not the whole document.

### DIFF
--- a/pkg/build/build_test.go
+++ b/pkg/build/build_test.go
@@ -76,22 +76,15 @@ pipeline:
 data:
   - name: ninja-turtles
     items:
-    - key: Michelangelo
-      value: orange
-    - key: Raphael
-      value: red
-    - key: Leonardo
-      value: blue
-    - key: Donatello
-      value: purple
+      Michelangelo: orange
+      Raphael: red
+      Leonardo: blue
+      Donatello: purple
   - name: animals
     items:
-    - key: dogs
-      value: loyal
-    - key: cats
-      value: angry
-    - key: turtles
-      value: slow
+      dogs: loyal
+      cats: angry
+      turtles: slow
 
 subpackages:
   - range: animals
@@ -105,19 +98,29 @@ subpackages:
 `
 
 	expected := []Subpackage{{
-		Name: "dogs",
-		Pipeline: []Pipeline{{
-			Runs: "dogs are loyal",
-		}},
-	}, {
 		Name: "cats",
 		Pipeline: []Pipeline{{
 			Runs: "cats are angry",
 		}},
 	}, {
+		Name: "dogs",
+		Pipeline: []Pipeline{{
+			Runs: "dogs are loyal",
+		}},
+	}, {
 		Name: "turtles",
 		Pipeline: []Pipeline{{
 			Runs: "turtles are slow",
+		}},
+	}, {
+		Name: "Donatello",
+		Pipeline: []Pipeline{{
+			Runs: "Donatello's color is purple",
+		}},
+	}, {
+		Name: "Leonardo",
+		Pipeline: []Pipeline{{
+			Runs: "Leonardo's color is blue",
 		}},
 	}, {
 		Name: "Michelangelo",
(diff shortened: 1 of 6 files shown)


pkg/build/build_test.go
```diff
@@ -96,72 +49,108 @@ func TestLoadConfiguration(t *testing.T) {
...
+    - key: Donatello
+      value: purple
+  - name: animals
+    items:
+    - key: dogs
+      value: loyal
+    - key: cats
+      value: angry
+    - key: turtles
+      value: slow
```
luhring, line 94: Did this end up being better/easier than a `map[string]string` approach? E.g. where the YAML would look like

```yaml
items:
  dogs: loyal
  cats: angry
  turtles: slow
```

imjasonh: I think we can't use a map here, because the ordering won't be preserved, and @kaniini indicated that ordering might matter. Maybe that's only true between explicitly defined subpackages, and if it doesn't hold for matrix-generated subpackage defs then we can go back to maps for this, and just sort the generated subpackages.

I do think that `key: value` is much easier to read and write than `- key: key; value: value` on multiple lines, so it'd be great if we could get back to a map.

luhring: We might be able to implement a custom unmarshaler for this part of the data structure. Where for people writing YAML it looks like it'd be `map[string[string`, but really on the Go side, it's something else that preserves order (like `[]struct{key, value string}` or something).

But also, I think what's already there is totally good enough, too 👍

kaniini: i would rather do the unmarshaller myself.

pkg/cli/build.go
```diff
@@ -110,8 +108,7 @@ func Build() *cobra.Command {
 	cmd.Flags().BoolVar(&useProot, "use-proot", false, "whether to use proot for fakeroot")
 	cmd.Flags().BoolVar(&emptyWorkspace, "empty-workspace", false, "whether the build workspace should be empty")
 	cmd.Flags().StringVar(&outDir, "out-dir", filepath.Join(cwd, "packages"), "directory where packages will be output")
-	cmd.Flags().StringVar(&template, "template", "", "template to apply to melange config (optional)")
-	cmd.Flags().StringVar(&dependencyLog, "dependency-log", "", "log dependencies to a specified file")
+	go cmd.Flags().StringVar(&dependencyLog, "dependency-log", "", "log dependencies to a specified file")
```
jdolitsky, line 111: mistake here?

jdolitsky (changes requested):

pkg/cli/build.go
imjasonh: Sure is! Good catch. 👍

pkg/build/build_test.go
imjasonh: This is done now. The result is that the order specified in the YAML map is not kept, and items are just sorted by key for stability, which matches how maps are YAML-marshaled anyway.

jdolitsky (approved):

imjasonh:
FWIW, this is the diff for wolfi-dev/os's glibc.yaml to no longer use Go templates: https://github.com/wolfi-dev/os/compare/main...imjasonh:wolfi-os:glibc-range

kaniini (approved):
solid work!

PR #180, opened 2022-11-22
Go pipeline

puerco:
This PR adds a new built-in pipeline to melange run go builds.

It supports a few options to pass to the compiler like tags, ldflags
and support to relocate the binary using a prefix and target directory.

Example usage:

```yaml
pipeline:
  - uses: fetch
    with:
      uri: https://github.com/puerco/hello/archive/refs/tags/v0.0.1.tar.gz
      expected-sha512: 8b6a997ef52711cfe29ede1510cd4b2.....
  - uses: go/build
    with:
      package: main.go
      output: hello
```

A simple example is included.

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@chainguard.dev>

### DIFF
--- a/examples/go-hello.yaml
+++ b/examples/go-hello.yaml
@@ -1,0 +1,28 @@
+package:
+  name: hello-go
+  version: v0.0.1
+  epoch: 0
+  description: "simple hello go project"
+  target-architecture:
+    - all
+
+environment:
+  contents:
+    repositories:
+      - https://packages.wolfi.dev/os
+
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://github.com/puerco/hello/archive/refs/tags/${{package.version}}.tar.gz
+      expected-sha512: 8b6a997ef52711cfe29ede1510cd4b2181e2397eb89e0114feb6b6829dc01e400154e8021fe8be58ae8cad1c97152721a61c3ab88f68009d1bb7b2263aac6a18
+  - uses: go/build
+    with:
+      packages: main.go
+      output: hello
+      
+      # Example to pass tags to compiler:
+      # tags: hello,bye
+
+      # Example to pass ldflags to compiler 
+      # ldflags: -buildid= sigs.k8s.io/release-utils/version.gitCommit=406d901bef12a92957f47f8a539d6bc6e6e13ca5
--- a/pkg/build/pipelines/go/build.yaml
+++ b/pkg/build/pipelines/go/build.yaml
@@ -1,0 +1,56 @@
+name: Run a build using the go compiler
+
+needs:
+  packages:
+    - go
+    - busybox
+    - ca-certificates-bundle
+
+inputs:
+  packages:
+    description: |
+      List of space-separated packages to compile. Files con also be specified.
+      This value is passed as an argument to go build.
+    required: true
+
+  tags:
+    description: |
+      A comma-separated list of build tags to pass to the go compiler
+  
+  output:
+    description: |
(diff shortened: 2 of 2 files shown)


tuananh:
Nice. Once it's merged, we can use this of many packages in wolfi-dev/os

pkg/build/pipelines/go/build.yaml
```diff
@@ -0,0 +1,55 @@
...
+needs:
+  packages:
+    - go
+    - busybox
+    - ca-certificates-bundle
+
+inputs:
+  package:
+    description: |
+      List of packages to compile
```
luhring, line 12: A few thoughts about user clarity...

1. If it's a list of packages, maybe the input name should be "packages"?
2. Should we clarify that this list should be _space-separated_ (I'm assuming this based on how we're passing this input to `go build` below.)
3. Should we mention in the description that files are also acceptable (I noticed the example has `package: main.go`, for instance.)

luhring (commented):
Great step forward, nice work! A few thoughts for your consideration.

pkg/build/pipelines/go/build.yaml
puerco: Yeah sounds good to me, one of the principles I'm trying to follow is to keep the language and terms the underlying compiler uses (go in this case). So I'm just relaying words :)

puerco:
Upudated the descriptions to capture @luhring 's feedback.

examples/go-hello.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - https://packages.wolfi.dev/os
+
+pipeline:
+  - uses: fetch
+    with:
+      uri: https://github.com/puerco/hello/archive/refs/tags/${{package.version}}.tar.gz
+      expected-sha512: 8b6a997ef52711cfe29ede1510cd4b2181e2397eb89e0114feb6b6829dc01e400154e8021fe8be58ae8cad1c97152721a61c3ab88f68009d1bb7b2263aac6a18
+  - uses: go/build
+    with:
+      package: main.go
```
luhring, line 21:
```suggestion
      packages: main.go
```

pkg/build/pipelines/go/build.yaml
luhring: The updates look great! 🎉

examples/go-hello.yaml
puerco: Thanks! Fixed.

luhring (approved):
This is great!

imjasonh (commented):
not blocking: What do we think about a `go/install` pipeline that just takes an import path (and optional flags, etc). This would also let us skip the `fetch` step, relying instead on Go modules to ensure integrity. Maybe `go/install` takes a module hash? 🤔

examples/go-hello.yaml
```diff
@@ -0,0 +1,28 @@
...
+  version: v0.0.1
+  epoch: 0
+  description: "simple hello go project"
+  target-architecture:
+    - all
+
+environment:
+  contents:
+    repositories:
+      - https://packages.wolfi.dev/os
```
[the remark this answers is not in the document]
puerco: Ah good catch, fixed.

puerco:
@imjasonh sounds great! I will work on that on a follow-up and update the example

kaniini (approved):

tuananh:
@kaniini how can i pass env var like CGO_ENABLED=0 to go/build?

PR #183, opened 2022-11-28
add conditionals ("if" clauses)

kaniini:
Allows things like:

```yaml
uses: fetch
if: ${{build.arch}} == 'x86_64'
with:
  uri: some-x86-64-binary-here
  ...
```

### DIFF
--- a/pkg/cond/parser_test.go
+++ b/pkg/cond/parser_test.go
@@ -15,6 +15,7 @@
 package cond
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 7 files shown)


pkg/cond/parser_test.go
```diff
@@ -0,0 +1,68 @@
+// Copyright 2022 Chainguard, Inc.
+//
```
jdolitsky, line 2: can we add a simple test for broken conditional? something like `foo ==` (missing second half)? It looks like this might cause a panic

kaniini: Sure, but panics are caught inside `goparsify`.

jdolitsky (dismissed):

jdolitsky (approved):

PR #185, opened 2022-11-29
Implement abstract interface for running things in a container

kaniini:
This is the first step to removing our dependency on Bubblewrap.

We introduce an abstract interface, `container.Runner`, which encapsulates the necessary logic to run a command in a container, given a `container.Config`, which contains a set of `container.BindMount` objects, an `Environment`, a `Logger`, and a `container.Capabilities` object which denotes what capabilities the container should have (presently networking).

`build.Context.WorkspaceCmd` is replaced with `build.Pipeline.workspaceConfig`, which determines the appropriate `container.Config` for each stage in the pipeline, which is then used in conjunction with a `container.Runner` to run commands in a container.

This interface should be sufficiently generic as to allow use of other container running strategies, such as Kubernetes Pods, lima, etc.


pkg/container/monitor_pipe.go
```diff
@@ -0,0 +1,32 @@
...
+
+package container
+
+import (
+	"bufio"
+	"io"
+	"log"
+)
+
+func MonitorPipe(logger *log.Logger, pipe io.ReadCloser, finish chan struct{}) {
```
imjasonh, line 23:
```suggestion
func monitorPipe(logger *log.Logger, pipe io.ReadCloser, finish chan struct{}) {
```

pkg/container/runner.go
```diff
@@ -0,0 +1,62 @@
...
+
+// GetRunner returns the preferred runner implementation for the
+// given environment.
+func GetRunner() Runner {
+	return BubblewrapRunner()
+}
+
+// MonitorCmd sets up the stdout/stderr pipes and then supervises
+// execution of an exec.Cmd.
+func MonitorCmd(cfg Config, cmd *exec.Cmd) error {
```
imjasonh, line 33:
```suggestion
func monitorCmd(cfg Config, cmd *exec.Cmd) error {
```

imjasonh (approved):
This looks great! Does this mean I could have a `uncontained` runner impl that just directly `os/exec`s things, and assumes/requires that it itself is running in a container?

jdolitsky (dismissed):

PR #208, opened 2022-12-12
Accept gs:// paths as --cache-dir

imjasonh:
If `--cache-dir=gs://bucket/path/to/cache` is passed for `melange build`, the matching contents of that path are fetched to a temp dir before being mounted into the build environment.

If `--cache-dir=gs://bucket/path/to/cache` is passed to `melange update-cache` then cache files are written to GCS at that location.


kaniini (dismissed):

pkg/build/build.go
```diff
@@ -777,6 +779,49 @@ func (ctx *Context) OverlayBinSh() error {
...
+			attrs, err := it.Next()
+			if err != nil {
+				return fmt.Errorf("failed to get next remote cache object: %w", err)
+			}
+			on := attrs.Name
+			rc, err := b.Object(on).NewReader(cctx)
+			if err != nil {
+				return fmt.Errorf("failed to get reader for next remote cache object %s: %w", on, err)
+			}
+			w, err := os.Create(filepath.Join(tmp, on))
```
kaniini, line 809: I wonder if it would be helpful to log each object it pulls from the cache.  That way we can be certain that it is still doing things?

imjasonh: Sure, done.

imjasonh:
I took a walk and I think we might want to do something smarter than this. Right now `--cache-dir=gs://my-bucket/` pulls down anything in `my-bucket`, puts it in a tmp dir, and all of that gets copied to the melange build environment, where _maybe_ one of those files gets used by a build's `fetch` pipeline.

This is wasteful, and especially so when we consider running `melange build --cache-dir=gs://my-bucket` over a bunch of YAMLs, since we'll keep refetching the GCS bucket and only using one file. 💣 

Instead I wonder if we should do something smarter about `--cache-dir` inspecting the pipeline YAML and only populating files it suspects will be used by the pipeline's `fetch`es, and only copying those into the build environment. Then we could have an `fs.FS` implementation backed by GCS that only pulls files it needs, and `melange build --cache-dir=gs://...` doesn't pull anything from GCS it doesn't actually need.

WDYT?

kaniini:
I like the idea.  Lets do that.

kaniini:
Per slack, we are going to move forward on this implementation and optimize later.

kaniini (approved):

PR #217, opened 2022-12-19
Include epoch in SBOM package version

luhring:
This is a follow-up to #202, which fixed the SBOM package version for subpackages. This PR applies the same fix to main packages as well.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -1086,7 +1086,7 @@ func (ctx *Context) BuildPackage() error {
 	if err := generator.GenerateSBOM(&sbom.Spec{
 		Path:           filepath.Join(ctx.WorkspaceDir, "melange-out", ctx.Configuration.Package.Name),
 		PackageName:    ctx.Configuration.Package.Name,
-		PackageVersion: ctx.Configuration.Package.Version,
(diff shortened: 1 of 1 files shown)


pkg/build/build.go
```diff
@@ -1086,7 +1086,7 @@ func (ctx *Context) BuildPackage() error {
 	if err := generator.GenerateSBOM(&sbom.Spec{
 		Path:           filepath.Join(ctx.WorkspaceDir, "melange-out", ctx.Configuration.Package.Name),
 		PackageName:    ctx.Configuration.Package.Name,
-		PackageVersion: ctx.Configuration.Package.Version,
+		PackageVersion: fmt.Sprintf("%s-r%d", ctx.Configuration.Package.Version, ctx.Configuration.Package.Epoch),
```
cpanato, line 1089: the `r` is expected?
besides that lgtm

cpanato (approved):

pkg/build/build.go
luhring: Yes — we want apk versions to look like `1.35.0-r4` (if the epoch is `4`)

PR #228, opened 2022-12-21
add news for melange 0.2.0

kaniini:


NEWS.md
```diff
@@ -1,3 +1,86 @@
...
+
+* Added initial support for generating SBOMs for each package.
+  These SBOMs will be used with other composition tools to generate
+  so-called multi-layer SBOMs.
+
+* Added support for breakpoint labels in pipelines.  These allow
+  a build to be suspended or resumed from a specific point in the
+  pipeline.  Care must be used to ensure the build environment
+  locations are supplied when resuming a build, or a new build
+  environment will be created.
```
erikaheidi, line 14: This is all great. The only suggestion I have is to change the part here that says "Care must be used..." to something like "In order to use this feature, a build environment location must be provided when resuming a build, otherwise a new build environment will be created."

erikaheidi (commented):
This is all great, I had only one suggestion. Thanks for providing examples, these will be very useful for updating the docs.

PR #229, opened 2022-12-23
Export functionality for config parsing

luhring:
### Background

Previously, the best method available for parsing a Melange config was to call `Load`, which required creating a `build.Context` first.

As our ecosystem consolidates its Melange config parsing logic into Melange itself (which is a good thing), it needs to be as easy as possible to get a decoded configuration to meet a given consumer's needs.

### What's changed

This PR introduces a new `ParseConfiguration` function, along with options for adjusting configuration parsing behavior.

This PR also introduces a `Logger` interface so that it's easier to use the stdlib `log.Logger`, _or not_, depending on one's needs.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -554,29 +554,109 @@ func WithEnvFile(envFile string) Option {
 	}
 }
 
-// Load the configuration data from the build context configuration file.
-func (cfg *Configuration) Load(ctx Context) error {
-	data, err := os.ReadFile(ctx.ConfigFile)
-	if err != nil {
-		return fmt.Errorf("unable to load configuration file: %w", err)
+type ConfigurationParsingOption func(*configOptions)
+
+type configOptions struct {
+	abstractFS     fs.FS
+	configFilePath string
+	envFilePath    string
+	logger         Logger
+}
+
+// include reconciles all given opts into the receiver variable, such that it is
+// ready to use for config parsing.
+func (options *configOptions) include(opts ...ConfigurationParsingOption) {
+	for _, fn := range opts {
+		fn(options)
+	}
+	if options.logger == nil {
+		options.logger = nopLogger{}
+	}
+}
+
+// open handles opening of a file with the given name, in such a way that
+// automatically chooses to use an abstract or concrete OS filesystem.
+func (options *configOptions) open(filepath string) (fs.File, error) {
+	if filesystem := options.abstractFS; filesystem != nil {
+		return filesystem.Open(filepath)
 	}
 
-	if err := yaml.Unmarshal(data, cfg); err != nil {
-		return fmt.Errorf("unable to parse configuration file: %w", err)
-	}
+	return os.Open(filepath)
+}
 
+// WithAbstractFS adjusts the behavior of configuration parsing to expect all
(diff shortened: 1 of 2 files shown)


pkg/build/build.go
```diff
@@ -554,30 +554,110 @@ func WithEnvFile(envFile string) Option {
...
+		return ""
+	}
+
+	commit := head.Hash().String()
+	options.logger.Printf("detected git commit for build configuration: %s", commit)
+	return commit
+}
+
+// ParseConfiguration returns a decoded build Configuration using the parsing options provided.
+func ParseConfiguration(opts ...ConfigurationParsingOption) (*Configuration, error) {
```
luhring, line 639: Maybe the config file itself should be a normal parameter since we'll always want to have it 🤔

pkg/build/build.go
```diff
@@ -554,30 +554,110 @@ func WithEnvFile(envFile string) Option {
...
+		return filesystem.Open(filepath)
+	}
+
+	return os.Open(filepath)
+}
+
+// WithAbstractFS adjusts the behavior of configuration parsing to expect all
+// file paths to be relative to the abstract fs.FS. This approach isn't supported
+// yet except for when opening the config file.
+func WithAbstractFS(filesystem fs.FS) ConfigurationParsingOption {
```
imjasonh, line 590: Nit: maybe just WithFS?

pkg/build/build.go
```diff
@@ -554,30 +554,110 @@ func WithEnvFile(envFile string) Option {
...
+	}
+	if options.logger == nil {
+		options.logger = nopLogger{}
+	}
+}
+
+// open handles opening of a file with the given name, in such a way that
+// automatically chooses to use an abstract or concrete OS filesystem.
+func (options *configOptions) open(filepath string) (fs.File, error) {
+	if filesystem := options.abstractFS; filesystem != nil {
```
imjasonh, line 580: Instead of switching here can we make the default fs a real filesystem fs?

pkg/build/logger.go
```diff
@@ -0,0 +1,12 @@
+package build
+
+type Logger interface {
```
imjasonh, line 3: There's already some logging abstraction in apko/melange, can we reuse it?

pkg/build/logger.go
```diff
@@ -0,0 +1,12 @@
...
+	Printf(format string, v ...any)
+	SetPrefix(prefix string)
+}
+
+type nopLogger struct{}
+
+func (n nopLogger) Printf(_ string, _ ...any) {}
```
imjasonh, line 10: You can drop the _s in this case

luhring: 👍

pkg/build/build.go
luhring: 👍

pkg/build/build.go
luhring: 👍

pkg/build/build.go
luhring: 👍

pkg/build/logger.go
luhring: I couldn't find an existing interface like this, but I definitely want to use what already exists where possible

imjasonh (approved):

PR #230, opened 2022-12-23
Add advisories and purls

luhring:
- Moves `Secfixes` parsing into Melange (follow-up work still needed to adapt Wolfi's secdb to this)
- Adds new `Advisories` section to Melange config, based on latest types from https://github.com/chainguard-dev/vex
- Adds the ability to produce package URLs ("purls") for Melange packages


puerco (approved):
Awesome! :rocket:

PR #232, opened 2022-12-26
New pipeline go/install

puerco:
This PR adds a small improvement to the `go/build` pipeline to allow running a build in a different directory than the current $CWD. More importantly, it introduces a new `go/install` pipeline [as suggested by @imjasonh](https://github.com/chainguard-dev/melange/pull/180#pullrequestreview-1192458615)  that runs a go build using `go install ...`.

Here's a sample from my tests:

```yaml
pipeline:
  - uses: go/install
    with:
      package: github.com/uservers/miniprow/actions/broker
      version: HEAD
```
Using go install lets us rely on the go binary to download all packages and avoid having to clone repos or `fetch` manually.

### DIFF
--- a/pkg/build/pipelines/go/build.yaml
+++ b/pkg/build/pipelines/go/build.yaml
@@ -10,7 +10,8 @@ inputs:
   packages:
     description: |
       List of space-separated packages to compile. Files con also be specified.
-      This value is passed as an argument to go build.
+      This value is passed as an argument to go build. All paths are relative
+      to inputs.modroot.
     required: true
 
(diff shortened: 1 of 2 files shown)


imjasonh (approved):
I like it! I think we'll need to consider how this will work with hermetic builds, but we can cross that bridge later.

imjasonh (approved):
Oh can you add an example? We can always use more examples.

puerco:
Yes, I'm working now on the docs for this one and `go/build`!

PR #235, opened 2022-12-29
Add purl generation and --namespace option to melange build

puerco:
This PR adds package URLs to entries in the SBOMs describing the packages built by melange. In order to configure the purl we now expose a `--namespace` flag that sets the string which will be used as the namespace value in the purl.

@luhring : This PR also removes the aliased `distro=` qualifier that used the same value as the namespace. As wolfi is a rolling (un)distro, we don't need to define it now (we only have one flavor at the moment). If we need a distro qualifier we can add it later.

Here's an example:

### Invocation:
```
melange build --namespace=wolfi --arch=x86_64 openssh.yaml
```

### Output (fragment from the produced openssh-server apk SBOM):
```json

"packages": [
    {
      "SPDXID": "SPDXRef-Package-openssh-server",
      "name": "openssh-server",
      "versionInfo": "9.0_p1-r0",
      "filesAnalyzed": true,
      "hasFiles": [
        "SPDXRef-File-/etc/ssh/sshd_config",
        "SPDXRef-File-/usr/sbin/sshd"
      ],
      "licenseConcluded": "NOASSERTION",
      "licenseDeclared": "ISC",
      "downloadLocation": "NOASSERTION",
      "copyrightText": "TODO\n",
      "checksums": [],
      "externalRefs": [
        {
          "referenceCategory": "PACKAGE_MANAGER",
          "referenceLocator": "pkg:apk/wolfi/openssh-server@9.0_p1-r0?arch=x86_64",
          "referenceType": "purl"
        }
      ],
      "packageVerificationCode": {
        "packageVerificationCodeValue": "01b354575a7938f3f64cdf7458059d85d46c6a68"
      }
    }
  ],
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/joho/godotenv v1.4.0
 	github.com/korovkin/limiter v0.0.0-20221015170604-22eb1ceceddc
 	github.com/oec/goparsify v0.2.1
+	github.com/package-url/packageurl-go v0.1.1-0.20220203205134-d70459300c8a
 	github.com/psanford/memfs v0.0.0-20210214183328-a001468d78ef
 	github.com/spf13/cobra v1.6.1
 	github.com/stretchr/testify v1.8.1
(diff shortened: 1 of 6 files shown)


luhring (approved):
>This PR also removes the aliased distro= qualifier that used the same value as the namespace. As wolfi is a rolling (un)distro, we don't need to define it now (we only have one flavor at the moment). If we need a distro qualifier we can add it later.

That makes sense. I haven't seen many distro package purls _not_ have the `distro` parameter, but I like the idea of waiting to add it until we absolutely need it.

PR #237, opened 2022-12-31
Advisories: Require pkg version for fixed status

luhring:
Signed-off-by: Dan Luhring <dluhring@chainguard.dev>


puerco (approved):
Nice ;)

PR #239, opened 2023-01-03
go/install: also require git

imjasonh:
A lot of `go install`s seem to require `git` to clone the source ([example](https://github.com/wolfi-dev/os/actions/runs/3831141870/jobs/6519896216#step:8:197)), so let's depend on it in the `go/install` pipeline too.

For [that PR](https://github.com/wolfi-dev/os/pull/360) I just added an explicit dependency on `git`, but ideally we shouldn't have to.

This might mean we should also add dependencies on `svn`, `hg`, etc., but for now let's just depend on `git`.


developer-guy (approved):
LGTM

luhring (approved):

PR #245, opened 2023-01-06
Close lingering file descriptor at index generation

puerco:
This commit fixes a bug that caused melange to leak file descriptors when generating the apk index, possibly hitting the max open files limit.

Signed-off-by: Adolfo García Veytia (Puerco) <puerco@chainguard.dev>

### DIFF
--- a/pkg/index/index.go
+++ b/pkg/index/index.go
@@ -21,10 +21,12 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
+
+	"github.com/korovkin/limiter"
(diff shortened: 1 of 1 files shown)


puerco:
Scratch that I think we need to throttle the loop

dlorenc:
Feel free to roll my parallelism change back if that's faster!

puerco:
Nah I chased the index generation to the end and this should do it. Ready to go :steam_locomotive:

puerco:
I was waiting for another process to finish so I added the throttler and a mutex to prevent a concurrent write on the packages slice.

dlorenc (approved):

PR #262, opened 2023-01-17
Add ruby Gem pipelines

strongjz:


pkg/build/pipelines/ruby/install.yaml
```diff
@@ -0,0 +1,20 @@
+name: Run a gem install
+
+needs:
+  packages:
+    - busybox
+    - ca-certificates-bundle
```
imjasonh, line 6: Does this pipeline need to depend on a package that provides `gem`?

imjasonh: In other words, I think this should depend on `ruby-3.2` so that it can know it has `gem` available. But this leads to a couple questions:

1. should it depend on `ruby-3.2` or `ruby-3.1`, or just anything that says it `provides: ruby`?
2. `ruby-3.2` isn't a package name in Alpine-land, so this package will be distro-specific if we name it `ruby-3.2`. This leads to a sort of general question about whether/how pipelines can depend on distro-agnostic package names, or distro-specific package names.

kaniini: We might be able to depend on `cmd:gem`.

pkg/build/pipelines/ruby/install.yaml
```diff
@@ -0,0 +1,33 @@
...
+      gem build ${{inputs.gem}}.gemspec
+      
+      gem install \
+      --local \
+      --install-dir ${{targets.destdir}}  \
+      --bindir ${{targets.destdir}}/usr/bin \
+      --ignore-dependencies \
+      --document ri \
+      --verbose \
+      -v ${{inputs.version}} \
```
kaniini, line 29: If `-v ${{inputs.version}}` is always used, then `version` should be `required`?

strongjz: good point

pkg/build/pipelines/ruby/install.yaml
strongjz: I was going to ask, but the test I was using already required ruby.

pkg/build/pipelines/ruby/build.yaml
```diff
@@ -0,0 +1,18 @@
+name: Build a ruby gem
+
+needs:
+  packages:
+    - busybox
+    - ca-certificates-bundle
+    - ruby-3.2
```
imjasonh, line 7: This will require a package named `ruby-3.2` to exist in the package repo (ruling out Alpine-based pipelines using this workflow)

This will also mean that folks who want to build a gem for `ruby-3.1` (or just `ruby` for Alpine) won't be able to.

We could create variations of this pipeline like `ruby-3.2/build`, `ruby-3.1/build`, etc., but that gets messy.

We might just want to put this pipeline in wolfi-dev/os for now, and update the build system to set `--pipeline-dir=./pipelines/` so Wolfi's configs can use it. We'll still probably need a `ruby-3.1` variant, but at least the damage will be localized to Wolfi.

imjasonh: Oh this seems to have already been addressed in https://github.com/chainguard-dev/melange/pull/262#discussion_r1105359773

Resolving this in favor of unresolving that.

pkg/build/pipelines/ruby/install.yaml
```diff
@@ -0,0 +1,39 @@
...
+      --bindir ${TARGET_DIR_BIN} \
+      --no-document \
+      --ignore-dependencies \
+      --local \
+      --verbose \
+      -v ${{inputs.version}} \
+      ${{inputs.gem}}
+      
+      cd ${{targets.destdir}}
+      rm -rf build_info cache extensions plugins
```
found-it, line 39: This was my mistake, it should be
```sh
cd ${TARGET_DIR_INSTALL} && rm -rf build_info cache extensions plugins
```

found-it: Or perhaps we turn this into a `ruby/clean` pipeline

pkg/build/pipelines/ruby/install.yaml
```diff
@@ -0,0 +1,39 @@
...
+      
+      gem install \
+      --install-dir ${TARGET_DIR_INSTALL}  \
+      --bindir ${TARGET_DIR_BIN} \
+      --no-document \
+      --ignore-dependencies \
+      --local \
+      --verbose \
+      -v ${{inputs.version}} \
+      ${{inputs.gem}}
```
found-it, line 36: What do you think about changing this to 

```
${{inputs.gem}}-${{inputs.version}}.gem
```

That way the user doesn't need to know the structure of the gem filename? We could add an optional `inputs.gem-filename` in case the gemspec generates a non-standard filename for the gem

found-it: Then the use would look like

```yaml
pipeline:
  uses: ruby/build
    gem: webrick
  uses: ruby/install
    gem: webrick
    version: ${{package.version}}
```

pkg/build/pipelines/ruby/install.yaml
found-it: @imjasonh @kaniini I'd love to get this merged in so I can use exercise it for all the ruby work I'm doing. Is this

> We might be able to depend on cmd:gem.

a possibility right now? Otherwise what do you think about leaving the ruby dependency off for now? Then in the future we could do something like the `go` pipelines? Looks like they just depend on the latest version of `go`, or at least whatever is published under that package

pkg/build/pipelines/ruby/install.yaml
strongjz: added a clean pipeline

pkg/build/pipelines/ruby/install.yaml
strongjz: nice, done

pkg/build/pipelines/ruby/install.yaml
strongjz: I added the which ruby test to force a user to have ruby installed.

kaniini: We can just require the user provide a ruby for now, yes.

imjasonh (approved):

PR #263, opened 2023-01-19
add multiple runner options

deitch:
Adds multiple runner (virtualizer) options. For the command execution parts, supports bubblewrap, docker and lima.

For lima, it looks for a VM named `melange-builder` with the right config. If it does not find it, it starts one. It then uses containerd on the lima VM to do the actual build via nerdctl. As with bubblewrap and docker, it builds the base filesystem. As with docker, it creates an OCI image and loads it into the container runner (in this case, containerd) to use as the image.

This builds on the apko without CLI in http://github.com/chainguard-dev/apko/pull/426 

Some of these designs are pre-docker, I adapted as well, but I might have changed something unexpectedly, so a good review is appreciated.


deitch:
I do not quite get why CI is failing. All failures are an attempt to build a package, fails at the same attempt to use `bwrap`. Since that didn't change from before, why does that fail now?

kaniini:
> I do not quite get why CI is failing. All failures are an attempt to build a package, fails at the same attempt to use bwrap. Since that didn't change from before, why does that fail now?

on minimal, I suspect it is because the `/bin/sh` symlink points to the wrong place.

on mbedtls, it is because `ninja` is provided by `samurai`, but the solver does not handle this it seems.

pkg/container/lima.yaml
```diff
@@ -0,0 +1,30 @@
+# we pretty much take the default of everything, except for specifying the images and the mounts
+
+images:
+# Try to use release-yyyyMMdd image if available. Note that release-yyyyMMdd will be removed after several months.
```
kaniini, line 4: We should generate wolfi-based images for lima, although that will probably require some platform adaptation work.

pkg/container/runner.go
```diff
@@ -17,31 +17,42 @@ package container
...
-		}
+// GetRunner returns the requested runner implementation.
+func GetRunner(s string) (Runner, error) {
+	switch s {
+	case "bubblewrap":
+		return BubblewrapRunner(), nil
+	case "docker":
+		return DockerRunner(), nil
+	case "lima":
+		return LimaRunner()
```
kaniini, line 53: shouldn't this be

```suggestion
		return LimaRunner(), nil
```

?

pkg/container/lima.yaml
deitch: That makes a lot of sense. Is that something we can do here, or is it enough that we can start with this and then iterate?

deitch:
> on minimal, I suspect it is because the /bin/sh symlink points to the wrong place.

Yes, that was it. It has to do with busybox links messed up.

deitch:
OK, got it. This is as a result of an apko decision, which just showed up here.

apko has the logic to install the busybox symlinks, but it only does so when preparing the tarball, not when building the layout on the guest directory. This is at least partially intentional; we do not want to install symlinks in the guest, as the guest might not be able to handle it, depending on the filesystem (linux and Unix-like are ok, others may not be).

We do not have an issue with the docker and lima runners, as they take the guest layout, tar it up (getting busybox symlinks at that stage) and then load it into docker or containerd. bubblewrap, on the other hand, just mounts the guest directory, so it never gets to the tarball stage, and hence to the busybox links installed.

```go
	loader := ctx.Runner.OCIImageLoader()
	if loader != nil {
		layerTarGZ, err := bc.ImageLayoutToLayer(overrides)
```

For bubblewrap, `loader == nil`.

I see 2 potential avenues to fix this.

* "no busybox symlinks until tarball" is correct: fix it in melange by creating the layer tar and then untarring it.
* "no busybox symlinks until tarball" is incorrect: fix it in apko by creating the busybox links at an earlier stage

Thoughts @kaniini ?

jdolitsky:
Just trying this now! Running into an issue with lima runner:
```
$ melange build minimal.yaml --arch aarch64 --signing-key melange.rsa --runner lima
2023/01/20 17:59:16 limactl [list melange-builder --json]
WARN[0000] No instance matching melange-builder found.  
2023/01/20 17:59:16 limactl [start --name melange-builder --tty=false /dev/stdin]
INFO[0000] Terminal is not available, proceeding without opening an editor 
FATA[0000] got empty st.yBytes                          
Error: exit status 1
2023/01/20 17:59:16 error during command execution: exit status 1
```

However, it looks like I can start the Lima VM myself with the following:
```
limactl start --tty=false --name melange-builder https://raw.githubusercontent.com/chainguard-dev/melange/acc604c2d1ae1d2575e04f2f9b531e8e0cb3e038/pkg/container/lima.yaml
```

Then it works!! 🎉 Holy cow

---

Taking it a step further...

With an apko config like this (apko.yaml) containing the minimal package built in lima:
```yaml
contents:
  repositories:
    - https://dl-cdn.alpinelinux.org/alpine/edge/main
  packages:
    - busybox
    - minimal
entrypoint:
  command: cat
cmd: /hello
```

and the latest apko from main, build an image:

```
apko build apko.yaml --build-arch aarch64 --repository-append "${PWD}/packages" --keyring-append "${PWD}/melange.rsa.pub" minimal:latest output.tar
```

load it:

```
docker load < output.tar
```

run it!
```
$ docker run --rm minimal:latest
hello
```

🤯

jdolitsky:
Running melange without the `--arch` flag appears to work at first, but then crashes with some issue relating to containerd:
```
2023/01/20 18:20:06 limactl [shell --workdir /tmp/melange melange-builder sudo nerdctl run --privileged --rm tonistiigi/binfmt:qemu-v7.0.0-28 --install all]
2023/01/20 18:20:06 limactl [shell --workdir /tmp/melange melange-builder sudo nerdctl run --privileged --rm tonistiigi/binfmt:qemu-v7.0.0-28 --install all]
2023/01/20 18:20:06 limactl [shell --workdir /tmp/melange melange-builder sudo nerdctl run --privileged --rm tonistiigi/binfmt:qemu-v7.0.0-28 --install all]
2023/01/20 18:20:06 limactl [shell --workdir /tmp/melange melange-builder sudo systemctl stop containerd]
2023/01/20 18:20:06 limactl [shell --workdir /tmp/melange melange-builder nerdctl run --detach --volume /tmp/melange/melange-workspace-2533812290:/home/build --volume /etc/resolv.conf:/etc/resolv.conf --env SOURCE_DATE_EPOCH=0 --env HOME=/home/build --env GOPATH=/home/build/.cache/go --platform=ppc64le 751f25bdba333dfca34c0f09a83c845b056ce18b46a0ae55241d6929bf9c779e /bin/sh -c while true; do sleep 5; done]
time="2023-01-21T00:20:06Z" level=fatal msg="invalid repository name (751f25bdba333dfca34c0f09a83c845b056ce18b46a0ae55241d6929bf9c779e), cannot specify 64-byte hexadecimal strings"
2023/01/20 18:20:06 ERROR: failed to build package. the build environment has been preserved:
2023/01/20 18:20:06 melange (minimal/ppc64le):   workspace dir: /tmp/melange/melange-workspace-2533812290
2023/01/20 18:20:06 melange (minimal/ppc64le):   guest dir: /tmp/melange/melange-guest-2867547032
time="2023-01-21T00:20:07Z" level=fatal msg="error unpacking image: error reading from server: EOF: unavailable"
time="2023-01-21T00:20:07Z" level=fatal msg="lstat /run/containerd/containerd.sock: no such file or directory"
time="2023-01-21T00:20:07Z" level=fatal msg="error reading from server: EOF: unavailable"
2023/01/20 18:20:07 ERROR: failed to build package. the build environment has been preserved:
2023/01/20 18:20:07 melange (minimal/s390x):   workspace dir: /tmp/melange/melange-workspace-892086360
2023/01/20 18:20:07 melange (minimal/s390x):   guest dir: /tmp/melange/melange-guest-1270157436
2023/01/20 18:20:07 ERROR: failed to build package. the build environment has been preserved:
2023/01/20 18:20:07 melange (minimal/x86):   workspace dir: /tmp/melange/melange-workspace-3243453554
2023/01/20 18:20:07 melange (minimal/x86):   guest dir: /tmp/melange/melange-guest-3901445681
2023/01/20 18:20:07 ERROR: failed to build package. the build environment has been preserved:
2023/01/20 18:20:07 melange (minimal/aarch64):   workspace dir: /tmp/melange/melange-workspace-3149525785
2023/01/20 18:20:07 melange (minimal/aarch64):   guest dir: /tmp/melange/melange-guest-4168995096
Error: failed to build package: unable to build guest: failed to load image into containerd: exit status 1 time="2023-01-21T00:20:05Z" level=info msg="apply failure, attempting cleanup" error="failed to extract layer sha256:34ee2ab7a4fbb1372713abe517e626107c4d673db750a87b43de4197cd22b156: failed to get reader from content store: content digest sha256:2d5e9e950480f5cf041d41e9d9a729cb51cb2c06a0f20aeec3cd505c40774621: not found" key="extract-880344689-2YbO sha256:34ee2ab7a4fbb1372713abe517e626107c4d673db750a87b43de4197cd22b156"
time="2023-01-21T00:20:05Z" level=fatal msg="failed to extract layer sha256:34ee2ab7a4fbb1372713abe517e626107c4d673db750a87b43de4197cd22b156: failed to get reader from content store: content digest sha256:2d5e9e950480f5cf041d41e9d9a729cb51cb2c06a0f20aeec3cd505c40774621: not found"

2023/01/20 18:20:07 error during command execution: failed to build package: unable to build guest: failed to load image into containerd: exit status 1 time="2023-01-21T00:20:05Z" level=info msg="apply failure, attempting cleanup" error="failed to extract layer sha256:34ee2ab7a4fbb1372713abe517e626107c4d673db750a87b43de4197cd22b156: failed to get reader from content store: content digest sha256:2d5e9e950480f5cf041d41e9d9a729cb51cb2c06a0f20aeec3cd505c40774621: not found" key="extract-880344689-2YbO sha256:34ee2ab7a4fbb1372713abe517e626107c4d673db750a87b43de4197cd22b156"
time="2023-01-21T00:20:05Z" level=fatal msg="failed to extract layer sha256:34ee2ab7a4fbb1372713abe517e626107c4d673db750a87b43de4197cd22b156: failed to get reader from content store: content digest sha256:2d5e9e950480f5cf041d41e9d9a729cb51cb2c06a0f20aeec3cd505c40774621: not found"
```

deitch:
FYI, the issue causing the build failures should be fixed once [apko PR 471](https://github.com/chainguard-dev/apko/pull/471) is in. That has one issue still blocking it.

deitch:
> Running into an issue with lima runner:
> However, it looks like I can start the Lima VM myself with the following:

Well, that doesn't sound too right. 😁 

Looking.

deitch:
@jdolitsky give another shot now with no lima runner yet started. It should work.

deitch:
> Running melange without the --arch flag appears to work at first, but then crashes with some issue relating to containerd:

OK, working on it

deitch:
As soon as [this apko one](https://github.com/chainguard-dev/apko/pull/473) is in, I will update the dependency and rebase here. It should fix the issues.

deitch:
apko PR is in, this imports it. Also rebased. Will let CI run, once it is clean, I will address the issue @jdolitsky raised above.

deitch:
Oh well, not quite there yet. Will keep at it.

kaniini:
> I see 2 potential avenues to fix this.
>
>    "no busybox symlinks until tarball" is correct: fix it in melange by creating the layer tar and then untarring it.
>    "no busybox symlinks until tarball" is incorrect: fix it in apko by creating the busybox links at an earlier stage
>
> Thoughts @kaniini ?

Yes, I see why this is a problem.  I think it would be cleaner to just always think of apko's output as a layer, either one that is sent to a registry (such as the local dockerd/containerd), or provided as a tarball.  So, we could have apko generate a tar instead of working directly in the GuestDir, and then just untar it to the GuestDir once it has done its thing.

jdolitsky:
Just checking back in - lima autostart works fine now! Still running into similar issues with multiarch melange build, but this is great for building for single arch on mac

jdolitsky:
In a bit more advanced case, trying to build [tree from wolfi](https://github.com/wolfi-dev/os/blob/main/tree.yaml) results in
```
Error: failed to build package: unable to build guest: failed to generate tarball for image: writing TAR archive failed: mismatched entries in filesystems disk vs memory
2023/01/25 19:13:41 error during command execution: failed to build package: unable to build guest: failed to generate tarball for image: writing TAR archive failed: mismatched entries in filesystems disk vs memory
```

kaniini (approved):

PR #270, opened 2023-01-23
pipelines: add working-directory modifier

kaniini:
This allows pipelines like:

```yaml
pipeline:
  - working-directory: /home/build/foo
    runs: |
      echo "current working directory: $(pwd)"
  - working-directory: /home/build/bar
    pipeline:
      - runs: |
          echo "current working directory: $(pwd)"
      - working-directory: /home/build/baz
        runs: |
          echo "current working directory: $(pwd)"
```

This will output:

```
current working directory: /home/build/foo
current working directory: /home/build/bar
current working directory: /home/build/baz
```

`working-directory` can also be used with `uses`, e.g.

```yaml
pipeline:
  - working-directory: /home/build/srcA
    uses: fetch
    with:
      ...
```

In that example, the fetched source will be populated in `/home/build/srcA`, rather than the default `/home/build`.

cc @luhring

### DIFF
--- a/pkg/build/pipeline.go
+++ b/pkg/build/pipeline.go
@@ -227,9 +227,14 @@ func (p *Pipeline) evalRun(ctx *PipelineContext) error {
 	p.With = mutateWith(ctx, p.With)
 	p.dumpWith()
 
+	workdir := "/home/build"
+	if p.WorkDir != "" {
+		workdir = mutateStringFromMap(p.With, p.WorkDir)
+	}
+
 	fragment := mutateStringFromMap(p.With, p.Runs)
 	sys_path := "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
-	script := fmt.Sprintf("#!/bin/sh\nset -e\nexport PATH=%s\n%s\nexit 0\n", sys_path, fragment)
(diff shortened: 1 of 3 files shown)


pkg/build/pipeline.go
```diff
@@ -227,9 +228,14 @@ func (p *Pipeline) evalRun(ctx *PipelineContext) error {
...
 
+	workdir := "/home/build"
+	if p.WorkDir != "" {
+		workdir = mutateStringFromMap(p.With, p.WorkDir)
+	}
+
 	fragment := mutateStringFromMap(p.With, p.Runs)
 	sys_path := "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
-	script := fmt.Sprintf("#!/bin/sh\nset -e\nexport PATH=%s\n%s\nexit 0\n", sys_path, fragment)
+	script := fmt.Sprintf("#!/bin/sh\nset -e\nexport PATH=%s\nmkdir -p '%s'\ncd '%s'\n%s\nexit 0\n", sys_path, workdir, workdir, fragment)
```
luhring, line 238: (non-blocking) it may help w/ readability in the future for inline scripts like this to use ["raw string literals"](https://go.dev/ref/spec#String_literals) (i.e. using back-ticks and literal newlines instead of `\n`)

luhring (approved):
Nice! 👏

PR #271, opened 2023-01-25
add build variable configuration block, --vars-file build option

kaniini:
This adds a new section to the melange configuration: a *variable configuration block*.

```yaml
vars:
  uri: https://example.com/foo.tar.gz
  expected-sha256: 2f2846c86932b2b70dcc791e58e8c2c1e11d0ec25ba3df5ed67641d73e763822
  build-directory: /home/build/foobar
  baz: "hello world"
```

These variables can be used as one would expect, e.g.

```yaml
pipeline:
  - working-directory: ${{vars.build-directory}}
    uses: fetch
    with:
      uri: ${{vars.uri}}
      expected-sha256: ${{vars.expected-sha256}}
  - runs: |
      echo "${{vars.baz}}"
```

A yaml file containing an alternative variable configuration block can be specified with `--vars-file`.  The variables defined in that alternative configuration block file will overwrite the variable data defined in the configuration file.

### DIFF
--- a/examples/vars.yaml
+++ b/examples/vars.yaml
@@ -1,0 +1,37 @@
+package:
+  name: hello
+  version: 2.12
+  epoch: 0
+  description: "an example of how conditionals influence build behavior"
+  target-architecture:
+    - all
+  copyright:
+    - paths:
+      - "*"
+      license: Not-Applicable
+  dependencies:
+    runtime:
+
+environment:
+  contents:
+    repositories:
+      - https://dl-cdn.alpinelinux.org/alpine/edge/main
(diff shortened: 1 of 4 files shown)


puerco (approved):
Love this, its great @kaniini :tada: 

I think we should document the format of the vars file somewhere in a follow-up.

jdolitsky (approved):

tuananh:
Love this one. The melange file will be a lot cleaner and most likely doesn't get updated as often as right now. this makes code review easier as well as automation easier (yaml formatting etc...)

PR #272, opened 2023-01-25
Work on guest interface for supporting multiple ways to build the guest.

azenla:
As part of my effort for supporting melange builds on macOS for macOS, I require the use of something other than apko. This PR makes the first steps at decoupling apko directly from the build and turns it into a pluggable interface. Each runner is responsible for specifying what guest it wants to use. This gives the option for an external runner to return a different guest building implementation to fulfill the environment request of a melange file.


azenla:
I will likely be rebasing this PR off of #263 as there are significant changes there. I will keep this PR in draft for the time being.

PR #256, opened 2023-01-14
Generate build environment SBOM

puerco:
This PR modifies melange to generate an SBOM describing the build environment assembled to build the apks in a build. The new SBOM lists all the SBOMs installed into the build environment in a package describing them as a bundle.

### Structure

The build environment package links to the produced apks identifying itself as a `BUILD_TOOL_OF` of the package. It looks like this:

```
 📂 SPDX Document apk-hello-2.12-r0
  │ 
  │ 📦 DESCRIBES 1 Packages
  │ 
  ├ hello-build@2.12-r0
  │  │ 🔗 29 Relationships
  │  ├ DEPENDS_ON PACKAGE glibc-locale-posix@2.36-r4
  │  ├ DEPENDS_ON PACKAGE wolfi-baselayout@20221118-r0
  │  ├ DEPENDS_ON PACKAGE glibc@2.36-r4
  │  ├ DEPENDS_ON PACKAGE binutils@2.39-r4
  │  ├ DEPENDS_ON PACKAGE libstdc++-dev@12.2.0-r6
  │  ├ DEPENDS_ON PACKAGE libgcc@12.2.0-r6
  │  ├ DEPENDS_ON PACKAGE libstdc++@12.2.0-r6
  │  ├ DEPENDS_ON PACKAGE gmp@6.2.1-r4
  │  ├ DEPENDS_ON PACKAGE isl@0.24-r2
  │  ├ DEPENDS_ON PACKAGE mpfr@4.2.0-r0
  │  ├ DEPENDS_ON PACKAGE mpc@1.2.1-r2
  │  ├ DEPENDS_ON PACKAGE zlib@1.2.13-r1
  │  ├ DEPENDS_ON PACKAGE gcc@12.2.0-r6
  │  ├ DEPENDS_ON PACKAGE linux-headers@5.19.17-r0
  │  ├ DEPENDS_ON PACKAGE glibc-dev@2.36-r4
  │  ├ DEPENDS_ON PACKAGE make@4.3-r1
  │  ├ DEPENDS_ON PACKAGE pkgconf@1.9.3-r3
  │  ├ DEPENDS_ON PACKAGE build-base@1-r3
  │  ├ DEPENDS_ON PACKAGE scanelf@1.3.4-r2
  │  ├ DEPENDS_ON PACKAGE libcrypto3@3.0.7-r1
  │  ├ DEPENDS_ON PACKAGE libssl3@3.0.7-r1
[...]

PR #273, opened 2023-01-27
Only create build env SBOM if flag set

jdolitsky:


imjasonh (approved):

imjasonh:
cc @puerco

luhring:
related: #256

luhring (approved):

kaniini:
I would rather not merge this.  The purpose of capturing the build environment in the SBOM is to allow the SBOM to be used as a source of truth when reproducing a build, and I have plans to make use of this functionality in Melange in the near future.  If the build environment is not captured by default, then we cannot move forward on verified reproducibility of Wolfi, unless an alternate design is implemented.

jdolitsky:
closing in favor of #276

Issue #264, opened 2023-01-23
FR: add `tag` and `expected-commit` to `git-checkout` pipeline


imjasonh:
https://github.com/chainguard-dev/melange/blob/main/pkg/build/pipelines/git-checkout.yaml

The `git-checkout` pipeline makes it really easy to `git clone` a branch. We should also support cloning a tag, and have a param to check that the cloned commit matches some expectation.

This would make it more similar to `fetch`, and could also help us make `git-checkout` use cases hermetic by pre-fetching (and verifying) source fetches from Git repos.

PR #275, opened 2023-01-27
add expected-commit to git-checkout

tuananh:
Fixes: #264 
Signed-off-by: Tuan Anh Tran <me@tuananh.org>

### DIFF
--- a/pkg/build/pipelines/git-checkout.yaml
+++ b/pkg/build/pipelines/git-checkout.yaml
@@ -24,7 +24,37 @@ inputs:
     description: |
       The branch to check out, otherwise HEAD is checked out.
 
+  tag:
+    description: |
+      The tag to check out.
+  
+  expected-commit:
+    description: |
+      The expected commit hash
+
 pipeline:
   - runs: |
-      [ -n '${{inputs.branch}}' ] && branch='--branch ${{inputs.branch}}'
-      git clone $branch --depth ${{inputs.depth}} ${{inputs.repository}} ${{inputs.destination}}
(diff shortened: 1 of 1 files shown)


kaniini (approved):

pkg/build/pipelines/git-checkout.yaml
```diff
@@ -24,7 +24,36 @@ inputs:
...
+      [ -n '${{inputs.tag}}' ] && clone_target='tags/${{inputs.tag}}'
+
+      git clone $clone_target --depth ${{inputs.depth}} ${{inputs.repository}} ${{inputs.destination}}
+
+      cd ${{inputs.destination}}
+
+      if [ -z "${{inputs.expected-commit}}" ]; then
+        echo "Warning (git-checkout): no expected-commit"
+      else
+        [ -n '${{inputs.branch}}' ] && remote_commit=$(git rev-list -1 origin/"${{inputs.branch}}")
```
imjasonh, line 51: Does it make sense to specify an expected commit when you specify a branch? Those are more or less expected to move as I understand it.

Maybe this should fail if you specify `branch`+`expected-commit`, and warn if you specify `tag` _without_ `expected-commit`, and verify when you specify `tag`+`expected-commit`.

What do you think?

tuananh: even though we know branch is a moving target, i kind of want to warn here because in the end, what we want is reproduciblility right? having a pinpoint git commit sha is still better if we use this pipeline in Wolfi.

kaniini (approved):

PR #288, opened 2023-02-14
package: ensure PT_INTERP is always added as an explicit dependency

### LINKED ISSUE #281: Binaries with DT_INTERP on /lib/ld-linux.so.2 require glibc
Please see https://github.com/wolfi-dev/os/pull/530

This change was necessary, even though the binaries are built using `CGO_ENABLED=0`

In dependency scanning: dependency scanning begins: https://github.com/chainguard-dev/melange/blob/main/pkg/build/package.go#L319

we want to pull out the PT_INTERP

possibly enabled by https://github.com/saferwall/elf

cc @kaniini

kaniini:
previously we depended on libc appearing in the DT_NEEDED list to pull in the ELF interpreter.

closes #281.

### DIFF
--- a/pkg/build/package.go
+++ b/pkg/build/package.go
@@ -295,6 +295,26 @@ func generateCmdProviders(pc *PackageContext, generated *Dependencies) error {
 	return nil
 }
 
+// findInterpreter looks for the PT_INTERP header and extracts the interpreter so that it
+// may be used as a dependency.
+func findInterpreter(bin *elf.File) (string, error) {
+	for _, prog := range bin.Progs {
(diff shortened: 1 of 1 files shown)


kaniini:
Hmm, this solution doesn't work for musl though, where the interp is strictly a symlink.

rawlingsj (approved):

imjasonh (approved):
(not blocking) Is there a test or example we can add to prevent regression?

kaniini:
> (not blocking) Is there a test or example we can add to prevent regression?

I'll try to come up with a very small go binary which works for this.

PR #297, opened 2023-02-23
add configurable package building options and if-conditionals for subpackages

kaniini:
Adds two new features:

**Build options**

These allow patching the build configuration with configurable build variants.  Packages can be added, or removed from the build dependencies, and build variables can be changed.  Additional variables reflecting the build option name are reflected in conditionals.

**Subpackage conditionals**

These control whether or not specific subpackages get built.  They are intended to be used with the `${{options.foo.enabled}}` variables, for example:

```yaml
subpackages:
  - if: ${{options.foo.enabled}} == 'true'
    name: 'foo'
    pipeline:
       ...
```

### DIFF
--- a/examples/options.yaml
+++ b/examples/options.yaml
@@ -1,0 +1,128 @@
+package:
+  name: curl
+  version: 7.87.0
+  epoch: 3
+  description: "URL retrieval utility and library"
+  target-architecture:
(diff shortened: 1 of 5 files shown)


jdolitsky (approved):
👍

PR #299, opened 2023-02-25
build: package: dereference symlinks across packages and read the real DT_SONAME instead of guessing

kaniini:


pkg/build/package.go
```diff
@@ -319,6 +319,37 @@ func findInterpreter(bin *elf.File) (string, error) {
...
+		targetPackageNames = append(targetPackageNames, subPkg.Name)
+	}
+
+	for _, pkgName := range targetPackageNames {
+		basePath := filepath.Join(pc.Context.WorkspaceDir, "melange-out", pkgName)
+
+		for _, libDir := range libDirs {
+			testPath := filepath.Join(basePath, libDir, realPath)
+
+			if _, err := os.Stat(testPath); err == nil {
```
imjasonh, line 344: This seems to take the first matching package. Is there any case where multiple subpackages might produce a candidate for the target lib?

I normally think of subpackages as unordered unless one depends on another, and this feels like a relatively unusual/subtle way for subpackages to depend on another in an ordered way.

Maybe we should fail if multiple candidates are found instead of taking the first?

imjasonh (commented):
This feels like an interesting enough change to warrant a test. I'm not sure I'd feel comfortable modifying this code a year from now without one, and the first thing I'd do is probably try to reverse engineer a test, poorly.

Even the way this surfaces for ruby 3.2 seems hard to replicate meaningfully.

Can you save poor Future Jason the headache and write one while we have all this valuable context? 🙏

found-it (approved):
LGTM! Great work

found-it:
Sorry @imjasonh looks like we reviewed at the same time. Your comments are great, I'll let @kaniini respond before kicking off an update of the ruby 3.2 packages

pkg/build/package.go
tuananh: or do it in a more predictable way? sort first then take first at least?

kaniini: It does not matter what order the packages are checked in.  All we are looking for is one package which has the target library, so we can learn the actual SONAME.  There would not be a scenario where building multiple variations of the same library results in a different SONAME, but not also a different symlink pattern.

kaniini:
We can rebuild the ruby 3.2 packages once the sdk image is regenerated.

PR #305, opened 2023-03-03
WIP: bump apko dep, handle breaking changes

imjasonh:
This changes deps on apko's `pkg/fs` to use `pkg/vfs` instead.

apko no longer seems to have `UseProot` -- do we no longer use proot?

cc @deitch


deitch:
> apko no longer seems to have UseProot -- do we no longer use proot?

Nope. Since apko does everything natively, no executing out, it doesn't need it.

> use pkg/vfs instead

I am not sure that actually is used inside apko either. Have a look in [pkg/apk/impl/fs](https://github.com/chainguard-dev/apko/tree/main/pkg/apk/impl/fs). There is both an interface definition and some implementations.

I had thought about putting them in `pkg/vfs/`, but we have been talking about upstreaming everything in `apk/impl/`, or, at the very least, moving it into its own repository, so I wanted it to have no external dependencies on anything in `chainguard-dev/apko`. I might have missed a thing or two, but overall it is standalone.

imjasonh:
Do you happen to know why the sshfs.yaml example would fail using vfs?

```
Error: failed to build package: unable to emit package: unable to write data tarball: writing TAR archive failed: readlink not supported by this fs: path (usr/sbin/mount.fuse.sshfs)
```

https://github.com/chainguard-dev/melange/actions/runs/4344653218/jobs/7588266733#step:5:493

deitch:
Usually that happens because whatever `fs.FS` implementation was used, it does not implement `ReadlinkFS`. I am not quite sure what happened here, except maybe that it uses `pkg/vfs` and that isn't fully compatible with the new requirements? Those usually are implemented via `apko/impl/fs/` (one of the filesystems in there).

FWIW,I have an implementation of melange using the latest apko (and supports limavm and docker, etc.). It _almost_ works. I ran into the issue with the mounted directory, where it created problems.

The remaining step is to be able to treat all build runners as remote: docker, lima, bwrap, even Kubernetes. I was almost there and then ended up digging much more deeply into apko compatibility, so it has been on hold.

If you need this urgently, I can go back into that and get it in place, see if it solves these issues. If this can wait a bit, I can finish the apko details, this is next on my list (and mostly done).

imjasonh:
> If you need this urgently, I can go back into that and get it in place, see if it solves these issues. If this can wait a bit, I can finish the apko details, this is next on my list (and mostly done).

I don't need this urgently at all, I was mainly hitting some version compatibility stuff in wolfictl, which depends on both apko and melange, and noticed that melange wasn't depending on the latest-and-greated apko yet. I assumed the bump would be easy, and here we are 🙃 

I'm fine waiting for you if you already have a good idea how this should work. Thanks for your great answers so far, and for understanding all of this better than I do 😆

PR #306, opened 2023-03-04
remove self-provided dependencies from the runtime dependency set

kaniini:
Although apk can handle self-dependencies just fine, it warns about them when using "apk dot".

Noticed while debugging a dependency issue in Wolfi.

### DIFF
--- a/pkg/build/package.go
+++ b/pkg/build/package.go
@@ -520,6 +520,28 @@ func (dep *Dependencies) Summarize(logger *log.Logger) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


imjasonh (approved):
Does it make sense to make this an error instead? And/or make it a lint failure in Wolfi at least? Silently fixing this for me is great, though. 👍

kaniini:
Nah, if we make it an error, the automatic dependency scanning will become a lot more complex.  Best to just silently drop the duplicate entries, which matches what `abuild` does.

PR #318, opened 2023-03-09
add convert subcommand

strongjz:
Open sourcing our internal mconvert tools for managing all our packages in [wolfi-dev/os](https://github.com/wolfi-dev/os) Started by @rawlingsj :partying_face: and extend by the rest of the team, python @strongjz gems @found-it


pkg/python/python.go
```diff
@@ -0,0 +1,359 @@
...
+	// transitive dependencies.
+	ToCheck []string
+}
+
+// TODO: Why isn't this pulled in from melange/build?
+type ImageContents struct {
+	Repositories []string `yaml:"repositories,omitempty"`
+	Keyring      []string `yaml:"keyring,omitempty"`
+	Packages     []string `yaml:"packages,omitempty"`
+}
```
found-it, line 79: This can likely be removed since it's in the same repository now

pkg/python/python.go
```diff
@@ -0,0 +1,359 @@
...
+
+	context := PythonContext{
+		PackageName: packageName,
+		Logger:      log.New(log.Writer(), "convert:python: ", log.LstdFlags|log.Lmsgprefix),
+		ToGenerate:  make(map[string]Package),
+	}
+	return context, nil
+}
+
+// Generate is the entrypoint to generate a ruby gem melange file. It handles
```
found-it, line 96: Looks like the python comments are still referring to ruby

found-it (changes requested):

pkg/python/python.go
strongjz: its still apko_types.ImageConfiguration if im not mistaken

found-it: Ah yes, I think that TODO is wrong. I do remember for some reason I had issues importing ImageContents from apko types but I was likely doing something wrong and settled with this workaround

pkg/convert/wolfios/wolfi-os.go
```diff
@@ -0,0 +1,148 @@
+package wolfios
```
found-it, line 1: Can we break this package out from under convert/apkbuild like we did with manifest and http?

pkg/convert/convert_test.go
```diff
@@ -0,0 +1,331 @@
+package convert
```
found-it, line 1: Can we rename this package to `apkbuild` to make it consistent with python and gem?

found-it (changes requested):

pkg/convert/convert_test.go
strongjz: it is an apkbuild file; if anything we should probably change python to pypi ?

pkg/convert/wolfios/wolfi-os.go
strongjz: done

tuananh:
hoooray !!!!

kaniini:
It would be nice to put all of the convert-related packages under pkg/convert as subpackages.

strongjz:
> 

So I understand, move python, apkbuild, and gem in pkg/convert/X ?

kaniini:
> So I understand, move python, apkbuild, and gem in pkg/convert/X ?

Yes!

kaniini (approved):

found-it (approved):
Looks good to me!

PR #320, opened 2023-03-14
container: bubblewrap runner: use --new-session to mitigate CVE-2017-5226

kaniini:
Without it, it is possible to escape the sandbox via TIOCSTI ioctls on the session PTY.

Related: https://github.com/containers/bubblewrap/issues/555
Related: https://github.com/containers/bubblewrap/issues/142
Related: https://news.ycombinator.com/item?id=30825088

### DIFF
--- a/pkg/container/bubblewrap_runner.go
+++ b/pkg/container/bubblewrap_runner.go
@@ -40,7 +40,8 @@ func (bw *BWRunner) Run(cfg *Config, args ...string) error {
 		"--dev", "/dev",
 		"--proc", "/proc",
 		"--chdir", "/home/build",
-		"--clearenv")
+		"--clearenv",
+		"--new-session")
 
 	if !cfg.Capabilities.Networking {
 		baseargs = append(baseargs, "--unshare-net")


rawlingsj (approved):

hartwork:
Hi, good move!

I see that commit 07cd62ea469a1d05bdaa628354fbe13158ab2f42 is not part of release 0.2.0 and that 0.2.0 is [packaged in Arch Linux and derivatives](https://repology.org/project/melange/versions). My vote for considering a new release and a CVE for this. It only takes filling this form: https://cveform.mitre.org/ .

kaniini:
Our invocation of bubblewrap does not pass through the terminal file descriptor, so we do not believe there is any practical exposure to CVE-2017-5226.  We would be likely to dispute any CVE filed for that reason.

hartwork:
To be sure I understand: if that is true, why would you want to pass `--new-session` then? How is `exec.Command` not passing the controlling terminal to bwrap? Please help me understand what I may be missing.

kaniini:
> To be sure I understand: if that is true, why would you want to pass --new-session then?

We *do* want the `setsid` effects for other reasons.  And it is better to have it anyway, just in case we begin to use PTYs in the future instead of pipes.

> How is `exec.Command` not passing the controlling terminal to bwrap? Please help me understand what I may be missing.

`exec.Command` just creates an execution object.  The actual execution happens here, after stdio redirection is set up to use pipes: https://github.com/chainguard-dev/melange/blob/main/pkg/container/runner.go#L49

hartwork:
> We _do_ want the `setsid` effects for other reasons.

Could you elaborate?

The title "to mitigate CVE-2017-5226" communicates a different picture. I'm confused.

PR #334, opened 2023-03-21
Try to fix a strange index generation bug.

dlorenc:
I've seen this occur a few times in prod, and it's just a guess, but I think new package versions aren't making it into the index when run in merge index mode.

I attributed it to a race condition (which I think happens too), but I think this is also an issue.


kaniini (approved):

kaniini:
Nice catch!

PR #337, opened 2023-03-22
add Update struct for identifying how a melange package can be updated

### LINKED ISSUE #336: Proposing new melange struct to help auto updating packages
Today the [Wolfi](https://github.com/wolfi-dev/os) project uses a custom [built bot](https://github.com/wolfi-dev/wolfictl/blob/main/docs/update.md) to detect new upstream package versions and automatically create a PR for the version bump, epoch reset and new expected sha.  Here's an example https://github.com/wolfi-dev/os/pull/801.

Today, this only covers a subset of packages in the Wolfi package repository, this is for two reasons
1. it was experimental to validate the approach 
2. a mapping file was used to map melange packages to APIs like [release monitor](https://release-monitoring.org/) and GitHub via its GraphQL API.  Example file https://github.com/wolfi-dev/wolfi-update-mapper/blob/main/DATA.md.  When new melange packages are added to Wolfi typically this file is not updated + not all use cases are covered by the current mapping columns.

Now that we have proved out the approach and learned more use cases it would be ideal if we added a section to the melange config that covered the mapping data.  This way we can describe how automation keeps a package up to date when contributing packages.  This proposal is generic and not tied to the wolfictl bot implementation, it could also be useful for users to help manually update future versions.

Proposed new optional `update:` yaml would look like this for describing a package covered by release monitor:

```yaml
package:
  name: bison
  version: 3.8.2
  epoch: 0
...
update:
  enabled: true # provide a flag to easily prevent a package from receiving auto update PRs
  shared: false # indicate that an update to this package requires an epoch bump of downstream dependencies, e.g. golang, java
  release-monitor: 
    identifier: 193  # Mandatory, ID number for release monitor
```

and for a package that lives in GitHub
```yaml
package:
  name: cosign
  version: 2.0.0
  epoch: 0
....
update:
  enabled: true # provide a flag to easily toggle a package from receiving auto update PRs
  shared: false # indicate that an update to this package requires an epoch bump of downstream dependencies, e.g. golang, java
  github: # alternative today is `release_monitor:`
    identifier: sigstore/cosign # Mandatory, org/repo for github
    strip-prefix: v # Optional, if the version obtained from the update service contains a prefix which needs to be stripped when updating the melange package
    use-tag: true # Optional, override the default of using a GitHub release to identify related tag to fetch.  Not all projects use GitHub releases but just use tags
    tag-filter: foo # Optional, filter to apply when searching tags on a GitHub repository, some repos maintain a mixture of tags for different major versions for example
```

If / when we want to support more APIs / services to query for new versions we can add more sections.  I did ponder instead of having a strict `release-monitor:` or `github:` type we could use an enum which means no changes would be required to melange to support future APIs.  However, fields like identifier and others may not mean the same for different services so being strict allows users to be clearer and automate validation.

I think this is a minimum set of fields we can use to keep a large number of malange packages up-to-date.

rawlingsj:
fixes #336

### DIFF
--- a/docs/UPDATE.md
+++ b/docs/UPDATE.md
@@ -1,0 +1,52 @@
+# Update
+
+`update:` provides information on how a melange package can be updated either manually or using automation for example as the [Wolfi project does](https://github.com/wolfi-dev/wolfictl/blob/main/docs/update.md).
+
+__IMPORTANT:__ Adding update configuration does not mean melange package will be kept up to date, it is a way to describe "how"
+it can be updated.
+
+There are currently two ways to describe where to search for latest versions of a package.
+
+ 1. `release-monitor:` to query https://release-monitoring.org/
+ 2. `github:` to query https://github.com via it's graphql API
+
+## Release Monitor
+
+This is a service that matched a package with an identifier.  The melange config should describe the Identifier in https://release-monitoring.org/
+
+```yaml
+package:
+  name: alsa-lib
+  version: 1.2.8
+
+...
+
+update:
+  enabled: true # provide a flag to easily prevent a package from receiving auto update PRs
+  shared: false # indicate that an update to this package requires an epoch bump of downstream dependencies, e.g. golang, java
+  release-monitor:
+    identifier: 38 # Mandatory, ID number for release monitor
+```
+
+## GitHub
+
+This assumes you are using the graphql API and the Identifier matches the organisation/repositoryname
+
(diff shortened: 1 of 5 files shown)


rawlingsj:
If this looks good I'll add update this PR with docs to describe the new `update` fields.

kaniini:
LGTM.

kaniini (approved):

PR #347, opened 2023-03-28
mutate fetch URL during renovate bump

rawlingsj:
this also moves the renovate code into the build package to avoid cyclic imports, as renovate needs the pipeline mutate fuctions however build/cache.go already uses renovate funtions


Elizafox:
I subsumed this in #586.

PR #349, opened 2023-03-29
Refactor --cache-dir and separate preloading concerns into --cache-source

### LINKED ISSUE #329: --cache-dir assumed to be both source and destination
If you use `--cache-dir gs://wolfi-sources/`, for example, you will wind up with a path: `gs:/wolfi-sources/...` in the current working directory.

The cache prepopulation should be split out into its own option, `--cache-src`.

kaniini:
Previously `--cache-dir` was assumed to be both a GCS bucket and a destination directory.  This made the feature unusable on Darwin hosts, which rejected the GCS bucket name as a valid filename on disk.

Fixes #329.

### DIFF
--- a/pkg/cli/build.go
+++ b/pkg/cli/build.go
@@ -107,7 +107,7 @@ func Build() *cobra.Command {
 	cmd.Flags().StringVar(&workspaceDir, "workspace-dir", "", "directory used for the workspace at /home/build")
 	cmd.Flags().StringVar(&pipelineDir, "pipeline-dir", "", "directory used to extend defined built-in pipelines")
 	cmd.Flags().StringVar(&sourceDir, "source-dir", "", "directory used for included sources")
-	cmd.Flags().StringVar(&cacheDir, "cache-dir", "/var/cache/melange", "directory used for cached inputs")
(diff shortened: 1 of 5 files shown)


pkg/cli/build.go
```diff
@@ -107,7 +109,8 @@ func Build() *cobra.Command {
 	cmd.Flags().StringVar(&workspaceDir, "workspace-dir", "", "directory used for the workspace at /home/build")
 	cmd.Flags().StringVar(&pipelineDir, "pipeline-dir", "", "directory used to extend defined built-in pipelines")
 	cmd.Flags().StringVar(&sourceDir, "source-dir", "", "directory used for included sources")
-	cmd.Flags().StringVar(&cacheDir, "cache-dir", "/var/cache/melange", "directory used for cached inputs")
+	cmd.Flags().StringVar(&cacheDir, "cache-dir", "./melange-cache/", "directory used for cached inputs")
+	cmd.Flags().StringVar(&cacheSource, "cache-source", "", "directory or bucket used for preloading the cache")
```
luhring, line 113: nit: (or for later...) — update help text and/or docs to specify this is for `fetch` only

luhring (approved):
Nice! 🎉

PR #350, opened 2023-03-29
add query and package-version commands

deitch:
Since melange already is parsing yaml files, might as well use it to report on information.

### DIFF
--- a/pkg/cli/query.go
+++ b/pkg/cli/query.go
@@ -1,0 +1,59 @@
+// Copyright 2023 Chainguard, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


jdolitsky (approved):
neat!

kaniini (approved):

pkg/cli/query.go
```diff
@@ -0,0 +1,59 @@
...
+	"context"
+	"fmt"
+	"os"
+	"text/template"
+
+	"chainguard.dev/melange/pkg/build"
+	"github.com/spf13/cobra"
+)
+
+func Query() *cobra.Command {
```
imjasonh, line 27: I like this!

I've been using [`yq`](https://github.com/mikefarah/yq) a bit for this kind of thing, which doesn't support Go templates, but it does have a pretty extensive query language, like `jq`, and can even update things in-place (if you accept its formatting decisions).

deitch: Credit to @kaniini on the `query` idea.

PR #351, opened 2023-03-29
directly parse configuration for query

deitch:
Previously we relied on `build.New()` to get a `BuildContext`, which has `bc.Configuration`, which we passed to the template.

But that assumed a lot of things about the build environment, since it is a build context. Those things sometimes break. In any case `build.ParseConfiguration()` is exported.

This changes it to just call `ParseConfiguration` directly.

### DIFF
--- a/pkg/cli/query.go
+++ b/pkg/cli/query.go
@@ -33,16 +33,15 @@ func Query() *cobra.Command {
 		Example: `  melange query config.yaml "{{ .Package.Name }}-{{ .Package.Version }}-{{ .Package.Epoch }}"`,
 		Args:    cobra.MinimumNArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


pkg/cli/query.go
```diff
@@ -33,24 +33,23 @@ func Query() *cobra.Command {
...
 		},
 	}
 
 	return cmd
 }
 
-func QueryCmd(ctx context.Context, pattern string, opts ...build.Option) error {
-
-	bc, err := build.New(opts...)
+func QueryCmd(ctx context.Context, configFile, pattern string) error {
```
imjasonh, line 43: This method doesn't need to be exported.

imjasonh (approved):

pkg/cli/query.go
deitch: True, but every other command in `cli/*.go` is, so I just followed the pattern. If we want to make those private, we should do it across the board.

PR #367, opened 2023-04-03
melange bump: optional flag to modify git-checkout pipeline expected-…

rawlingsj:
…commit value


pkg/renovate/bump/bump_test.go
```diff
@@ -61,6 +62,47 @@ func TestBump_versions(t *testing.T) {
...
+func TestBump_withExpectedCommit(t *testing.T) {
+
+	dir := t.TempDir()
+
+	tests := []struct {
+		name           string
+		newVersion     string
+		expectedCommit string
+	}{
+		{name: "expected_commit.yaml", newVersion: "7.0.1", expectedCommit: "dbd7bc96fd6cd383b8e895dc4a928d808541bb17"},
```
imjasonh, line 74: Can this just be a single test that starts with what's currently at like 79? If there's only one case here, it's probably not worth the table-driven scaffolding.

pkg/renovate/bump/testdata/expected_commit.yaml
```diff
@@ -0,0 +1,11 @@
+package:
+  name: cheese
+  version: 6.8
+  epoch: 2
+  description: "a cheesy library"
+
+pipeline:
+  - uses: git-checkout
+    with:
+      repository: cheese/crisps
```
imjasonh, line 10: mmmm 😋

pkg/renovate/bump/bump_test.go
rawlingsj: the test was generated from the IDE and does make it easier for folks adding new cases but TBH I don't mind either way.  Happy to switch it to a single test if you prefer.  FWIW as this is merged I have opened a second PR to discuss as I did noticed that the test data meant the test didn't actually verify the new sha was correct.  https://github.com/chainguard-dev/melange/pull/368

PR #395, opened 2023-04-19
Fix parse configuration's use of abstract filesystems

luhring:
This PR fixes two issues:

1. A test for the bump logic was using `assert.NoError` instead of `require.NoError`, so it panicked when the test proceeded to use a nil object after an error case.

2. `ParseConfiguration` wasn't correctly handling abstract filesystems. In general, when using an `fs.FS`, the code should be able to trust that the config file path is meant for that filesystem, but it was using `filepath.Base` in anticipation of receiving an absolute filepath for the actual OS filesystem. This PR detects if the caller is setting their own `fs.FS` and uses the given path "as-is" if so. In the future, we should adjust `ParseConfiguration` to always use a provided `fs.FS` and thus always trust that the given config path is correct as-is.

These issues were caught via tests in `wolfictl`.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -765,15 +765,13 @@ type configOptions struct {
 
 // include reconciles all given opts into the receiver variable, such that it is
 // ready to use for config parsing.
-func (options *configOptions) include(dirPath string, opts ...ConfigurationParsingOption) {
+func (options *configOptions) include(opts ...ConfigurationParsingOption) {
 	for _, fn := range opts {
 		fn(options)
 	}
+
 	if options.logger == nil {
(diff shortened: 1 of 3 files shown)


jdolitsky (approved):

luhring:
We believe these E2E test failures to be _unrelated to the changes in this PR_, caused instead by [this issue](https://github.com/chainguard-dev/apko/issues/627), and as proven by [this dummy PR](https://github.com/chainguard-dev/melange/pull/396).

jdolitsky (approved):

imjasonh (approved):

kaniini (approved):

PR #406, opened 2023-04-28
add `ignore-regex-patterns` update config to indicate you want to ign…

rawlingsj:
…ore string patterns that match an upstream version


rawlingsj:
Why use a slice of regex patterns rather than a single string, I hear you ask... maintainability.  Regex are not a great UX and modifying existing regex is daunting when there's no tests.  Using a regex in YAML doesn't have tests so it is easier for maintainers and contributors to add a new pattern if and when needed.

amouat (approved):

PR #425, opened 2023-05-04
build: package: skip SONAME analysis when ELF interpreter setting is present

kaniini:
End-user programs are the only type of ELF object where it would make sense for an ELF interpreter to be set on the on the ELF object.  Accordingly, treat any object which has an ELF interpreter set on it as an end-user program and skip SONAME analysis for them.

Certain end-user programs such as the ones included with OpenJDK 8 have a bogus SONAME (`lib.so`) set on them, which otherwise should not be there.

### DIFF
--- a/pkg/build/package.go
+++ b/pkg/build/package.go
@@ -464,7 +464,10 @@ func generateSharedObjectNameDeps(pc *PackageContext, generated *Dependencies) e
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


imjasonh (approved):

luhring (approved):
Wishlist item: test 😺

PR #428, opened 2023-05-05
Add an optional "deps" paramter to the go/build pipeline.

dlorenc:
We can use this to hoist up some of the imperative "go get'ing" we do to fix CVEs into a structured format.

### DIFF
--- a/docs/PIPELINES-GO.md
+++ b/docs/PIPELINES-GO.md
@@ -97,6 +97,10 @@ you can define the following values:
   ldflags:
     description:
       List of [pattern=]arg to pass to the go compiler with -ldflags
+
+  deps:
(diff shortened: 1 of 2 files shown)


dlorenc:
I've tested this against the existing pipelines that use go build and they don't break!

docs/PIPELINES-GO.md
```diff
@@ -97,6 +97,10 @@ you can define the following values:
   ldflags:
     description:
       List of [pattern=]arg to pass to the go compiler with -ldflags
+
+  deps:
```
imjasonh, line 101: Maybe `update_deps`?

imjasonh (approved):

PR #446, opened 2023-05-13
add extra logging when runner fails to TestUsability

rawlingsj:
a recent change made bubblewrap the default when running melange build natively on mac, as a result builds fail to start...

```
2023/05/13 11:29:15 error during command execution: unable to run container
```

This change adds a warning when the check fails and offers the user help of how to fix.  For example now I need to run `melange build --runner docker`

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -583,7 +583,7 @@ func New(opts ...Option) (*Context, error) {
 
 	// Check that we actually can run things in containers.
 	if !runner.TestUsability() {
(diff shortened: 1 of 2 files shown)


imjasonh (approved):

kaniini:
hmm, we should fix it.  bubblewrap shouldn't ever be the default on mac

deitch:
You're right it shouldn't. I'll fix that. 

I originally had an auto detect in there, removed it for some reason lost in the mists of time. Will get it back.

PR #464, opened 2023-05-23
add a kubernetes pod runner

joshrwolf:
adds a `kubernetes` runner.

this runner is unique in 2 meaningful ways:

- it's a remote runner, and uses `WorkspaceTar` to retrieve the workspace post build
  - this involved re-introducing `WorkspaceTar` to the build flow, and nooping the existing local builders (`docker` and `bwrap`) implementation of `WorkspaceTar`
- it's a more configurable runner, and relies on it's own config init semantics. this PR keeps things simple and only loads the config from a `.melange.k8s.yaml` file, but future PRs will expand this with more loaders


pkg/container/kubernetes_runner.go
```diff
@@ -0,0 +1,407 @@
...
+
+func KubernetesRunner(logger log.Logger) Runner {
+	// TODO: Parse this from the config
+	repo, _ := name.NewRepository("ttl.sh/melange")
+	return &k8s{
+		logger:    logger,
+		prefix:    "melange-",
+		namespace: "default",
+		repo:      repo,
+	}
```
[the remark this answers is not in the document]
kaniini: I agree that a configuration file for the cluster makes the most sense.

examples/k8s-runner.yaml
```diff
@@ -0,0 +1,42 @@
+---
+# SPDX-FileCopyrightText: 2022 Chainguard, Inc
```
kaniini, line 2:
```suggestion
# SPDX-FileCopyrightText: 2023 Chainguard, Inc
```

pkg/container/kubernetes_runner.go
```diff
@@ -0,0 +1,436 @@
+package container
```
kaniini, line 1: Needs the legal text at the top.

pkg/container/kubernetes_runner.go
```diff
@@ -0,0 +1,436 @@
...
+					Requests: corev1.ResourceList{
+						corev1.ResourceCPU:    resource.Quantity(k.CPU),
+						corev1.ResourceMemory: resource.Quantity(k.Memory),
+					},
+				},
+			}},
+		},
+	}
+
+	if cfg.Arch.ToAPK() == "aarch64" {
```
kaniini, line 182: Why not set `kubernetes.io/arch` unconditionally?  `kubernetes.io/arch: amd64` is a no-op.

Something like:

```go
builder.Spec.NodeSelector = map[string]string{
  "kubernetes.io/arch": cfg.Arch.String(),
  "cloud.google.com/compute-class": "Scale-Out",
}
```

cfg.Arch default string representation is always the same as `runtime.GOARCH`, which matches the kubernetes selector.  This futureproofs us for when we eventually add `riscv64`, `ppc64le` or `s390x`.

pkg/build/build.go
```diff
@@ -1690,6 +1699,46 @@ func (ctx *Context) BuildPackage() error {
 		return err
 	}
 
+	// TODO: Should this go here?
```
kaniini, line 1702: I think this should be factored out to a new function, perhaps `WorkspaceCopyBack` or similar.

pkg/build/build.go
```diff
@@ -294,6 +296,13 @@ type Configuration struct {
 	VarTransforms []VarTransforms `yaml:"var-transforms,omitempty"`
 	// Optional: Deviations to the build
 	Options map[string]BuildOption `yaml:"options,omitempty"`
+
+	// Resources specifies the CPU and RAM resources used to build the package.
+	// This is only relevant for the kubernetes runner
```
kaniini, line 301: It might be useful for `lima` runner too.

imjasonh:
I'd like to get this in, even if we don't use it in wolfi yet, just so I can use it myself and put some miles on it. Is there anything major blocking this? An e2e test that uses KinD could also be pretty valuable, to guard against regressions.

joshrwolf:
> I'd like to get this in, even if we don't use it in wolfi yet, just so I can use it myself and put some miles on it. Is there anything major blocking this? An e2e test that uses KinD could also be pretty valuable, to guard against regressions.

I spent some cycles cleaning this up end of last week, I'll update this PR and bring it out of draft today. I'd also like to get people playing with it, and we can make it pretty and iterate before using it in wolfi

pkg/container/kubernetes_runner.go
```diff
@@ -0,0 +1,527 @@
...
+	Provider    string            `json:"provider" yaml:"provider"`
+	Repo        string            `json:"repo" yaml:"repo"`
+	Namespace   string            `json:"namespace" yaml:"namespace"`
+	Annotations map[string]string `json:"annotations,omitempty" yaml:"annotations,omitempty"`
+	Labels      map[string]string `json:"labels,omitempty" yaml:"labels,omitempty"`
+
+	StartTimeout time.Duration `json:"startTimeout" yaml:"startTimeout"`
+	BuildTimeout time.Duration `json:"buildTimeout" yaml:"buildTimeout"`
+
+	PodTemplate *KubernetesRunnerConfigPodTemplate `json:"podTemplate,omitempty" yaml:"podTemplate,omitempty" ignored:"true"`
```
imjasonh, line 346: Can we just use the real `corev1.PodTemplate` here?

pkg/container/kubernetes_runner.go
```diff
@@ -0,0 +1,527 @@
...
+
+	// A "burstable" QOS is really the only thing that makes sense for ephemeral builder pods
+	// Only set the requests, and not the limits
+	Resources *KubernetesRunnerConfigResources `json:"resources,omitempty" yaml:"resources,omitempty"`
```
imjasonh, line 350: https://pkg.go.dev/k8s.io/api/core/v1#ResourceList ?

This lets us request GPUs and other exotic platform-specific hardware, if we ever have a need to. And more importantly, it gives us less code we own :)

imjasonh (approved):
Let's gooooooo!

pkg/container/kubernetes_runner.go
joshrwolf: I started that way, then decided it's probably too much rope for end users to strangle themselves with.

granted we're the end user here, so that might be okay?

imjasonh: I don't feel strongly either way, and it's not a one-way door. Same with the other comment about `ResourceList`.

imjasonh (approved):

imjasonh (dismissed):

pkg/container/kubernetes_runner.go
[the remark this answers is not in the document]
joshrwolf: for this PR, I kept config loading to a `.melange.k8s.yaml` file, but left the option open for expanding this into more loaders in the future

pkg/container/kubernetes_runner.go
```diff
@@ -0,0 +1,518 @@
...
+	RuntimeClassName *string           `json:"runtimeClassName,omitempty" yaml:"runtimeClassName,omitempty"`
+	Volumes          []corev1.Volume   `json:"volumes,omitempty" yaml:"volumes,omitempty"`
+}
+
+// NewKubernetesConfig returns a default Kubernetes runner config setup
+func NewKubernetesConfig(opt ...KubernetesRunnerConfigOptions) *KubernetesRunnerConfig {
+	cfg := &KubernetesRunnerConfig{
+		Provider:     "generic",
+		Namespace:    "default",
+		Repo:         "ttl.sh/melange",
```
jonjohnsonjr, line 368: This pushes to `ttl.sh` by default? I don't think we should do that.

imjasonh: +1, but I think we can fix this in a followup PR, where the repo must be configured somehow. This is sort of like requiring `KO_DOCKER_REPO` to be set basically.

joshrwolf: I've got it staged as a follow up, right now it's overlayed as `defaults -> .melange.k8s.yaml -> package config -> env var`, but this can be changed.

I'd like to keep the config handling as a separate PR though, as it somewhat detracts from the basic functionality of this.

However if we're worried about silently defaulting to pushing to a public repo (a valid concern), I'm ok with erroring out when no repo is specified for now!

imjasonh (approved):

joshrwolf:
> @imjasonh is there plan to remove `ttl.sh`? it looks kinda sketchy here.

yes, we don't currently use the k8s runner anywhere serious yet, but as we plan on rolling it out more broadly we'll definitely force a default instead of falling back to `ttl.sh`

tuananh:
> > @imjasonh is there plan to remove `ttl.sh`? it looks kinda sketchy here.
> 
> yes, we don't currently use the k8s runner anywhere serious yet, but as we plan on rolling it out more broadly we'll definitely force a default instead of falling back to `ttl.sh`

is there anything specific about that melange image on ttl.sh? why can't we just use one from chainguard?

joshrwolf:
that config is the location of where the build workspaces are _pushed_ by the client and fetched by the builder. there's no significance to `ttl.sh/melange` other than the repo where things are pushed, it just simplified the dev flow a bit during these early stages

siennathesane:
> we don't currently use the k8s runner anywhere serious yet

I'm currently using the k8s runner in both Rancher Desktop & AWS EKS, for an extra bullet point. I absolutely agree on making `ttl.sh` configurable, even if it's just to point to your own internal `ttl.sh` instance.

PR #495, opened 2023-06-05
Wrap Glob in a sync.Once

jonjohnsonjr:
Every time we called BuildFlavor, we were doing a ton of filesystem work. This caches the result of that work. I have not investigated if we can make that work faster.


kaniini (dismissed):

jonjohnsonjr:
Ah, this won't work because we're copying a mutex around? 🤔 

Will have to look at this again later.

jonjohnsonjr:
Changed tack on this entirely. On my machine, this helps `wolfictl text` out a lot, bringing `dag.NewPackages` down from ~3s to ~200ms.

kaniini (approved):

PR #529, opened 2023-07-03
Add otel spans

jonjohnsonjr:
Given this and these:
* https://github.com/chainguard-dev/apko/pull/783
* https://github.com/chainguard-dev/go-apk/pull/72

We can do nice things like see where `melange build` latency is being spent across both melange and apko.

We spend a good amount of time in BuildGuest (working on that: https://github.com/chainguard-dev/apko/issues/781):
image

Loading the image into docker takes longer than I'd expect:

image

But this will be particularly useful when we hit some pathological case where we have unexpected latency.


jonjohnsonjr:
Once we merge the go-apk and apko PRs, I'll need to bump melange's dep on them and fix some missing context parameters to make sure we have a full chain of context propagation.

imjasonh (approved):

PR #537, opened 2023-07-05
docs: explain how build cache works practically

luhring:
This is just a docs update, in hopes of helping future Melangers understand how the build cache can help them without having to learn the hard way. 😃


joshrwolf:
this is awesome, the one __nit__ I would make is a comment on how `--cache-dir` is different than `--apk-cache-dir`

joshrwolf (approved):

luhring:
> this is awesome, the one **nit** I would make is a comment on how `--cache-dir` is different than `--apk-cache-dir` so folks don't get confused

Makes total sense!

_[Starts figuring out what --apk-cache-dir does.... 😬]_

PR #539, opened 2023-07-07
Change git-checkout depth default to 1

luhring:
Deep cloning can cost a lot of time on a per-dev-cycle basis. This changes the default depth to `1` — and it can still be adjusted to any number on a per-build-config basis as needed.

### DIFF
--- a/pkg/build/pipelines/git-checkout.yaml
+++ b/pkg/build/pipelines/git-checkout.yaml
@@ -18,7 +18,7 @@ inputs:
   depth:
     description: |
       The depth to use when cloning.
-    default: 50
+    default: 1
 
   branch:
     description: |


imjasonh:
I went looking to see if anything sets `depth:` today, and the only ones that do, [`argon2`](https://github.com/wolfi-dev/os/blob/38fa4e538a1f385bc85e180a6b812542dd5807de/argon2.yaml#L19) and [`ffmpeg`](https://github.com/wolfi-dev/os/blob/38fa4e538a1f385bc85e180a6b812542dd5807de/ffmpeg.yaml), use it to set `depth: 1`.

We might find some other builds that fail because they'd assumed a depth >1 in their build, and we won't find them until we bump their epochs or do a world-rebuild. That seems pretty unlikely though, and the fix is to just change `depth: 50` again (and remember that's a thing we might want to do).

+1

imjasonh (approved):

PR #542, opened 2023-07-11
pull in apko with fix for blank SOURCE_DATE_EPOCH

deitch:
Pulls in the fix from apko from [this PR](https://github.com/chainguard-dev/apko/pull/803)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,9 +3,9 @@ module chainguard.dev/melange
 go 1.20
 
 require (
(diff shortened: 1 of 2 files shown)


deitch:
Interesting. The failure is not on this branch, but on main as well.

kaniini:
I suspect it is due to a change on PyPI

kaniini (approved):

Issue #541, opened 2023-07-10
Lima runner build fails on tar


zmarouf:
Hello 👋 

i've encountered the following error while attempting a build from my M1 MBP using the Lima runner.

This happens towards the end.
```sh
ℹ️  x86_64    | retrieving workspace from builder: 9ea5607d6df670fb2f0db1b5f161df8cfbb703b2fa3de32f83458784062f2f2e
ℹ️  x86_64    | limactl [shell --workdir /tmp/melange melange-builder nerdctl exec -i tar cf - -C /home/build melange-out]
time="2023-07-10T15:20:45Z" level=fatal msg="no such container tar"
2023/07/10 17:20:45 ERROR: failed to build package. the build environment has been preserved:
ℹ️  x86_64    |   workspace dir: /tmp/melange/melange-workspace-2138480114
ℹ️  x86_64    |   guest dir: /tmp/melange/melange-guest-711452594
2023/07/10 17:20:45 error during command execution: failed to build package: retrieving workspace: exit status 1
```
Using the latest melange image which is `sha256:a2a9c76c6d533dac0960eb2cc83366f1f67d087a3eaebb140e68da4a8a768e80` 

Looks like the `nerdctl` command is missing the podID when attempting to create the archive.
It seems to be coming from here
 https://github.com/chainguard-dev/melange/blob/2bf0a3f93de92373a830a84fb218da0b07dc2fed/pkg/container/lima_runner.go#L259

Other functions use a different flow
```
baseargs = append(baseargs, cfg.PodID)
err := l.nerdctl(ctx, melangeVMName, nil, nil, nil, baseargs...)
```

I don't write Go but it seems that adding`cfg.PodID` to the function call might do the trick?

Happy to test!

deitch:
Finally got it replicated. Had to go down some interesting rabbit holes relating to lima strictness on yaml testing, but I can replicate this.

deitch:
Yeah, you are right. It is missing the pod name.

deitch:
Fixed by #543

PR #543, opened 2023-07-11
lima startup issues fixed

deitch:
1. `limactl` now accepts a more normal `-` for stdin
2. There was a misalignment in our template `lima.yaml`. Older limactl did not care, but newer is stricter (as it should be)


deitch:
Also fixes #541 by adding the containerID, and running in a goroutine so it does not block.

kaniini (approved):

PR #532, opened 2023-07-04
Fix/python version issue

mesaglio:
This pr solve #531 issue.

I change the `SetupContext` function to return an array of `*PythonContext` to force a test with `!` in version.

PR #544, opened 2023-07-11
add dir option to ruby pipelines as not all gemspecs live in the root…

rawlingsj:
… folder


deitch (approved):

deitch:
Overriding and merge as these are known false errors that are being fixed in https://github.com/chainguard-dev/melange/pull/532

rawlingsj:
thank you

PR #545, opened 2023-07-12
K8s runner various fixes

joshrwolf:
adds some minimal effort bug fixes / qol improvements:

bug fixes:

- fixes the cases where the build workspace contained symlinks/targets that weren't previous unpacked, causing them to be missed the `pkg.Emit()`

qol:

- add a retryable workspace tar fetcher (inspired by [`kubectl cp`](https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/cp/cp.go#L236))
- logs logs logs! most of the errors pop up for k8s specific reasons, so surface these better

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -34,6 +34,7 @@ import (
 	apko_types "chainguard.dev/apko/pkg/build/types"
 	apko_iocomb "chainguard.dev/apko/pkg/iocomb"
 	apko_log "chainguard.dev/apko/pkg/log"
+	apkofs "github.com/chainguard-dev/go-apk/pkg/fs"
 	"go.opentelemetry.io/otel"
 	"k8s.io/kube-openapi/pkg/util/sets"
 
@@ -1924,6 +1925,7 @@ func (b *Build) RetrieveWorkspace(ctx context.Context, cfg *container.Config) er
 	defer span.End()
 
 	b.Logger.Infof("retrieving workspace from builder: %s", cfg.PodID)
+
 	r, err := b.Runner.WorkspaceTar(ctx, b.containerConfig)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


pkg/container/kubernetes_runner.go
```diff
@@ -569,3 +568,91 @@ func (k *k8sLoader) LoadImage(ctx context.Context, layer ggcrv1.Layer, arch apko
...
+				Stdout:    true,
+				Stderr:    true,
+			}, scheme.ParameterCodec)
+
+		exec, err := remotecommand.NewSPDYExecutor(t.restConfig, "POST", req.URL())
+		if err != nil {
+			return fmt.Errorf("failed to create remote command executor: %v", err)
+		}
+
+		return exec.StreamWithContext(context.Background(), remotecommand.StreamOptions{
```
jonjohnsonjr, line 625: Please propagate context 🙏

joshrwolf: I missed this thanks! this also made me realize I missed passing the actual retry 🤯 

will refactor things slightly to make sure context is passed while still implementing `io.Reader`

pkg/build/build.go
```diff
@@ -1924,41 +1925,96 @@ func (b *Build) RetrieveWorkspace(ctx context.Context, cfg *container.Config) er
...
+			if err := dfs.Link(hdr.Linkname, hdr.Name); err != nil {
 				return err
 			}
-			f.Close()
+		default:
+			return fmt.Errorf("unsupported tar type %d for %s", hdr.Typeflag, hdr.Name)
+		}
+
+		// TODO: The primary caller of this function is the KubernetesRunner, which
+		// uses the busybox variant of `tar` that doesn't support xattrs. So this
```
kaniini, line 2007: Could we not make it use libarchive-tools instead?

PR #546, opened 2023-07-13
Stop breaking github action.

mesaglio:
To stop other pr gh-action brakes, we can comment the test while I work on the fix in this days.


mesaglio:
@rawlingsj @deitch 
I am trying to change that in the tests, make httptestserver request.

mesaglio:
This fix the issue in python convert test.

This problem was because the last version generated by the pypi `jsonschema` package has a library that when created the pr did not have and it failed to validate the number of packages it should generate. What I did was that in the tests I consumed the httptest instead of pypi. This could have happened with the `botocore` library too.

kaniini (approved):

PR #551, opened 2023-07-17
K8s runner fetch workspace tgz

joshrwolf:
- changes the workspace fetcher to use a tgz stream instead
- adds a retryable workspace fetcher for the k8s runner. this was a significant source of flakes that is mostly eliminated now
- updates the other runners that use workspace fetcher to use tgz (`lima`)

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -16,6 +16,7 @@ package build
 
 import (
 	"archive/tar"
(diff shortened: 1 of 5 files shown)


joshrwolf:
keeping this in draft until #548 lands

mattmoor (approved):

PR #552, opened 2023-07-17
Default remove builder

joshrwolf:
Change the default for runners that create a persisting build environment (everything except `bwrap`) to always destroy the build environment when the build succeeds/fails.

This means runners like `docker`/`kubernetes` don't leave dangling resources around on failures unless the user opts in (`--debug-runner`).

I originally coupled this to the existing `--debug` flag, but then changed my mind and used `--debug-runner`. I'm not married to the semantics so please suggest otherwise!

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -403,6 +403,7 @@ type Build struct {
 	imgRef             string
 	containerConfig    *container.Config
 	Debug              bool
+	DebugRunner        bool
 	LogPolicy          []string
 
 	EnabledBuildOptions []string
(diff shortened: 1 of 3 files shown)


imjasonh (approved):

pkg/build/build.go
```diff
@@ -1645,6 +1654,13 @@ func (b *Build) BuildPackage(ctx context.Context) error {
 		if err := b.Runner.StartPod(ctx, cfg); err != nil {
 			return fmt.Errorf("unable to start pod: %w", err)
 		}
+		if !b.DebugRunner {
```
kaniini, line 1657:
```suggestion
		if !b.DebugRunner && !b.IsBuildLess() {
```

joshrwolf: this conditional is already part of the outer:

```go
if !b.IsBuildLess() {
```

kaniini (approved):

PR #566, opened 2023-07-22
skip the cache mount for kubernetes runner builds

joshrwolf:
the melange cache mounts are primarily designed for bind-mounted runners, which k8s is not. in remote execution environments, it's usually just faster to redownload things than to push/pull them in.

using the cache mount for k8s runner ends up just creating more things that need to be transferred over the network via kontext -> OCI registry, and can lead to pretty silly processing if you're not careful.

this also prevents @mattmoor from destroying his poor computer/network


pkg/container/kubernetes_runner.go
```diff
@@ -396,12 +396,16 @@ func (k *k8s) filterMounts(mount BindMount) bool {
 		return true
 	}
 
+	if mount.Destination == "/var/cache/melange" {
```
mattmoor, line 399: Is there a constant that we can reference here?

joshrwolf: there weren't, but I added some!

PR #567, opened 2023-07-24
Add support s3 as cache

gtn3010:


kaniini:
Needs `go mod tidy`, but it would be nice to abstract this a little more than just an `if`/`else` block.

gtn3010:
Hi, Could you give me some hints? Like should I create a new package like remote cache with methods fetch bucket/upload bucket for each provider?

PR #569, opened 2023-07-24
Fix subpackage SBOM generation

jonjohnsonjr:
We need to generate SBOMs after we fetch the workspace, since the subpackage filesystem may not be available locally.


pkg/build/build.go
```diff
@@ -1688,8 +1688,6 @@ func (b *Build) BuildPackage(ctx context.Context) error {
 
 	// run any pipelines for subpackages
 	for _, sp := range b.Configuration.Subpackages {
-		langs := []string{}
```
joshrwolf, line 1691: this is a fun little 🐛

joshrwolf (approved):

kaniini (approved):

PR #571, opened 2023-07-25
add k8s runner config loading from envvars

joshrwolf:
adds a 3rd config loading strategy to put us at:

1. env (for certain fields)
2. `.melange.k8s.yaml`
3. defaults

adds tests to cover

### DIFF
--- a/pkg/container/kubernetes_runner.go
+++ b/pkg/container/kubernetes_runner.go
@@ -33,6 +33,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 	"github.com/imdario/mergo"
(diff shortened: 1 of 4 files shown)


pkg/container/kubernetes_runner.go
```diff
@@ -465,7 +467,7 @@ func NewKubernetesConfig(opt ...KubernetesRunnerConfigOptions) *KubernetesRunner
 		Provider:     "generic",
 		Namespace:    "default",
 		Repo:         "ttl.sh/melange",
-		StartTimeout: 5 * time.Minute,
+		StartTimeout: 10 * time.Minute,
```
joshrwolf, line 470: sneaking this in as well since ~5m seems to border too closely to gke autopilot provisioning times

mattmoor (dismissed):

mattmoor (approved):

PR #586, opened 2023-07-28
Make var transforms work in bump

Elizafox:
My previous work has now come to a head and now var transforms work in bump with this patch.

Not all variables are available for substitution at the moment, as that would complicate the logic considerably. Variables that are presently supported for substitution:

- ${{package.name}}
- ${{package.version}}
- ${{package.epoch}}
- ${{build.arch}}

This is sufficient for now. All packages that use var transforms only transform the version at present.

### DIFF
--- a/pkg/config/vars.go
+++ b/pkg/config/vars.go
@@ -36,6 +36,7 @@ const (
 	SubstitutionBuildArch            = "${{build.arch}}"
 )
 
+// Get get variables from configuration and return them in a map
 func (cfg Configuration) GetVarsFromConfig() (map[string]string, error) {
 	nw := map[string]string{}
 
@@ -53,6 +54,7 @@ func (cfg Configuration) GetVarsFromConfig() (map[string]string, error) {
 	return nw, nil
 }
 
+// Perform variable substitutions from the config on a given map
 func (cfg Configuration) PerformVarSubstitutions(nw map[string]string) error {
 	for _, v := range cfg.VarTransforms {
 		nk := fmt.Sprintf("${{vars.%s}}", v.To)
--- a/pkg/config/vars.go
+++ b/pkg/config/vars.go
@@ -36,7 +36,7 @@ const (
 	SubstitutionBuildArch            = "${{build.arch}}"
 )
 
-// Get get variables from configuration and return them in a map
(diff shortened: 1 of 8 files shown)


pkg/config/vars.go
```diff
@@ -0,0 +1,74 @@
...
+	SubstitutionTargetsDestdir       = "${{targets.destdir}}"
+	SubstitutionSubPkgDir            = "${{targets.subpkgdir}}"
+	SubstitutionHostTripletGnu       = "${{host.triplet.gnu}}"
+	SubstitutionHostTripletRust      = "${{host.triplet.rust}}"
+	SubstitutionCrossTripletGnuGlibc = "${{cross.triplet.gnu.glibc}}"
+	SubstitutionCrossTripletGnuMusl  = "${{cross.triplet.gnu.musl}}"
+	SubstitutionBuildArch            = "${{build.arch}}"
+)
+
+func (cfg Configuration) GetVarsFromConfig() (map[string]string, error) {
```
kaniini, line 39: Add a docstring?

pkg/config/vars.go
```diff
@@ -0,0 +1,74 @@
...
+			return nil, err
+		}
+
+		nw[nk] = nv
+	}
+
+	return nw, nil
+}
+
+func (cfg Configuration) PerformVarSubstitutions(nw map[string]string) error {
```
kaniini, line 56: Docstring?

pkg/util/map_subst.go
```diff
@@ -0,0 +1,38 @@
...
+
+package util
+
+import (
+	"fmt"
+
+	"chainguard.dev/melange/pkg/cond"
+)
+
+func MutateStringFromMap(with map[string]string, input string) (string, error) {
```
kaniini, line 23: Docstring?

Elizafox: Original didn't have one. I fixed it.

pkg/config/vars.go
Elizafox: Added.

pkg/config/vars.go
Elizafox: Added.

kaniini (approved):

PR #588, opened 2023-08-02
Version transform block in melange

Elizafox:
This is preliminary work to add reverse version mapping to map package versions to valid APK versions.

This only teaches Melange about the proposed new block, there are no functional changes beyond this as of yet.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -314,6 +314,8 @@ type Update struct {
 	ReleaseMonitor *ReleaseMonitor `yaml:"release-monitor,omitempty"`
 	// The configuration block for updates tracked via the Github API
 	GitHubMonitor *GitHubMonitor `yaml:"github,omitempty"`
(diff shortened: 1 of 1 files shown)


pkg/config/config.go
```diff
@@ -314,6 +314,8 @@ type Update struct {
 	ReleaseMonitor *ReleaseMonitor `yaml:"release-monitor,omitempty"`
 	// The configuration block for updates tracked via the Github API
 	GitHubMonitor *GitHubMonitor `yaml:"github,omitempty"`
+	// The configuration block for transforming the `package.version` into an APK version
+	VersionTransform *VersionTransform `yaml:"version-transform,omitempty"`
```
kaniini, line 318: I would make it an array, that way multiple version transforms can stack.

```suggestion
	VersionTransform []VersionTransform `yaml:"version-transform,omitempty"`
```

kaniini (dismissed):

kaniini (approved):

PR #590, opened 2023-08-04
build: package: add pkgconf-based SCA to catalog SDKs which use it

kaniini:
This creates providers such as `pc:glib-2.0=2.76.4`.

It can optionally create *dependencies* too, which result in the relevant `-dev` packages relating to pkg-config dependencies being pulled in as build-time dependencies.  This feature is disabled until we have rebuilt enough of Wolfi to make it tenable.

### DIFF
--- a/pkg/build/package.go
+++ b/pkg/build/package.go
@@ -28,10 +28,12 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 	"text/template"
 
+	"github.com/chainguard-dev/go-pkgconfig"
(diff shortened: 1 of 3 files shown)


pkg/build/package.go
```diff
@@ -28,10 +28,12 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 	"text/template"
 
+	"github.com/chainguard-dev/go-pkgconfig"
```
luhring, line 36: Nice, I hadn't seen this

luhring (approved):

PR #595, opened 2023-08-09
Readme revs and fixes for issues raised by `golint`

jessp01:
- Some typo corrections
- Doc line above exported consts and functions
- Implemented some golint suggestions

I also added `.pre-commit-config.yaml` to the repo. To use it:

- `pip install pre-commit`
- Install the following Go packages:
  * go install github.com/go-critic/go-critic/cmd/gocritic@latest
  * go install https://pkg.go.dev/golang.org/x/tools/cmd/goimports@latest
  * go install golang.org/x/lint/golint@latest
- In your clone, run: `pre-commit install`

This will create .git/hooks/pre-commit and following that, these tests will run every time you invoke `git commit`:
```sh
go fmt...................................................................Passed
go imports...............................................................Passed
go vet...................................................................Passed
go lint..................................................................Passed
go-critic................................................................Passed
```

For now, I commented the `go-imports`, `go-vet`, `go-lint` and `go-critic` as they still have findings that should be looked it but I recommend enabling these once that's done.

To run all checks, uncomment them in `.pre-commit-config.yaml`, invoke `pre-commit install` and run `pre-commit run --all-files`
[melange-lint.log](https://github.com/chainguard-dev/melange/files/12305591/melange-lint.log)

Attached is the full output of such a run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ To use the examples, you'll generally want to mount your current directory into 
 docker run --privileged -v "$PWD":/work cgr.dev/chainguard/melange build examples/gnu-hello.yaml
 ```
 
-Running outside of a container requires [Docker](https://docs.docker.com/get-docker/), but should also work with other runtimes such as [podman](https://podman.io/getting-started/installation).
+Running outside of a container requires [Docker](https://docs.docker.com/get-docker/), but should also work with other run-times such as [podman](https://podman.io/getting-started/installation).
 
 Alternatively, if you're on a Mac, you can use the [melange instructions for Lima](https://github.com/chainguard-dev/melange/blob/main/mac/README.md#Lima) to run an Alpine Linux VM.
 
@@ -146,15 +146,15 @@ pipeline:
 
 Melange provides the following default substitutions which can be referenced in the build file pipeline:
 
-| **Substitution**         | **Description**                                   |
-|--------------------------|---------------------------------------------------|
-| `${{package.name}}`      | Package name                                      |
-| `${{package.version}}`   | Package version                                   |
-| `${{package.epoch}}`     | Package epoch                                     |
-| `${{targets.destdir}}`   | Directory where targets will be stored            |
-| `${{targets.subpkgdir}}` | Directory where subpackage targets will be stored |
+| **Substitution**         | **Description**                                    |
+|--------------------------|----------------------------------------------------|
+| `${{package.name}}`      | Package name                                       |
+| `${{package.version}}`   | Package version                                    |
+| `${{package.epoch}}`     | Package epoch                                      |
+| `${{targets.destdir}}`   | Directory where targets will be stored             |
+| `${{targets.subpkgdir}}` | Directory where sub-package targets will be stored |
 
-An example build file pipeline with subsitutuions:
+An example build file pipeline with substitutions:
 
 ```yaml
 pipeline:
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Key features:
  - **Multi-architecture by default.** QEMU is used to emulate various
    architectures, avoiding the need for cross-compilation steps.
 
-## Why
+## Motivation
 
 Secure software factories are the evolution of DevOps, allowing a
 user to prove the provenance of all artifacts incorporated
@@ -31,25 +31,25 @@ the form of OCI container images with [apko][apko].
 You can install Melange from Homebrew:
 
 ```shell
-brew install melange
+$ brew install melange
 ```
 
 You can also install Melange from source:
 
 ```shell
-go install chainguard.dev/melange@latest
+$ go install chainguard.dev/melange@latest
 ```
 
 You can also use the Melange container image:
 
 ```shell
-docker run cgr.dev/chainguard/melange version
+$ docker run cgr.dev/chainguard/melange version
 ```
 
 To use the examples, you'll generally want to mount your current directory into the container and provide elevated privileges, e.g.:
 
 ```shell
-docker run --privileged -v "$PWD":/work cgr.dev/chainguard/melange build examples/gnu-hello.yaml
+$ docker run --privileged -v "$PWD":/work cgr.dev/chainguard/melange build examples/gnu-hello.yaml
 ```
 
 Running outside of a container requires [Docker](https://docs.docker.com/get-docker/), but should also work with other run-times such as [podman](https://podman.io/getting-started/installation).
@@ -103,13 +103,13 @@ pipeline:
 We can build this with:
 
 ```shell
-melange build examples/gnu-hello.yaml
+$ melange build examples/gnu-hello.yaml
 ```
 
 or, with Docker:
 
 ```shell
-docker run --privileged --rm -v "${PWD}":/work \
+$ docker run --privileged --rm -v "${PWD}":/work \
   cgr.dev/chainguard/melange build examples/gnu-hello.yaml
 ```
 
@@ -118,8 +118,9 @@ This will create a `packages` folder, with an entry for each architecture suppor
 If you want to sign your apks, create a signing key with the `melange keygen` command:
 
 ```shell
-melange keygen
+$ melange keygen
 ```
+
 ```
  generating keypair with a 4096 bit prime, please wait...
  wrote private key to melange.rsa
--- a/docs/BUILD-PROCESS.md
+++ b/docs/BUILD-PROCESS.md
@@ -111,4 +111,4 @@ etc. - all `runs` commands inside pipelines run natively on the processor.
 On the other hand, when building for alternative architecture, e.g. for arm64 while on amd64, the commands are run
 using [binfmt_misc](https://en.wikipedia.org/wiki/Binfmt_misc) user-mode emulation.
 
-melange does not need to do anything to make this work, provided `binfmt_misc` is installed on the host system.
+`melange` does not need to do anything to make this work, provided `binfmt_misc` is installed on the host system.
--- a/docs/md/melange_completion.md
+++ b/docs/md/melange_completion.md
@@ -26,25 +26,30 @@ MacOS:
 
 Zsh:
 
+
+```sh
 # If shell completion is not already enabled in your environment you will need
 # to enable it.  You can execute the following once:
 
 $ echo "autoload -U compinit; compinit" >> ~/.zshrc
+```
 
+```sh
 # To load completions for each session, execute once:
 $ melange completion zsh > "${fpath[1]}/_melange"
 
 # You will need to start a new shell for this setup to take effect.
+```
 
 Fish:
-
+```sh
 $ melange completion fish | source
 
 # To load completions for each session, execute once:
 $ melange completion fish > ~/.config/fish/completions/melange.fish
-
-
 ```
+
+```sh
 melange completion [bash|zsh|fish|powershell]
 ```
 
--- a/docs/md/melange_convert_apkbuild.md
+++ b/docs/md/melange_convert_apkbuild.md
@@ -15,14 +15,14 @@ Converts an APKBUILD package into a melange.yaml
 
 Converts an APKBUILD package into a melange.yaml.
 
-```
-melange convert apkbuild [flags]
+```sh
+$ melange convert apkbuild [flags]
 ```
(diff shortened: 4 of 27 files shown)


kaniini:
A lot of these comments do not actually provide any useful documentation.  For example, you add documentation comments like:

```patch
+ // BubblewrapName = "bubblewrap"
 const BubblewrapName = "bubblewrap"
```

Other comments added are just rewording of the type names, for example:

```patch
+ // KubernetesRunnerConfigPodTemplate pod configuration template
 type KubernetesRunnerConfigPodTemplate struct {
```

I don't see any reason why these changes should be merged.  Please split this PR into smaller ones addressing specific concerns, and leave out the docstrings added just to have a docstring.

jessp01:
Hi @kaniini,

I agree with you that some of these doc strings are of no real value. Having said that, `golint` does provide some useful insights (some of these are reflected in the commits I made in this pull and others can be seen in the full log I attached). My goal in "fixing" the unimportant ones is to get to a point where there are no warnings so that it will be easy to pick up on the important findings. If you prefer, we can make `golint` ignore non-significant findings instead but that would involve the same amount of work, really. What do you think?

kaniini:
We already do linting on Melange using `golangci-lint`, which is already configured to focus on the important findings.

kaniini:
As an aside, fixing code simply to appease a linter, has in the [past resulted in the introduction of new CVEs](https://research.swtch.com/openssl).  Code changes must be properly justified, e.g. they should explain why the linter violation is noteworthy, and how the change remediates the linter violation while retaining correct functionality.

jessp01:
Hi @kaniini ,

> As an aside, fixing code simply to appease a linter, has in the [past resulted in the introduction of new CVEs](https://research.swtch.com/openssl).

While I agree with you in principle (impossible not to, really), I don't think one could argue that adding such comments can cause any harm. At any rate, as we both agree those particular comments do not add value, I removed them. Please review.

README.md
```diff
@@ -13,7 +13,7 @@ Key features:
  - **Multi-architecture by default.** QEMU is used to emulate various
    architectures, avoiding the need for cross-compilation steps.
 
-## Why
+## Motivation
```
kaniini, line 16: Why?

README.md
```diff
@@ -31,28 +31,28 @@ the form of OCI container images with [apko][apko].
 You can install Melange from Homebrew:
 
 ```shell
-brew install melange
+$ brew install melange
```
kaniini, line 34: No, we intentionally do not include the `$` prefix, because we want it to be easy for users to copy and paste commands out of the documentation.

README.md
```diff
@@ -31,28 +31,28 @@ the form of OCI container images with [apko][apko].
...
 
 To use the examples, you'll generally want to mount your current directory into the container and provide elevated privileges, e.g.:
 
 ```shell
-docker run --privileged -v "$PWD":/work cgr.dev/chainguard/melange build examples/gnu-hello.yaml
+$ docker run --privileged -v "$PWD":/work cgr.dev/chainguard/melange build examples/gnu-hello.yaml
 ```
 
-Running outside of a container requires [Docker](https://docs.docker.com/get-docker/), but should also work with other runtimes such as [podman](https://podman.io/getting-started/installation).
+Running outside of a container requires [Docker](https://docs.docker.com/get-docker/), but should also work with other run-times such as [podman](https://podman.io/getting-started/installation).
```
kaniini, line 55: Runtime is correct here, Run-time is not.  https://en.wikipedia.org/wiki/Runtime_system

docs/BUILD-PROCESS.md
```diff
@@ -111,4 +111,4 @@ etc. - all `runs` commands inside pipelines run natively on the processor.
 On the other hand, when building for alternative architecture, e.g. for arm64 while on amd64, the commands are run
 using [binfmt_misc](https://en.wikipedia.org/wiki/Binfmt_misc) user-mode emulation.
 
-melange does not need to do anything to make this work, provided `binfmt_misc` is installed on the host system.
+`melange` does not need to do anything to make this work, provided `binfmt_misc` is installed on the host system.
```
kaniini, line 114: Why?

docs/md/melange_completion.md
```diff
@@ -26,25 +26,30 @@ MacOS:
 
 Zsh:
 
+
+```sh
```
kaniini, line 30: These docs are autogenerated, please don't modify them.

docs/md/melange_convert_apkbuild.md
```diff
@@ -15,14 +15,14 @@ Converts an APKBUILD package into a melange.yaml
 
 Converts an APKBUILD package into a melange.yaml.
 
-```
-melange convert apkbuild [flags]
+```sh
+$ melange convert apkbuild [flags]
```
kaniini, line 19: As above, we elide the `$` where we can in order to allow users to copy and paste.

hack/make-devenv.sh
```diff
@@ -19,7 +19,7 @@ run_builder() {
     if ! (docker inspect melange-inception:latest >/dev/null 2>&1 ); then
         set -e
         mkdir _output > /dev/null 2>&1 || : 
-        docker run --rm -v $(pwd):/melange -w /melange \
+        docker run --rm -v "$(pwd):/melange" -w /melange \
```
kaniini, line 22: This is superfluous, but if you're going to do it, it would be:

```suggestion
        docker run --rm -v "$(pwd)":/melange -w /melange \
```

hack/make-devenv.sh
```diff
@@ -31,7 +31,7 @@ run_builder() {
 
 run() {
     docker run --rm --privileged -w /melage -v /var/run/docker.sock:/var/run/docker.sock \
-        -v $(pwd):/melage -ti ${IMAGE_TAG}:latest hack/make-devenv.sh setup
+        -v "$(pwd):/melage" -ti ${IMAGE_TAG}:latest hack/make-devenv.sh setup
```
kaniini, line 34: As above, and also lets fix the typo since we are here:

```suggestion
        -v "$(pwd)":/melange -ti ${IMAGE_TAG}:latest hack/make-devenv.sh setup
```

pkg/cli/index.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// Index Creates a repository index from a list of package files
```
kaniini, line 24:
```suggestion
// Index is a constructor for a cobra.Command which wraps the IndexCmd function.
```

pkg/cli/index.go
```diff
@@ -45,6 +46,7 @@ func Index() *cobra.Command {
 	return cmd
 }
 
+// IndexCmd generates a new index
```
kaniini, line 49:
```suggestion
// IndexCmd is the backend implementation of the "melange index" command.
```

pkg/cli/sign.go
```diff
@@ -21,6 +21,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// SignIndex Sign an APK index
```
kaniini, line 24:
```suggestion
// SignIndex is a constructor which returns a cobra.Command which wraps the SignIndexCmd function.
```

pkg/cli/sign.go
```diff
@@ -40,6 +41,7 @@ func SignIndex() *cobra.Command {
 	return cmd
 }
 
+// SignIndexCmd called from SignIndex()
```
kaniini, line 44:
```suggestion
// SignIndexCmd is the backend implementation of the "melange sign-index" command.
```

pkg/cli/update_cache.go
```diff
@@ -15,6 +15,7 @@
 package cli
 
 import (
+	// Not sure why these are blankly imported?
```
kaniini, line 18: These imports can just be removed.

pkg/cli/update_cache.go
```diff
@@ -24,6 +25,7 @@ import (
 	"chainguard.dev/melange/pkg/renovate/cache"
 )
 
+// UpdateCache Update a source artifact cache
```
kaniini, line 28:
```suggestion
// UpdateCache is a constructor for a cobra.Command which provides the "melange update-cache" command.
```

pkg/container/kubernetes_runner.go
```diff
@@ -456,6 +456,7 @@ type KubernetesRunnerConfig struct {
 	baseConfigFile string
 }
 
+// KubernetesRunnerConfigPodTemplate pod configuration template
```
kaniini, line 459: Superfluous.  Remove.

pkg/container/kubernetes_runner.go
```diff
@@ -621,6 +620,7 @@ func (c KubernetesRunnerConfig) defaultBuilderPod(cfg *Config) *corev1.Pod {
 
 type KubernetesRunnerConfigOptions func(*KubernetesRunnerConfig)
 
+// WithKubernetesRunnerConfigBaseConfigFile sets path to config file
```
kaniini, line 623:
```suggestion
// WithKubernetesRunnerConfigBaseConfigFile sets the path to a config file which contains KubernetesRunner-specific options
```

pkg/http/http.go
```diff
@@ -45,7 +45,7 @@ func NewClient(rl *rate.Limiter) *RLHTTPClient {
 // sha256 hash of it.
 //
 // On success, it will return the sha256 hash as a string.
-func (r *RLHTTPClient) GetArtifactSHA256(ctx context.Context, artifactURI string) (string, error) {
+func (c *RLHTTPClient) GetArtifactSHA256(ctx context.Context, artifactURI string) (string, error) {
```
kaniini, line 48: Why?

jessp01: Because in line 22 you have:
```go
func (c *RLHTTPClient) Do(req *http.Request) (*http.Response, error) {
```

`golint` flagged it with: `receiver name r should be consistent with previous receiver name c for RLHTTPClient`, in this case, I don't think it's completely irrelevant. IMHO, consistency can only help. 
While we're looking at this one though, not sure why the receiver is needed here but I could be missing something....

hack/make-devenv.sh
jessp01: I would never use spaces in paths but I've known people (especially on Mac) to do so.. in terms of where the quotes should go, both will result in the same expansion but to me, my form is a bit more readable but it's not something I can claim to be important so I've no objection to changing it if you prefer.

kaniini:
> Please review.

I have reviewed, let me know when you want additional review.

jessp01:
I added two comments following your review. Let me know what you think. I'll then create a new pull with the changes we both agree on to make things more orderly.

docs/BUILD-PROCESS.md
jessp01: To me, using backticks for command/package names is good practice and I've seen you do the same in your documents as well. But, if you don't like the backticks or feel that in this case it should be the project name, rather than the executable, then perhaps "Melange" is more correct here?

hack/make-devenv.sh
kaniini: No.

`"$(pwd):/melange"` and `"$(pwd)":/melange` are two very different constructs from shell grammar perspective.  Your version may be equivalently safe against spaces in practice, but is represented as a single token in the AST, rather than two concatenated tokens.

By isolating the variable expansion into a subtoken, we have more predictable behavior when evaluating the script across various shells.  See also [POSIX XCU section 2.3, item 4](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_02_03), which specifically identifies this quoting behavior.

pkg/http/http.go
kaniini: Ok.

kaniini:
I have followed up on your comments.  I do not think we need a separate PR.

jessp01:
Closing in favour of https://github.com/chainguard-dev/melange/pull/598

docs/md/melange_completion.md
jessp01: @kaniini ,  since these are auto generated, would it not be better to add them to `.gitignore`?

PR #622, opened 2023-08-21
Add ${{targets.contextdir}}

kaniini:
Adds `${{targets.contextdir}}` variable expansion which maps to the current package output context directory.  This is intended to be used as a destination target instead of `${{targets.destdir}}` and `${{targets.subpkgdir}}`, allowing pipelines to be more reusable in subpackages, without hacks like `${{inputs.subpackage}}`.


kaniini:
Looks like guac will have to be fixed up once this lands.

PR #624, opened 2023-08-21
Fix the links to commands, fix the URLs generated.

### LINKED ISSUE #623: Links to melange commands are broken.
All the links here:
https://github.com/chainguard-dev/melange/blob/main/docs/md/melange.md

Are broken, for example:
https://github.com/chainguard-dev/melange/blob/main/open-source/melange/reference/melange_sign

Screenshot 2023-08-21 at 15 11 20

vaikas:
Fixes #623 

Since it would seem that the existing behavior creates something elsewhere, added a new Makefile target that produces correct links within the repo, and to control that needed to add a --suffix flag (previously, it hardcoded '/').

Also add two previously missed files:
`query`, and `package-version`.

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -25,7 +25,7 @@ jobs:
           version: v1.53
           args: --timeout=5m
 
-      - run: make docs
+      - run: make docs-repo
 
       - run: |
           go mod tidy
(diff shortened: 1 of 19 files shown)


vaikas:
if I run make docs from the head it also reports diffs, so I'm a little confused what's going on here.

luhring (approved):

PR #637, opened 2023-08-29
Normalize python version so --python-version 3 works correctly.

vaikas:
If you use --python-version 3 it produces python-3 when it should produce python3 for runtime / package dependencies, so add a function for normalizing things.

### DIFF
--- a/pkg/convert/python/python.go
+++ b/pkg/convert/python/python.go
@@ -202,6 +202,15 @@ func (c *PythonContext) findDep(ctx context.Context) error {
 	return c.findDep(ctx)
 }
 
+func normalizePythonVersion(version string) string {
+	// If the version is "3" then it's "python3" instead of "python-3"
+	// TODO(vaikas): Are there other special cases like this?
+	if version == "3" {
+		return "python3"
+	}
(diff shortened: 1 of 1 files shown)


pkg/convert/python/python.go
```diff
@@ -259,7 +268,7 @@ func (c *PythonContext) generateEnvironment(pack Package) apkotypes.ImageConfigu
 		"wolfi-base",
 		"busybox",
 		"build-base",
-		"python-" + c.PythonVersion,            //Set the python version requested
+		normalizePythonVersion(c.PythonVersion),
```
kaniini, line 271:
```suggestion
		"python" + c.PythonVersion,
```

vaikas: Don't we want python3 or python-3.11 here since the normalize bakes in the python? Or maybe I'm missing something here?

vaikas:
Talked with @kaniini last night, and while this works, the better long term solution is to create a python-3 package that has the version stream, so that we can pick up newer ones without explicitly having to do this. I'll work on that package and hopefully we won't even need this anymore.

Issue #638, opened 2023-08-29
Question about the baked in repos.


vaikas:
Would it be bad...

Currently when you run melange convert python, it adds the wolfi repos by default, here:
https://github.com/chainguard-dev/melange/blob/main/pkg/convert/python/python.go#L269

I think it would be nice to be able to pass a flag that those do not get added. Was thinking of just adding a flag --no-wolfi-repo (or something like that) and then a conditional right around the line above.

Context is to try to reduce manual mucking about after a file gets converted, and removing these for wolfi-dev/os repos is usually something that one always does manually. But maybe this assumption is wrong?

kaniini:
I think any wolfi-specific behavior should be encoded in wolfictl, not melange.

vaikas:
ok, but as a babystep:
 * Remove the hardcodes above (this also applies to things like gem, maybe others)
 * Add them as 'additional-repos' and 'additional-keyrings' in the "calling functions"
 * Add flag to not do ☝️ to keep the existing behaviour as is
 
 Then fix the other users to pass them in explicitly? WDYT?

kaniini:
Yep, exactly, sounds good to me.

vaikas:
This has been done.

PR #641, opened 2023-08-30
Add --wolfi-defaults flag, clean up flag handling.

vaikas:
Prior to this change, wolfi repo / keyring was hardcoded to python/gem convert command. Add a flag --wolfi-defaults if set to false will not do that. Motivation is that then removing them is manual work, that shouldn't be necessary. Default behaviour is unchanged.
Also unified the flag handling so that the common flags are specified in one place and not replicated.
Also fixed couple of .Equal commands where the expected / got were reversed which made it wonky to read the output.


vaikas:
Fixes: #638

kaniini (approved):

PR #642, opened 2023-08-31
add builtin pipelines for python

imjasonh:
Also use them in `melange convert python`, resolving a TODO.

### DIFF
--- a/pkg/build/pipelines/python/build-wheel.yaml
+++ b/pkg/build/pipelines/python/build-wheel.yaml
@@ -1,0 +1,17 @@
+name: Build a Python wheel
+
+needs:
+  packages:
+    - busybox
+
+pipeline:
+  - runs: |
+      if ! [ -x "$(command -v python)" ]; then
+        echo 'Error: Python is not installed.'
+        exit 1
+      fi
+
+  - runs: |
+      python -m build
+      python3 -m installer -d "${{targets.contextdir}}" dist/*.whl
+      find ${{targets.contextdir}} -name "*.pyc" -exec rm -rf '{}' +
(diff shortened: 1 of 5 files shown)


pkg/build/pipelines/python/build-wheel.yaml
```diff
@@ -0,0 +1,17 @@
+name: Build a Python wheel
+
+needs:
+  packages:
+    - busybox
+
+pipeline:
+  - runs: |
+      if ! [ -x "$(command -v python)" ]; then
```
kaniini, line 9:
```suggestion
      if ! [ -x "$(command -v python3)" ]; then
```

pkg/build/pipelines/python/build.yaml
```diff
@@ -0,0 +1,15 @@
+name: Build a Python wheel
```
vaikas, line 1: What's the difference between this and build-wheel above?

pkg/build/pipelines/python/build-wheel.yaml
```diff
@@ -0,0 +1,17 @@
+name: Build a Python wheel
```
vaikas, line 1: Could we add some comments here on when one would use which variant? Even if it's heuristics?

pkg/build/pipelines/python/build.yaml
imjasonh: build-wheel uses python3 specifically to build a .whl file. This `build` seems to basically just copy .py files into the right place, which is as close as Python gets to "building" things. 🤷

pkg/build/pipelines/python/build-wheel.yaml
imjasonh: I would, but I don't actually know myself. There seem to be 80 ways of packaging stuff in Python, and this is just a few of them. Each project seems to do it a slightly different and fun way.

vaikas (approved):

PR #643, opened 2023-09-01
Add flags for resolving git tags, release-monitoring

vaikas:
In particular, before this one had to do a bit of manual muckering to figure out what the github repo was, what the commit, was etc. So, have the machine do it.

Introduces `--use-github` flag which will then go to GitHub and find not only the release pointed to by pypi, which appears to be often wrong, but also the latest release, and barring that will fetch the "newest" tag by commit timestamp and then use that to construct the fetch pipeline. In some cases there can be a disagreement between what the pypi says is the version vs. what the latest version is out there (hi conda-build), so add some heuristics to figure out the source of truth from the github repo.

Introduces `--use-relmon` flag which will go check to see if a pypi package has release monitoring for this package and if so, will construct a `Update` section for it. If it can't find this, and the --use-github is above, and we find the repo then we'll use that instead.

Add ProjectURLs to pypi struct, since Homepage is not enough, there can be things like Source, etc. that we use to find where the code lives.

Add `Update` section to manifest, so we can actually use it above. Side note, I'm curious why we redefine it instead of using the one from pkg/config.

I've ran through this manually through various packages and everything seems to be looking quite nice, with various heuristics (some with releases, some without, some with multiple versions, etc.). Here are some examples:
https://gist.github.com/vaikas/75fd0d78cf6d6e3170bb71fe14c85e0e
https://gist.github.com/vaikas/504cb9d8cde0ce2f5db3965f3c69f87f
https://gist.github.com/vaikas/ee27a4b6f0baad12445d940d1693e3cd

Anyways, all these features are behind feature flags, so unless specified, nothing changes.


vaikas:
An example invocation that failed before with, say conda-build, now succeeds with using github as source of truth:
```
GITHUB_TOKEN="NOPE-USE-YOURS-PLEASE" ./melange --use-github --use-relmon --python-version 3 --wolfi-defaults=false convert python conda-build
```

ajayk (approved):

PR #650, opened 2023-09-01
Change GeneratedMelangeConfig to embed pkg/config/config instead of redefining it.

vaikas:
In order to make the tests pass, I had to change the tests (which I always dislike), but after talking it through, the Environment is required field, so make that change in tests. But I think not having to double-define things is a good thing.


vaikas:
Talked to @kaniini and the right thing is to make the Environment required, and adjust the tests.

kaniini (approved):

PR #651, opened 2023-09-02
add kms signer

### LINKED ISSUE #441: KMS for signing packages
Similar to cosign project, does it make sense melange can sign with KMS too? I understand that the end game is keyless signing but in the midtime could it be a good option to avoid creating key and manage it.

### LINKED ISSUE #443: KMS for signing packages
Similar to cosign project, does it make sense melange can sign with KMS too? I understand that the end game is keyless signing but in the midtime could it be a good option to avoid creating keys and manage it.

lpcalisi:
- closes #443
    - create signer package to implement signers for packages & indexes.
    - move signers from build.
    - add implementation for KMS.

*NOTE: Not all KMS signers supports SHA-1*

This PR depends on https://github.com/chainguard-dev/go-apk/pull/107


lpcalisi:
@kaniini

PR #660, opened 2023-09-05
feat: add output logs for the apkbuild converter

stormqueen1990:
### Changes

Add generated file logs for the APKBUILD converter to match the behaviour of the remaining file converters. This makes the resulting folder more visible to the user.


vaikas (approved):
Thanks!!!

joshrwolf (approved):

PR #663, opened 2023-09-06
Add build-gpep517 pipeline for Python

Dentrax:
This PR adds a new Python pipeline called: `python/build-gpep517`.

In case if upstream project does not provide neither `pyproject.toml` or `setup.py`, so `python/build` and `python/build-wheel` pipelines does not work as expected since both pipelines uses `build` module and depends on aforementioned files to run correctly. Invoking the `python-build` with `-m gpep517` rather than `-m build`, resolves this issue.

### DIFF
--- a/pkg/build/pipelines/python/build-gpep517.yaml
+++ b/pkg/build/pipelines/python/build-gpep517.yaml
@@ -1,0 +1,17 @@
+name: Build a Python wheel with gpep517
+
+needs:
+  packages:
+    - busybox
+
+pipeline:
+  - runs: |
+      if ! [ -x "$(command -v python3)" ]; then
+        echo 'Error: Python is not installed.'
+        exit 1
+      fi
+
+  - runs: |
+      python3 -m gpep517 build-wheel --wheel-dir dist --output-fd 3 3>&1 >&2
+      python3 -m installer -d "${{targets.contextdir}}" dist/*.whl
+      find ${{targets.contextdir}} -name "*.pyc" -exec rm -rf '{}' +


pkg/build/pipelines/python/build-gpep517.yaml
```diff
@@ -0,0 +1,17 @@
+name: Build a Python wheel with gpep517
+
+needs:
+  packages:
+    - busybox
+
+pipeline:
+  - runs: |
+      if ! [ -x "$(command -v python3)" ]; then
```
joshrwolf, line 9: how about adding this in `needs.packages`? I think you also want `py3-gpep517` there as well?

joshrwolf (approved):

pkg/build/pipelines/python/build-gpep517.yaml
```diff
@@ -0,0 +1,14 @@
...
+    - busybox
+    - python3
+    - py3-gpep517
+    - py3-installer
+
+pipeline:
+  - runs: |
+      python3 -m gpep517 build-wheel --wheel-dir dist --output-fd 3 3>&1 >&2
+      python3 -m installer -d "${{targets.contextdir}}" dist/*.whl
+      find ${{targets.contextdir}} -name "*.pyc" -exec rm -rf '{}' +
```
kaniini, line 14: This part should be a different action, since it is unrelated to building the package.

pkg/build/pipelines/python/build-gpep517.yaml
```diff
@@ -0,0 +1,14 @@
+name: Build a Python wheel with gpep517
+
+needs:
+  packages:
+    - busybox
+    - python3
+    - py3-gpep517
```
kaniini, line 7:
```suggestion
    - py3-gpep517
    - py3-wheel
```

kaniini:
Why not just use `gpep517` directly in `python/build`?  As far as I can tell `gpep517` does handle legacy setup.py scripts?

Dentrax:
> As far as I can tell gpep517 does handle legacy setup.py scripts?

Hmmm, how can we be sure that it won't break the existing as-is flow? I was thinking to invoke `gpep517` module if there isn't any `setup.py`.

Dentrax:
Any thoughts? @kaniini

Does it make sense to check `pyproject.toml` and `setup.py` files whether exist, if so use as-is, otherwise fallback to `gpep517` way?

Dentrax:
How should we proceed here?

cc @imjasonh

PR #665, opened 2023-09-07
docs: add documentation for built-in pipelines

stormqueen1990:
Add some documentation on how to add new built-in pipelines and the process to bump the melange dependency on wolfictl.


luhring (approved):
🤩

PR #674, opened 2023-09-11
Fix some python generation issues:

dlorenc:
* python3 instead of python-3
* use build-wheel instead of build
* add py3-build and py3-installer to build-wheel

### DIFF
--- a/pkg/convert/python/python.go
+++ b/pkg/convert/python/python.go
@@ -434,18 +434,13 @@ func (c *PythonContext) generatePipeline(ctx context.Context, pack Package, vers
 
 	pythonBuild := config.Pipeline{
 		Name: "Python Build",
-		Uses: "python/build",
+		Uses: "python/build-wheel",
 	}
 
-	pythonInstall := config.Pipeline{
-		Name: "Python Install",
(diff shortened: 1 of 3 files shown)


vaikas (commented):
https://github.com/chainguard-dev/melange/pull/637

FYI, there was some discussion here wrt python3 vs. python-3, ☝️

The other changes look great!

pkg/convert/python/python.go
```diff
@@ -309,7 +309,12 @@ func (c *PythonContext) generatePackage(pack Package, version string) config.Pac
 
 	c.Logger.Printf("[%s] Run time Deps %v", pack.Info.Name, pack.Dependencies)
 
-	pack.Dependencies = append(pack.Dependencies, "python-"+c.PythonVersion)
+	// The package is called python3, but versions are provided with python-3.xx
```
kaniini, line 312: Wolfi provides both the `python3` and `python-3` virtuals now, so this shouldn't be necessary.

dlorenc: yAY!

kaniini (approved):

PR #679, opened 2023-09-13
Add pecl pipelines for phpize & install.

vaikas:
Signed-off-by: Ville Aikas <vaikas@chainguard.dev>

These are used here: https://github.com/wolfi-dev/os/pull/5508


pkg/build/pipelines/pecl/install.yaml
```diff
@@ -0,0 +1,19 @@
+name: Installs and enables a PHP PECL module.
+
+needs:
+  packages:
+    - automake
+    - busybox
+    - cmake
+    - php-8.1-dev
```
kaniini, line 8: In general I would prefer to not tie these to specific PHP versions, and instead have the user specify that a PHP version is present.  I think we can do that by depending on a `php-dev` virtual package, and requiring that the user install the php-dev package from the version stream they actually want in the build recipe.

vaikas: makes sense!

kaniini, line 8: This one still has the `php-8.1-dev`.

pkg/build/pipelines/pecl/install.yaml
```diff
@@ -0,0 +1,18 @@
+name: Installs and enables a PHP PECL module.
+
+needs:
+  packages:
+    - automake
+    - busybox
+    - cmake
```
kaniini, line 7: I'm curious about the cmake here.  Was there a PHP extension that used CMake instead of the normal PHP build system or something?

kaniini (approved):

pkg/build/pipelines/pecl/install.yaml
vaikas: Oh no earlier when I used make I think it was suggested to me that I should use cmake, but maybe it was just autoconf/make

vaikas:
FYI, just noting this discussion here in case somebody gets hit by this in the future:
https://github.com/wolfi-dev/wolfictl/issues/397

PR #681, opened 2023-09-15
Add regression tests for workdir propagation, fix long-standing bug with propagation across referenced pipelines

kaniini:
A melange pipeline can reference other pipelines with the `uses` statement, e.g.

```yaml
pipeline:
  - uses: foo
```

However, a long standing bug has existed where the pipeline settings did not necessarily propagate downstream.  This bug recently became exposed by refactoring of the pipeline code to perform downward propagation at configuration load time rather than runtime.

We also add regression tests for various forms of downward propagation in the config struct.


Elizafox:
LGTM.

Elizafox (approved):

Issue #685, opened 2023-09-15
git-checkout rejects real tag SHA of annotated tag


wlynch:
There are 2 different types of Git tags:

1. Lightweight tags, where the tag ref points to the commit SHA directly.
2. Annotated tags, where the tag ref points to a different Git object that houses the tag message, signature, etc, and points to the underlying commit SHA.

```
$ git ls-remote https://github.com/cert-manager/cert-manager | grep v1.13.0 
0475beb9395b4eb46a89f284d73209e84e84707f        refs/tags/v1.13.0  # Actual value of the annotated tag
d34bd7aa15055b428bf865851bc420412f3710e6        refs/tags/v1.13.0^{}  # Value of the unpeeled commit SHA
...
```

The `git-checkout` command seems to only respect unpeeled commit SHAs. If you try to give it the actual annotated tag SHA, it will fail:

Input:
```yaml
pipeline:
  - uses: git-checkout
    with:
      repository: https://github.com/cert-manager/cert-manager
      tag: v${{package.version}}
      expected-commit: 0475beb9395b4eb46a89f284d73209e84e84707f
```

Output:
```
ℹ️  aarch64   | Error (git-checkout): expect commit 0475beb9395b4eb46a89f284d73209e84e84707f, got d34bd7aa15055b428bf865851bc420412f3710e6
2023/09/15 20:07:48 ERROR: failed to build package. the build environment has been preserved:
ℹ️  aarch64   |   workspace dir: /tmp/melange-workspace-2791399927
ℹ️  aarch64   |   guest dir: /tmp/melange-guest-248615042
Error: failed to build package: unable to run pipeline: exit status 1
2023/09/15 20:07:48 error during command execution: failed to build package: unable to run pipeline: exit status 1
```

We actually probably want the true annotated tag value in most cases, since **this is where Git tag signatures are located**.

Annotated tag data (though not the best example since it's not signed)

```
$ git cat-file tag v1.13.0
object d34bd7aa15055b428bf865851bc420412f3710e6
type commit
tag v1.13.0
tagger Tim Ramlot <42113979+inteon@users.noreply.github.com> 1694529201 +0200

v1.13.0
```

<details><summary>Actual signed git tag from another repo (gitsign)</summary>
<p>

```
**object 8955100f1adbce701dbe4b76e7a8575126f526b7
type commit
tag v0.7.0
tagger Billy Lynch <billy@chainguard.dev> 1684775232 -0400

v0.7.0
[...]

PR #686, opened 2023-09-15
git-checkout: Allow tags to matched annotated tag SHAs, don't allow

wlynch:
fuzzy matching of refs.

Previously, we were somewhat loose with allowed revisions - the branch check for origin/foo could technically match a tag named origin/foo, and vice-versa for tag names. This changes the behavior to use rev-parse --verify to guarantee that we are looking up by the intended branch/tag name.

This also looses the allowed revisions for tags to allow the true SHA for an annotated tag. Previously we only allowed the unpeeled values and would reject the true SHA.

Fixes https://github.com/chainguard-dev/melange/issues/685

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -21,6 +21,7 @@ jobs:
     strategy:
       matrix:
         example:
+        - git-checkout.yaml
         - gnu-hello.yaml
         - mbedtls.yaml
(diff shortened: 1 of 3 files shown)


Elizafox (approved):

Elizafox:
LGTM. I'll let someone else review it before merging.

PR #700, opened 2023-09-19
Add post-file walk linting and empty package linting

Elizafox:
This checks for empty packages and also adds a facility to allow for post-walk lints that can walk the tree or do whatever.

### DIFF
--- a/pkg/build/linter.go
+++ b/pkg/build/linter.go
@@ -28,21 +28,6 @@ type LinterContext struct {
 	chk     *config.Checks
 }
 
-func (lctx LinterContext) getPackageOptions() (*config.PackageOption, error) {
-	if lctx.pkgname == lctx.cfg.Package.Name {
-		return &lctx.cfg.Package.Options, nil
-	}
-
-	for _, spkg := range lctx.cfg.Subpackages {
-		if lctx.pkgname == spkg.Name {
-			return &spkg.Options, nil
-		}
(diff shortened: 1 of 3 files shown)


Elizafox:
Hmm, some packages are failing. Should I check subpackages as well?

Elizafox:
So I've discovered in my travels that gdk-pixbuf had an empty metapackage that wasn't necessary, and exa had an empty docs package. So these weren't spurious failures at all. The linter did what it was supposed to do.

This linter is likely to show a lot of empty packages. I don't know if I should make this an optional lint for the moment, or rip the band-aid off now.

pkg/build/linter.go
```diff
@@ -141,12 +167,52 @@ func worldWriteableLinter(_ LinterContext, path string, d fs.DirEntry) error {
...
+	if foundfile {
+		return nil
+	}
+
+	options, err := lctx.getPackageOptions()
+	if err != nil {
+		return err
+	}
+
+	if options.NoProvides {
```
kaniini, line 201: I would prefer a direct signal (such as disabling the linter) rather than magical thinking here.

kaniini (dismissed):

pkg/build/linter.go
Elizafox: Done.

kaniini (approved):

PR #707, opened 2023-09-21
Linter: only check package dir, not the whole workspace

Elizafox:
We shouldn't scan the entire workspace, just the package.


Elizafox:
Superseded by #708

PR #721, opened 2023-09-27
Track vendored deps for .PKGINFO

jonjohnsonjr:
If we discover a vendored dependency (outside of libdirs), we will use that to self-satisfy any generated runtime dependencies. Notably, we will not use these vendored dependencies to satisfy any explicitly configured runtime dependencies so that we have an escape hatch for overriding the SCA.

This also adds `# vendored = ...` comments to .PKGINFO so that we can debug cases that would otherwise appear as a missing runtime dependency.

(Draft for now because I would like to do a dry run of this new logic against existing packages to see if there are any unexpected diffs.)

### DIFF
--- a/pkg/build/package.go
+++ b/pkg/build/package.go
@@ -54,6 +54,8 @@ import (
 // concurrent builds on giant machines, and uses only 1 core on tiny machines.
 var pgzipThreads = min(runtime.GOMAXPROCS(0), 8)
 
+var libDirs = []string{"lib", "usr/lib", "lib64", "usr/lib64"}
+
 func min(l, r int) int {
 	if l < r {
 		return l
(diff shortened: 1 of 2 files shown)


jonjohnsonjr:
I am pretty comfortable with this change. I ran against wolfi HEAD last night and produced https://gist.github.com/jonjohnsonjr/4cb59e0503cd16d064184d0b82c20050

There are a handful of things that seem off, filed https://github.com/wolfi-dev/os/issues/6199 for those.

Packages that will have vendored things post bump:

* `ssdeep`
* `dotnet-7`
* `dotnet-6`
* `graphene`
* `gstreamer`
* `libdrm`
* `libeconf`
* `libpciaccess`
* `libpfm`
* `libsolv`
* `libva`
* `libvdpau`
* `libvterm`
* `libxkbcommon`
* `opus`
* `unibilium`
* `utf8proc`

It seems like only the dotnet packages really make sense here and the others will need to be fixed, but the part that removes these libraries from being provided has already landed in melange. This change only removes some of these libraries from being depended on if we vendor it.

imjasonh (approved):

PR #724, opened 2023-09-28
Add stripped file linter

Elizafox:
This checks to make sure a file has no debug sections.

### DIFF
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -15,21 +15,24 @@
 package linter
 
 import (
+	"debug/elf"
 	"fmt"
 	"io/fs"
+	"path/filepath"
 	"regexp"
 
 	"chainguard.dev/melange/pkg/config"
 )
 
 type LinterContext struct {
-	pkgname string
-	cfg     *config.Configuration
-	chk     *config.Checks
+	pkgname    string
+	realfspath string
+	cfg        *config.Configuration
+	chk        *config.Checks
 }
 
-func NewLinterContext(name string, cfg *config.Configuration, chk *config.Checks) LinterContext {
-	return LinterContext{name, cfg, chk}
+func NewLinterContext(name string, realfspath string, cfg *config.Configuration, chk *config.Checks) LinterContext {
+	return LinterContext{name, realfspath, cfg, chk}
 }
 
 type linterFunc func(lctx LinterContext, path string, d fs.DirEntry) error
@@ -55,6 +58,7 @@ var linterMap = map[string]linter{
 	"usrlocal":   linter{usrLocalLinter, "This package should be a -compat package"},
 	"varempty":   linter{varEmptyLinter, "Remove any offending files in /var/empty in the pipeline"},
 	"worldwrite": linter{worldWriteableLinter, "Change the permissions of any world-writeable files in the package, disable the linter, or make this a -compat package"},
+	"strip":      linter{strippedLinter, "Properly strip all binaries in the pipeline"},
 }
 
 var postLinterMap = map[string]postLinter{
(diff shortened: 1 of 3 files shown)


pkg/linter/linter.go
```diff
@@ -156,6 +161,40 @@ func worldWriteableLinter(_ LinterContext, path string, d fs.DirEntry) error {
...
+	if err != nil {
+		// We don't particularly care if this fails, it means it's probably not an ELF file
+		fmt.Printf("WARNING: Could not open file '%s' as executable: %v\n", path, err)
+		return nil
+	}
+	defer file.Close()
+
+	// No debug sections allowed
+	if file.Section(".debug") != nil || file.Section(".zdebug") != nil {
+		return fmt.Errorf("File is not stripped")
```
imjasonh, line 192: It might be helpful for this error message to include the file path.

pkg/linter/linter.go
```diff
@@ -156,6 +161,40 @@ func worldWriteableLinter(_ LinterContext, path string, d fs.DirEntry) error {
...
+	mode := info.Mode()
+	if mode&0111 == 0 && !isObjectFileRegex.MatchString(ext) {
+		// Not an executable or library
+		return nil
+	}
+
+	file, err := elf.Open(filepath.Join(lctx.realfspath, path))
+	if err != nil {
+		// We don't particularly care if this fails, it means it's probably not an ELF file
+		fmt.Printf("WARNING: Could not open file '%s' as executable: %v\n", path, err)
```
imjasonh, line 185:
```suggestion
		fmt.Printf("WARNING: Could not open file %q as executable: %v", path, err)
```

pkg/linter/linter.go
```diff
@@ -156,6 +161,40 @@ func worldWriteableLinter(_ LinterContext, path string, d fs.DirEntry) error {
...
+	}
+
+	ext := filepath.Ext(path)
+	mode := info.Mode()
+	if mode&0111 == 0 && !isObjectFileRegex.MatchString(ext) {
+		// Not an executable or library
+		return nil
+	}
+
+	file, err := elf.Open(filepath.Join(lctx.realfspath, path))
```
imjasonh, line 182: Can we use `elf.NewFile` and extract the image from the FS, instead of having this take `realfspath`? This might make it easier to run this on arbitrary FSes, e.g., non-unpacked tars from apks streaming in over HTTP.

Elizafox: Yeah, sure. I'll modify it to take the fs path.

pkg/linter/linter.go
Elizafox: It prints it with the error, see Line 254.

pkg/linter/linter.go
Elizafox: Thanks!

pkg/linter/linter.go
Elizafox: So there's a problem with this approach:

fs.Open doesn't implement io.ReaderAt. >_<

Elizafox: Fixed it with a crappy workaround for now.

pkg/linter/linter.go
```diff
@@ -212,22 +277,21 @@ func (lctx LinterContext) LintPackageFs(fsys fs.FS, linters []string) error {
 
 			err = linter.LinterFunc(lctx, path, d)
```
imjasonh, line 278: Can LinterFunc take the fs.FS now instead of having it in the LinterContext?

(Not blocking, this lgtm)

imjasonh (approved):

pkg/linter/linter.go
Elizafox: The linter context contains more than that, and may be needed for future linters.

PR #728, opened 2023-09-29
Add section to linter docs about disabling linters

Elizafox:
This is useful for packagers to know when they should turn off a linter, and when to do so.


docs/LINTER.md
```diff
@@ -26,3 +26,25 @@ To handle these cases, we use the convention that subpackages ending in `-compat
 To encourage this convention, we have a linter that will fail if a package builds components into locations that are likely to be for upstream compatibility, where the package is not named like `-compat`.
 
 When a linter finds an issue, you should normally have the main package install into `/usr/bin`, and have a `-compat` subpackage that moves or symlinks the files into the expected location outside of the normal location.
+
+### Disabling lints
+
+When a lint fails, it is sending a clear signal that something is wrong. Therefore, ints should only be disabled after discussion with other team members ensuring it is in fact the right thing to do.
```
imjasonh, line 32:
```suggestion
When a lint fails, it is sending a clear signal that something is wrong. Therefore, lints should only be disabled after discussion with other team members ensuring it is in fact the right thing to do.
```

imjasonh (approved):

denhamparry (approved):
LGTM 👍

PR #730, opened 2023-10-01
drop sirupsen logger

imjasonh:
We moved to our own logger implementation around https://github.com/chainguard-dev/melange/pull/399 in April. This was one remaining vestige, which resulted in logs being written to stderr when the log policy disallowed it.

If this looks good we should probably remove sirupsen/logger from go-apk too, and not bother passing the `nil` logger. I checked that passing a `nil` logger to `SignIndex` doesn't result in a panic, and does log the lines.


imjasonh:
There's also https://pkg.go.dev/github.com/chainguard-dev/go-apk/pkg/logger -- maybe I don't understand how logging is supposed to work.

PR #733, opened 2023-10-02
melange bump: only reset the epoch if version changes, else increment it

rawlingsj:
This can happen when an upstream deletes a tag and re-uses it with a different commit, so we need to bump the epoch rather than reset it. Example https://github.com/wolfi-dev/os/pull/6126/commits/4a7ae4002985401afaab03b50f9cff8073462ec8

If there are no version or expected commit changes we still expect the epoch to be bumped

### DIFF
--- a/pkg/renovate/bump/bump.go
+++ b/pkg/renovate/bump/bump.go
@@ -74,7 +74,6 @@ func New(opts ...Option) renovate.Renovator {
 	return func(ctx context.Context, rc *renovate.RenovationContext) error {
 		log.Printf("attempting to bump version to %s", bcfg.TargetVersion)
 
-		// Find the package.version node first and change it.
 		packageNode, err := renovate.NodeFromMapping(rc.Configuration.Root().Content[0], "package")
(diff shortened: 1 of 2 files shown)


pkg/renovate/bump/bump.go
```diff
@@ -83,6 +84,11 @@ func New(opts ...Option) renovate.Renovator {
 		if err != nil {
 			return err
 		}
+
+		resetEpoch := true
```
jonjohnsonjr, line 88: Nit: Rather than introduce `resetEpoch`, I'd hoist the `epochNode` stuff up to here and inline the conditional as `versionNode.Value != bcfg.TargetVersion`, which would bring the `if the version is changing...` comment closer to the logic/context.

kaniini: Same, lets just host the `epochNode` up.

rawlingsj: sgtm

mattmoor (approved):

PR #738, opened 2023-10-04
don't fail on lint findings by default

imjasonh:
TODO: tests

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -91,6 +91,7 @@ type Build struct {
 	Debug              bool
 	DebugRunner        bool
 	LogPolicy          []string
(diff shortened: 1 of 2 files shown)


pkg/build/build.go
```diff
@@ -252,6 +253,14 @@ func WithConfig(configFile string) Option {
 	}
 }
 
+// WithFailOnLint sets whether or not to fail on lint errors.
+func WithFailOnLint(fail bool) Option {
```
mattmoor, line 257: I thought we had a way to make linter's warn by default.  This makes it a binary thing globally and we should be able to toggle each linter on as things are clean.

I'll take a stab at this.

PR #747, opened 2023-10-10
simplify --fail-on-lint-warning and fix error message

imjasonh:
This uses [`errors.Join`](https://pkg.go.dev/errors#Join) to collect multiple lint findings and report on them (or not) all at once, instead of failing on the first one when there may be more to find.

This also removes the `warn` callback pattern and just has `LintPackageFs` return any `Join`ed errors, or `nil` if none are found. In this change, `BuildPackage` is responsible for erroring or warning when lint errors are found, instead of burying that deep inside `LintPackageFs`.

This PR originally started as an attempt to fix a malformed error output when linting produced errors:

```
make LINT=yes package/curl-rustls
...
2023/10/10 20:40:01 error during command execution: failed to build package: package linter warning: %!w(<nil>)
```


pkg/linter/linter.go
```diff
@@ -314,33 +314,24 @@ func (lctx LinterContext) LintPackageFs(fsys fs.FS, warn func(error), linters []
...
 	// Run post-walking linters
+	var errs []error
 	for _, linterName := range postLinters {
 		linter := postLinterMap[linterName]
-		err := linter.LinterFunc(lctx, fsys)
-		if err != nil {
-			if linter.FailOnError {
-				return fmt.Errorf("Linter %s failed; suggest: %s", linterName, linter.Explain)
-			}
-			warn(err)
```
mattmoor, line 341: This eliminates the ability to have individual linters be fatal.

mattmoor: Basically: we need a way to introduce a linter as just a warning, get it clean, and then make it fatal.

If we land this then we can't make anything required without making everything required and it will become a requirement to get linters clean across Wolfi before landing any of its logic, which I don't think we want.

imjasonh: WDYT about having `--fail-on-lint-warning=empty,setuidgid,foo,bar` which selects which ones will be considered fatal?

mattmoor: That seems fine, but I still think linters should be able to have defaults.  This could be through the flag defaults, I have no preference here.

PR #748, opened 2023-10-11
validate pipeline fields together

imjasonh:
I spent an embarrassing amount of time trying to debug why this failed:

```
- runs: autoconf/make
  with:
    opts: operator
```

When I figured it out, I wanted an error to stop me from making that mistake again.

This is incomplete, there are other combinations of pipeline fields that should be an error, but this is a start at least.

### DIFF
--- a/examples/options.yaml
+++ b/examples/options.yaml
@@ -4,10 +4,7 @@ package:
   epoch: 3
   description: "URL retrieval utility and library"
   copyright:
-    - paths:
-        - "*"
-      attestation: TODO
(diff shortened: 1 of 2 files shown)


imjasonh:
It looks like tini.yaml does this in Wolfi: https://github.com/wolfi-dev/os/blob/main/tini.yaml

I'm not sure if this is a bug, and that `opts:` are just ignored in this case. If that's so, it's even clearer this should be an error so folks don't point this footgun at themselves.

PR #760, opened 2023-10-16
Add APK linting to Melange

Elizafox:
🎉 This allows for linting of APK files directly.

This uses the same default linters as melange does, so no functional changes.

Warnings and errors are reported appropriately.

I haven't constructed a bad APK yet, so that is not yet tested, but in theory it should work.


pkg/linter/defaults/defaults.go
```diff
@@ -0,0 +1,28 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package defaults
+
+var DefaultLinters = []string{
```
imjasonh, line 17: I think we can remove this package, and use the one in `linter`.

pkg/config/config.go
```diff
@@ -186,7 +187,7 @@ func (p *Package) FullCopyright() string {
 // Computes the list of package or subpackage linters, taking into account default linters.
 // This includes the default linters as well, unless disabled.
 func (chk *Checks) GetLinters() []string {
-	linters := defaultLinters
+	linters := linter_defaults.DefaultLinters
```
imjasonh, line 190:
```suggestion
	linters := linter.DefaultLinters
```

imjasonh:
> I haven't constructed a bad APK yet, so that is not yet tested, but in theory it should work.

I'm fairly sure we've constructed tons of bad APKs, we just need to find a good one from packages.wolfi.dev 😆

luhring:
> > I haven't constructed a bad APK yet, so that is not yet tested, but in theory it should work.
> 
> I'm fairly sure we've constructed _tons_ of bad APKs, we just need to find a good one from packages.wolfi.dev 😆

This one looks empty (aside from the SBOM) in case that helps! https://apk.dag.dev/https/packages.wolfi.dev/os/aarch64/kubeflow-pipelines-2.1.3-r7.apk@sha1:61a3e8a429e3fd32c49fe3ade6b757a4a03cfe2c/

pkg/linter/defaults/defaults.go
Elizafox: Done! I don't know what I was thinking there.

Elizafox:
> > > I haven't constructed a bad APK yet, so that is not yet tested, but in theory it should work.
> > 
> > 
> > I'm fairly sure we've constructed _tons_ of bad APKs, we just need to find a good one from packages.wolfi.dev 😆
> 
> This one looks empty (aside from the SBOM) in case that helps! https://apk.dag.dev/https/packages.wolfi.dev/os/aarch64/kubeflow-pipelines-2.1.3-r7.apk@sha1:61a3e8a429e3fd32c49fe3ade6b757a4a03cfe2c/

I will write a test for that using that APK then :).

imjasonh (dismissed):

imjasonh (approved):

PR #763, opened 2023-10-17
pipelines: add npm-install pipeline

julienv3:
This PR adds an `npm/install` pipeline supporting overrides.

### Functional Changes

Notes:

### DIFF
--- a/pkg/build/pipelines/npm/install.yaml
+++ b/pkg/build/pipelines/npm/install.yaml
@@ -1,0 +1,74 @@
+name: Install a portable npm package.
+
+needs:
+  packages:
+    - nodejs
+    - busybox
+
+inputs:
+  package:
+    description: |
+      The name of the package to npm install.
+    required: true
+
+  version:
+    description: |
+      The version of the package to npm install.
+    required: true
+
+  prefix:
+    description: |
+      The -prefix argument to pass to npm install; where /bin and /lib will be copied to.
+    default: "${{targets.destdir}}/usr/local/"
+
+  overrides:
+    description: |
+      Space, comma or newline-separated list of package@version to use in npm overrides, e.g. "yargs@^17.0.0 get-stdin@^9.0.0".
+
+pipeline:
+  - runs: |
+      # Portable global install of the package.
+      # Simplest way to get both bin and lib directories installed.
+      # Does not support overrides; to be fixed below if needed.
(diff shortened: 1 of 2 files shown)


pkg/build/pipelines/npm/install.yaml
```diff
@@ -0,0 +1,74 @@
...
+          # Create a package.json with the dependency installed above, at which point
+          # `npm i --install-strategy=shallow .` will output the exact same content the install above did.
+          echo "{
+            \"dependencies\": {
+              \"${{inputs.package}}\": \"${{inputs.version}}\"
+            }
+          }" > package.json
+          
+          # However, we also add `overrides`, to override specific vulnerable libraries.
+          node -e "
```
julienv3, line 48: ### Problem
There is unfortunately no programatic way to add overrides to an npm `package.json` file aside from actual file content rewrite.

### Proposal
This JS snippet is used to allow a cleaner `package@version` syntax like `npm install` does.
Node has built-in package.json loading and JSON stringifying support, a few string operations are used to support comments (since pipeline inputs have to be strings, inline comments are actually part of them) and transform the package@version syntax to actual JSON, so we can end up with an input like
```
pipeline:
  - uses: npm/install
    with:
      package: ${{package.name}}
      version: ${{package.version}}
      overrides: |
        yargs@^17.0.0       # If yargs had a CVE fixed in ^17.0.0
        get-stdin@^9.0.0    # If get-stdin had a CVE fixed in ^9.0.0
```

### Alternative
We could blindly inline ${{input.overrides}} with the echo call above, something like 
```
          echo "{
              \"dependencies\": {
                  \"${{inputs.package}}\": \"${{inputs.version}}\"
              },
              \"overrides\": {
                  ${{inputs.overrides}}
              }
          }" > package.json
```

However, that means the input has to be valid JSON, quotes and all, which leads to the not-super-user-friendly
```
pipeline:
  - uses: npm/install
    with:
      package: ${{package.name}}
      version: ${{package.version}}
      overrides: |
        \"yargs\": \"^17.0.0\",
        \"get-stdin\": \"^9.0.0\"
```
Which is stuck with JSON syntax downsides (no trailing comma allowed, no comment support to label which CVE's you're fixing with every override, escaped double quotes).

imjasonh (approved):

PR #764, opened 2023-10-17
Add multiple Python packages post-linter

Elizafox:
Take two after resolving git issues.

This implements a lint checking if a package contains multiple Python packages.

It's rather simplistic at the moment but I think this should get the job done.

### Linter

### DIFF
--- a/pkg/linter/linter_test.go
+++ b/pkg/linter/linter_test.go
@@ -39,7 +39,7 @@ func Test_emptyLinter(t *testing.T) {
 			Epoch:   0,
 			Checks: config.Checks{
 				Enabled:  []string{"empty"},
-				Disabled: []string{"dev", "opt", "setuidgid", "srv", "strip", "tempdir", "usrlocal", "varempty", "worldwrite"},
+				Disabled: []string{"dev", "opt", "pythonmultiple", "setuidgid", "srv", "strip", "tempdir", "usrlocal", "varempty", "worldwrite"},
 			},
 		},
 	}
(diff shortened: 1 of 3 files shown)


pkg/linter/linter_test.go
```diff
@@ -190,6 +190,96 @@ func Test_optLinter(t *testing.T) {
...
+	dir, err := os.MkdirTemp("", "melange.XXXXX")
+	defer os.RemoveAll(dir)
+	assert.NoError(t, err)
+
+	cfg := config.Configuration{
+		Package: config.Package{
+			Name:    "testpythonmultiplepackages",
+			Version: "4.2.0",
+			Epoch:   0,
+			Checks: config.Checks{
```
imjasonh, line 203: (not blocking) To minimize diffs in future changes, it might be helpful to have a `checkOnly("foo")` helper method in this test package, that returns a `config.Checks` with only one check enabled.

imjasonh:
The change itself looks good. Do we have data about how many existing latest packages might get tripped up by this?

pkg/linter/linter_test.go
Elizafox: yeah. I did this at some point but it's probably gonna need to be a different PR.

Elizafox:
> The change itself looks good. Do we have data about how many existing latest packages might get tripped up by this?

Doing a run on all py3 packages, will report back when it's done.

Elizafox:
Here are the failures I collected from our built APK's: https://gist.github.com/Elizafox/369640b36cb9584d217af9ee0e10e32c

Some of those are false positives, like _foo packages. I might disregard them. Others the linter should be disabled on.

This will probably remain a warning-only linter for the forseeable future.

imjasonh:
That gist is really helpful!! Thanks for adding this, let's get it into use.

imjasonh (approved):

PR #781, opened 2023-10-24
Fix ownership not being preserved

### LINKED ISSUE #501: File Ownership not preserved
When using `chown` or similar in melange, I would expect the changes to make it into the tar file, but this doesn't seem to be the case.

For example, with this melange file:

```
package:
  name: owntest
  version: 0.0.0
  epoch: 0
  dependencies:
    runtime:

environment:
  contents:
    repositories:
      - https://packages.wolfi.dev/os
    packages:
      - busybox
  accounts:
    users:
      - username: test
        uid: 1001

pipeline:
  - name: testy
  - runs: |
      touch file
      install -m744 -o test -D file ${{targets.destdir}}/etc/test/file
      chown test file
      cp file ${{targets.destdir}}/etc/
      chown test ${{targets.destdir}}/etc/file
      ls -l ${{targets.destdir}}/etc/file
      ls -l /${{targets.destdir}}/etc/test/file
      ls -l ${{targets.destdir}}/etc/test
```

I get the following results:

```
tar --same-owner -tpvf packages/aarch64/owntest-0.0.0-r0.apk
-rw-r--r--  0 root   root      209  1 Jan  1970 .PKGINFO
drwxr-xr-x  0 root   root        0  1 Jan  1970 etc
-rw-r--r--  0 root   root        0  1 Jan  1970 etc/file
drwxr-xr-x  0 root   root        0  1 Jan  1970 etc/test
-rwxr--r--  0 root   root        0  1 Jan  1970 etc/test/file
drwxr-xr-x  0 root   root        0  1 Jan  1970 var
drwxr-xr-x  0 root   root        0  1 Jan  1970 var/lib
drwxr-xr-x  0 root   root        0  1 Jan  1970 var/lib/db
drwxr-xr-x  0 root   root        0  1 Jan  1970 var/lib/db/sbom
-rw-r--r--  0 root   root     2875  1 Jan  1970 var/lib/db/sbom/owntest-0.0.0-r0.spdx.json
```

But I expected `/etc/file` and `/etc/test/file` to be owned by `test`.

The owner is `test` in the debug output from melange.

epsilon-phase:
Fix permission issues resulting from running with the build user under Docker. and make the GID/UID remapping work to correct that.

Fixes #501 

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

I have not tested this yet.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -1241,6 +1241,9 @@ func (b *Build) BuildPackage(ctx context.Context) error {
 	}
 
 	// clean build environment
+	// TODO(epsilon-phase): implement a way to clean up files that are not owned by the user
+	// that is running melange. files created inside the build not owned by the build user are
+	// not be possible to delete with this strategy.
 	if err := os.RemoveAll(b.WorkspaceDir); err != nil {
 		b.Logger.Printf("WARNING: unable to clean workspace: %s", err)
 	}
--- a/pkg/container/docker_runner.go
+++ b/pkg/container/docker_runner.go
@@ -204,10 +204,11 @@ func (dk *docker) Run(ctx context.Context, cfg *Config, args ...string) error {
 	}
 	defer cli.Close()
 
-	// TODO(kaniini): We want to use the build user here, but for now lets keep
-	// it simple.
+	// TODO(kaniini): We want to use the build user here, but for now lets keep it simple.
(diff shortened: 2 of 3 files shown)


Elizafox (approved):

Elizafox:
This looks good and passes tests. Merging :).

PR #788, opened 2023-10-25
SCA: add python dependency generator

kaniini:
For packages containing Python modules, automatically add `python3~$VERSION` as a runtime dependency if a `python` dependency is not already specified.  This helps us to provide more specific version constraints for our Python module packages in Wolfi, which in turn makes Python upgrades less disruptive.

### SCA Changes

### DIFF
--- a/pkg/build/package.go
+++ b/pkg/build/package.go
@@ -653,6 +653,66 @@ func generatePkgConfigDeps(pc *PackageBuild, generated *config.Dependencies) err
 	return nil
 }
 
+// generatePythonDeps generates a python3~$VERSION dependency for packages which ship
+// Python modules.
+func generatePythonDeps(pc *PackageBuild, generated *config.Dependencies) error {
+	var pythonModuleVer string
+	pc.Logger.Printf("scanning for python modules...")
+
+	fsys := readlinkFS(pc.WorkspaceSubdir())
+	if err := fs.WalkDir(fsys, ".", func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		// Python modules are installed in paths such as /usr/lib/pythonX.Y/site-packages/...,
+		// so if we find a directory named site-packages, and its parent is a pythonX.Y directory,
+		// then we have a Python module directory.
+		basename := filepath.Base(path)
+		if basename != "site-packages" {
+			return nil
+		}
+
+		parent := filepath.Dir(path)
+		basename = filepath.Base(parent)
+		if !strings.HasPrefix(basename, "python") {
+			return nil
+		}
+
+		// This probably shouldn't ever happen, but lets check to make sure.
+		if !d.IsDir() {
+			return nil
+		}
+
+		// This takes the X.Y part of the pythonX.Y directory name as the version to pin against.
+		// If the X.Y part is not present, then pythonModuleVer will remain an empty string and
+		// no dependency will be generated.
+		pythonModuleVer = basename[6:]
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	// Nothing to do...
+	if pythonModuleVer == "" {
+		return nil
+	}
+
+	// Do not add a Python dependency if one already exists.
+	for _, dep := range pc.Dependencies.Runtime {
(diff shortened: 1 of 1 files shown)


pkg/build/package.go
```diff
@@ -653,6 +653,54 @@ func generatePkgConfigDeps(pc *PackageBuild, generated *config.Dependencies) err
...
+		if !strings.HasPrefix(basename, "python") {
+			return nil
+		}
+
+		// This probably shouldn't ever happen, but lets check to make sure.
+		if !d.IsDir() {
+			return nil
+		}
+
+		pythonModuleVer = basename[6:]
```
imjasonh, line 678: What will this do if there happens to be a directory in the package named `python` for whatever reason? Would it create a runtime dep on `python~`? Is that allowed?

I think it would be great to have some test cases to act as runnable documentation about what this check expects to find and produce.

kaniini: It would set `pythonModuleVer` to `""`, which is the default anyway, thus no dependency would be added.

pkg/build/package.go
```diff
@@ -653,6 +653,66 @@ func generatePkgConfigDeps(pc *PackageBuild, generated *config.Dependencies) err
...
+	pc.Logger.Printf("scanning for python modules...")
+
+	fsys := readlinkFS(pc.WorkspaceSubdir())
+	if err := fs.WalkDir(fsys, ".", func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		// Python modules are installed in paths such as /usr/lib/pythonX.Y/site-packages/...,
+		// so if we find a directory named site-packages, and its parent is a pythonX.Y directory,
```
imjasonh, line 669: Should we look for `/usr/lib/pythonX.Y` specifically? Or would it make sense to find and process `/foo/bar/baz/pythonX.Y/site-packages` if we ever found it?

Either way this seems like a useful test case to document what paths are and aren't matched by this.

pkg/build/package.go
```diff
@@ -653,6 +653,66 @@ func generatePkgConfigDeps(pc *PackageBuild, generated *config.Dependencies) err
...
+		}
+
+		parent := filepath.Dir(path)
+		basename = filepath.Base(parent)
+		if !strings.HasPrefix(basename, "python") {
+			return nil
+		}
+
+		// This probably shouldn't ever happen, but lets check to make sure.
+		if !d.IsDir() {
```
imjasonh, line 683: This too, if I wrote a regular file called `/usr/lib/pythonX.Y/site-packages` we'd expect this to do nothing right?

imjasonh (commented):
I like this change! If I'm understanding correctly, it means that `py3-requests` etc don't need to declare any runtime dep on `python3`, or `python-3.12`, and when they get rebuilt with `python-3.13` they'll now magically gain a runtime dep on `python-3.13` specifically, instead of the sliding `python`, is that right?

kaniini:
> I like this change! If I'm understanding correctly, it means that py3-requests etc don't need to declare any runtime dep on python3, or python-3.12, and when they get rebuilt with python-3.13 they'll now magically gain a runtime dep on python-3.13 specifically, instead of the sliding python, is that right?

Yes.  This is intended to make future transitions less painful, by allowing older system python to be used for situations where not all packages are yet on the new system python.

pkg/build/package.go
kaniini: There is presently no infrastructure to write test cases for the SCA engine.  I do not believe this PR should be blocked on that.

pkg/build/package.go
kaniini: Yes, site-packages must always be a directory.  However, as previously stated, there is presently no infrastructure to write test cases for the SCA engine, and I do not believe this PR should be blocked on that.

pkg/build/package.go
kaniini: > Or would it make sense to find and process /foo/bar/baz/pythonX.Y/site-packages if we ever found it?

Yes, it would.  Applications can adjust the module search path, but the pyX.Y component is always present.

imjasonh (approved):

PR #789, opened 2023-10-25
Add Python docs linter

### LINKED ISSUE #778: Add Python docs linter
There should be a linter to detect Python docs directories that haven't been stripped or in incorrect locations.

Elizafox:
This checks if a doc or docs "package" was left in site-packages. This seems to be a recurring issue, so let's make a lint for it.

Fixes #778

### Functional Changes

Notes:

### Linter

### DIFF
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -109,10 +109,15 @@ var postLinterMap = map[string]postLinter{
 		FailOnError: false,
 		Explain:     "Verify that this package is supposed to be empty; if it is, disable this linter; otherwise check the build",
 	},
+	"pythondocs": postLinter{
+		LinterFunc:  pythonDocsPostLinter,
+		FailOnError: false,
+		Explain:     "Remove all docs directories from the package",
+	},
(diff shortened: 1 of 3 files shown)


pkg/linter/linter.go
```diff
@@ -109,10 +109,15 @@ var postLinterMap = map[string]postLinter{
 		FailOnError: false,
 		Explain:     "Verify that this package is supposed to be empty; if it is, disable this linter; otherwise check the build",
 	},
+	"pythondocs": postLinter{
+		LinterFunc:  pythonDocsPostLinter,
+		FailOnError: false,
+		Explain:     "Remove all docs directories from the package",
```
imjasonh, line 115: Is there a pipeline we should provide to split them out? `python/docs` or `split/pythondocs` or something? We have `split/manpages` today, for example.

Elizafox: Hmm. We could do it! I think we should do that in a separate request but I'm absolutely for it 100%. I think we should give some thought to that change.

imjasonh (approved):

PR #801, opened 2023-10-27
Add SBOM linter

Elizafox:
This ensures a package won't write to the SBOM directory inadvertently.

### Linter

### DIFF
--- a/pkg/linter/linter_test.go
+++ b/pkg/linter/linter_test.go
@@ -57,7 +57,7 @@ func Test_emptyLinter(t *testing.T) {
 	linters := cfg.Package.Checks.GetLinters()
 	assert.Equal(t, linters, []string{"empty"})
 	fsys := os.DirFS(dir)
-	lctx := NewLinterContext(cfg.Package.Name, fsys)
+	lctx := NewLinterContext(cfg.Package.Name, fsys, linter_defaults.LintersBuild)
 	called := false
 	assert.NoError(t, lctx.LintPackageFs(fsys, func(err error) {
 		called = true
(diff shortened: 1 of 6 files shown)


pkg/linter/linter_test.go
```diff
@@ -87,7 +87,7 @@ func Test_usrLocalLinter(t *testing.T) {
 	linters := cfg.Package.Checks.GetLinters()
 	assert.Equal(t, linters, []string{"usrlocal"})
 	fsys := os.DirFS(dir)
-	lctx := NewLinterContext(cfg.Package.Name, fsys)
+	lctx := NewLinterContext(cfg.Package.Name, fsys, linter_defaults.LintersBuild)
```
imjasonh, line 90: It's a minor smell that we have to update all these callsites because of this change. Is there a way to not have to?

imjasonh (approved):

pkg/linter/linter_test.go
Elizafox: I'm not sure. I want to put in the context what kind of lints we are running, and it's important so we don't run the sbom linter on APK's which will always fail. It could be arguably user error at that point, but I want to pass that information somehow to the linters so they are not inadvertently run.

We could put it in the linter declarations themselves what kind of linters they are as perhaps a bit flag? I don't know. It'd be code churn too. Maybe a bit cleaner?

Elizafox: Your call, though. Open to either solution :).

Elizafox: I actually think I have a better idea... moment...

imjasonh (approved):
Nice!

PR #802, opened 2023-10-28
Add .apko.lock.json to APK control section

jonjohnsonjr:
This generates the locked apko config for the guest container image that is used to build the APK and puts writes it as .apko.lock.json in the APK's control section.

We could alternatively stick this in the (or another) SBOM, but that has triggered false positives with security scanners in the past.

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -94,6 +94,8 @@ type Build struct {
 	FailOnLintWarning  bool
 
 	EnabledBuildOptions []string
+
+	resolvedApkoConfig apko_types.ImageConfiguration
 }
 
 var ErrSkipThisArch = errors.New("error: skip this arch")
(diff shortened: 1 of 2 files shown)


pkg/build/build.go
```diff
@@ -538,6 +540,26 @@ func WithPackageCacheDir(apkCacheDir string) Option {
...
+		return fmt.Errorf("resolving package versions: %w", err)
+	}
+	resolved := make([]string, 0, len(pkgs))
+	for _, pkg := range pkgs {
+		resolved = append(resolved, fmt.Sprintf("%s=%s", pkg.Name, pkg.Version))
+	}
+	ic.Contents.Packages = resolved
+	ic.Archs = []apko_types.Architecture{b.Arch}
+	ic.Contents.Repositories = append(ic.Contents.Repositories, b.ExtraRepos...)
+	ic.Contents.Keyring = append(ic.Contents.Keyring, b.ExtraKeys...)
```
mattmoor, line 556: This seems fine for now, but can I request that we hoist the resolution / locking logic into a shared place so that this and `tf-apko` and anywhere else we do this don't diverge?

It'd actually also make a great `apko` sub-comment to produce the resolved config (I know we have `show-packages`, but I'm thinking something focused around this logic specifically).

pkg/build/package.go
```diff
@@ -243,6 +243,14 @@ func (pc *PackageBuild) generateControlSection(ctx context.Context) ([]byte, err
 		return nil, fmt.Errorf("unable to build control FS: %w", err)
 	}
 
+	locked, err := json.Marshal(pc.Build.resolvedApkoConfig)
+	if err != nil {
+		return nil, fmt.Errorf("unable to marshal apko config: %w", err)
+	}
+	if err := fsys.WriteFile(".apko.lock.json", locked, 0644); err != nil {
```
mattmoor, line 250: (i know this is a bike-shed, but...) WDYT about `.LOCK.apko.json` of `.BUILDLOCK.apko.json` 🤔 

My thinking is that we use `.apko.yaml` elsewhere to indicate things are the apko dialect of yaml.  Here it's json, but similar idea.  The "build" qualifier is so it's clearer what it's the locked apko configuration of.

PR #803, opened 2023-10-30
Filter out noise from opening non-ELF files with the stripped linter

Elizafox:
Other errors are signs things are quite amiss and should be warned, but a bad magic number value can occur for harmless reasons like opening a script or some improperly set executable bit. So there's no point in reporting these errors.

No functional linter changes beyond this cosmetic change during build.

### DIFF
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -296,6 +296,16 @@ func strippedLinter(lctx LinterContext, path string, d fs.DirEntry) error {
 	}
 	defer reader.Close()
 
+	// Check magic
+	buf := make([]byte, 3)
+	if _, err = reader.Read(buf); err != nil {
+		return fmt.Errorf("Could not read from file: %v", err)
+	}
+	if string(buf) != "ELF" {
+		// This is probably just a script or something. Filter it for less noise.
+		return nil
+	}
+
 	// TODO(Elizafox) - .a object files, which needs archive handling support.
 
 	// XXX(Elizafox) - fs.Open doesn't support the ReaderAt interface so we copy it to a temp file.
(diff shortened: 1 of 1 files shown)


pkg/linter/linter.go
```diff
@@ -315,7 +318,14 @@ func strippedLinter(lctx LinterContext, path string, d fs.DirEntry) error {
 
 	file, err := elf.NewFile(tempfile)
 	if err != nil {
-		// We don't particularly care if this fails, it means it's probably not an ELF file
+		// XXX(Elizafox) - I hate Go's error handling and there's no better way.
+		// It literally gives us nothing but the string. Why Go... WHY...???
```
jonjohnsonjr, line 322: You can check if it's an `elf.FormatError`: https://pkg.go.dev/debug/elf#FormatError

Elizafox: The problem with that, is that then I need to check what *kind* of format error. Not all format errors are created equal. A bad magic number can be ignored, but should we ignore bad arch or other issues which could be part of larger issues?

kaniini: I don't follow.  debug/elf does not throw an error for "bad arch".

Elizafox: Not arch, class. My bad.

https://cs.opensource.google/go/go/+/refs/tags/go1.21.3:src/debug/elf/file.go;l=301

jonjohnsonjr: I think it's always going to be a bit of a heuristic if we're automatically detecting this stuff (rather than having it be explicit in the melange config. E.g. a package might contain a file with some randomly generated data that _happens_ to start with `.ELF` but isn't actually an ELF file, so we might fail with that Class error even though we should just ignore it.

Not sure if we'd rather than false positives (fail on invalid ELF files) or false negatives (ignore invalid ELF files)... probably false positives, right?

Might be better to just do this check yourself here (`Peek` the first 4 bytes with a `bufio.Reader`) rather than calling into `elf.NewFile` and having to decipher the error?

Elizafox: I guess I could. I'm not sure if that would be cleaner.

Elizafox:
Closing for now as @jonjohnsonjr is going to refactor a bunch of stuff and this will need to be rewritten.

PR #806, opened 2023-10-30
Add linters for documentation and object files

epsilon-phase:
The documentation linter checks for the presence of documentation files in non-documentation packages.

The object linter checks for the presence of intermediate object files in the generated packages.

### Functional Changes

Notes:
No breaking changes

### SCA Changes

Notes:
No breaking changes

### Linter

Notes:

### DIFF
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -65,11 +65,23 @@ var linterMap = map[string]linter{
 		FailOnError: false,
 		Explain:     "If this package is creating /dev nodes, it should use udev instead; otherwise, remove any files in /dev",
 	},
+	"documentation": linter{
+		LinterFunc:  documentationLinter,
+		LinterClass: linter_defaults.LinterClassApk | linter_defaults.LinterClassBuild,
+		FailOnError: false,
+		Explain:     "Place documentation into a separate package or remove it",
+	},
 	"opt": linter{
 		LinterFunc:  optLinter,
 		LinterClass: linter_defaults.LinterClassBuild | linter_defaults.LinterClassApk,
 		FailOnError: false,
 		Explain:     "This package should be a -compat package",
+	},
+	"object": linter{
+		LinterFunc:  objectLinter,
+		LinterClass: linter_defaults.LinterClassBuild | linter_defaults.LinterClassApk,
+		FailOnError: false,
+		Explain:     "This package contains intermediate object files",
 	},
 	"sbom": linter{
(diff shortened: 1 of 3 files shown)


pkg/linter/linter.go
```diff
@@ -157,6 +169,7 @@ var isVarEmptyRegex = regexp.MustCompile("^var/empty/")
 var isCompatPackageRegex = regexp.MustCompile("-compat$")
 var isObjectFileRegex = regexp.MustCompile(`\.(a|so|dylib)(\..*)?`)
 var isSbomPathRegex = regexp.MustCompile("^var/lib/db/sbom/")
+var isDocumentationFileRegex = regexp.MustCompile(`(?:READ(?:ME)?|CREDITS|\.(?:md|docx?|rst|[0-9]))$`)
```
Elizafox, line 172: I've seen README files named like `READ.ME` before. Maybe just add an optional qualifier for the dot?

I would also add `TODO`

It might be useful to check for .`[0-9][a-zA-Z]` because of e.g. `foo.3p`

Otherwise looks great.

pkg/linter/linter.go
```diff
@@ -178,6 +191,20 @@ func optLinter(_ LinterContext, path string, _ fs.DirEntry) error {
...
+func objectLinter(_ LinterContext, path string, _ fs.DirEntry) error {
+	if filepath.Ext(path) == ".o" {
+		return fmt.Errorf("Package contains intermediate object file '%s', this is usually wrong", path)
+	}
+
+	return nil
+}
+
+func documentationLinter(lc LinterContext, path string, _ fs.DirEntry) error {
+	if isDocumentationFileRegex.MatchString(path) && !strings.HasSuffix(lc.pkgname, "doc") {
```
Elizafox, line 203: I'd make this `-doc` otherwise looks great.

pkg/linter/linter.go
```diff
@@ -178,6 +191,20 @@ func optLinter(_ LinterContext, path string, _ fs.DirEntry) error {
 
 	return nil
 }
+func objectLinter(_ LinterContext, path string, _ fs.DirEntry) error {
+	if filepath.Ext(path) == ".o" {
+		return fmt.Errorf("Package contains intermediate object file '%s', this is usually wrong", path)
```
Elizafox, line 196: I would add a note saying how to resolve the issue. I would note that `.o` files should usually be removed.

Elizafox (approved):

PR #813, opened 2023-11-02
Add a new go pipeline to only manage deps

hectorj2f:

This PR creates a separate go pipeline to manage golang dependencies keeping the current go/build compatible with the initial logic. With this new go pipeline, we can now separate the treatment of go dependencies from the go build commands that many times is handled in a different way depending of the project, e.g. using make targets.

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:


hectorj2f:
I feel we could extend this pipeline to support `go mod edit` calls in the future.

PR #847, opened 2023-11-16
default --use-github=true

imjasonh:
This only does anything interesting for `melange convert python` but this opts us in to using `git-checkout` instead of `fetch`.

Also drop some unnecessary additions to the package.


luhring (dismissed):
🎉

luhring (approved):

PR #851, opened 2023-11-17
fix and continuously validate SBOMs

imjasonh:


pkg/sbom/implementation.go
```diff
@@ -350,7 +352,7 @@ func buildDocumentSPDX(spec *Spec, doc *bom) (*spdx.Document, error) {
 	}
 
 	spdxDoc := spdx.Document{
-		ID:      fmt.Sprintf("SPDXRef-DOCUMENT-%s", fmt.Sprintf("apk-%s-%s", spec.PackageName, spec.PackageVersion)),
+		ID:      "SPDXRef-DOCUMENT",
```
puerco, line 355: While we are at this, I recommend fixing the namespace (line 367) to make it unique to the package and deterministic. You can hash the previous value here to for it, eg:

```
https://spdx.org/spdxdocs/chainguard/melange/668e3da6b9af6045a9...,
```

pkg/sbom/implementation.go
```diff
@@ -90,6 +93,7 @@ func GenerateAPKPackage(spec *Spec) (pkg, error) {
 		Copyright:        spec.Copyright,
 		Namespace:        spec.Namespace,
 		Arch:             spec.Arch,
+		Originator:       "Organization: " + cases.Upper(language.English).String(spec.Namespace),
```
puerco, line 96: I think this is wrong, this should be the supplier:

```suggestion
		Supplier:       "Organization: " + cases.Upper(language.English).String(spec.Namespace),
```

The originator should be the maintainer of the package listed in the APK ([see the field here, for example](https://pkgs.alpinelinux.org/package/edge/main/x86/cjson)). Unfortunately, I think we are not populating those in wolfi (which is something we should fix).

imjasonh: https://pkg.go.dev/chainguard.dev/apko/pkg/sbom/generator/spdx#Package doesn't have a `Supplier`, should I add one?

imjasonh: Better yet, should I switch apko+melange to use something like https://pkg.go.dev/github.com/spdx/tools-golang@v0.5.3/spdx/v2/v2_3#Package ?

puerco: If we are changing the custom types, I would vote to switch the innards of all of our SBOM tools to protobom, that's why we created it (and it uses the spdx go tools to render the doc).

imjasonh: Sounds good. I'll leave that for now, but we I like that plan.

.github/workflows/e2e.yaml
```diff
@@ -28,21 +25,30 @@ jobs:
...
         make melange
-        ./melange keygen && mv melange.rsa.pub /etc/apk/keys
+        ./melange keygen
         ./melange build --pipeline-dir=pipelines examples/${{matrix.example}} --arch=x86_64 --empty-workspace
 
+    - name: Check SBOM Conformance
+      run: |
+        set -euxo pipefail
+        for f in packages/x86_64/*.apk; do
+          tar -Oxf "$f" var/lib/db/sbom > sbom.json
```
mattmoor, line 45: Can this have multiple files in it, or are we guaranteed that each APK only contributes its own SBOM?

.github/workflows/e2e.yaml
```diff
@@ -13,12 +13,9 @@ jobs:
   examples:
     name: build examples
     runs-on: ubuntu-latest
-    container:
-      image: alpine:latest
-      options: |
-        --cap-add NET_ADMIN --cap-add SYS_ADMIN --security-opt seccomp=unconfined --security-opt apparmor:unconfined
```
mattmoor, line 19: I'm curious why this is changing?  Should we switch it to Wolfi?

imjasonh: I wanted to `docker run` the NTIA checker image, and `alpine:latest` doesn't have docker. Rather than trying to figure out how to make that work for Alpine or Wolfi, I just dropped out of the container.

AFAICT we were running inside `alpine:latest` as a holdover from the days where `apko` etc needed `apk` to be available, and beyond that it was just a convenience for installing `bubblewrap`, which is also just an `apt-get` away.

.github/workflows/e2e.yaml
imjasonh: I don't think it can have multiple files, but I don't know for sure. If multiples are found they'll be concatenated in `sbom.json` and presumably something will fail.

If you have shell-fu to check that better than I am, please suggest it.

mattmoor (approved):

PR #870, opened 2023-12-04
cleanup: don't use pkg/errors

imjasonh:
https://github.com/pkg/errors is deprecated and archived for more than two years, since `fmt.Errorf` added support for `%w` in Go 1.13, [~three years ago](https://go.dev/blog/go1.13-errors).

This removes our usage of `pkg/errors` using this analyzer: https://github.com/zchee/go-analyzer/tree/main/pkgerrors


jonjohnsonjr (commented):
Drop from `go.mod` as well?

PR #876, opened 2023-12-04
update release to add some clarification regarding the homebrew

cpanato:
- update release to add some clarification regarding the homebrew


vaikas (approved):
Thank you so much!!!

PR #880, opened 2023-12-08
Set a default env var for GOMODCACHE.

dlorenc:
Melange supports a filesystem based cache, but it's not used anywhere AFAICT. To use it for Go, we've documented how to set this env var. Instead of making everyone set it in every pipeline and build, let's just set it to the correct path by default in melange itself.

This dramatically improves build times for me.

### Functional Changes

Notes:

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -841,16 +841,20 @@ func ParseConfiguration(configurationFilePath string, opts ...ConfigurationParsi
 	}
 
 	const (
-		defaultEnvVarHOME   = "/home/build"
-		defaultEnvVarGOPATH = "/home/build/.cache/go"
+		defaultEnvVarHOME       = "/home/build"
+		defaultEnvVarGOPATH     = "/home/build/.cache/go"
+		defaultEnvVarGOMODCACHE = "/var/cache/melange/gomodcache"
 	)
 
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -841,17 +841,21 @@ func ParseConfiguration(configurationFilePath string, opts ...ConfigurationParsi
...
+		defaultEnvVarGOMODCACHE = "/var/cache/melange/gomodcache"
 	)
 
-	if cfg.Environment.Environment["HOME"] == "" {
-		cfg.Environment.Environment["HOME"] = defaultEnvVarHOME
-	}
-	if cfg.Environment.Environment["GOPATH"] == "" {
-		cfg.Environment.Environment["GOPATH"] = defaultEnvVarGOPATH
+	var setIfEmpty = func(key, value string) {
+		if cfg.Environment.Environment[key] == "" {
```
luhring, line 850: This may be a pointless idea, but there's a chance it's better to use the `, ok =` style of check for map value emptiness rather than an empty string, in case we think the user would ever want to explicitly control the env variable to be empty at some point.

luhring (approved):

kaniini (approved):

PR #905, opened 2023-12-26
DNM: WIP dagger integration.

dlorenc:
### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:


pkg/container/dagger_runner.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	d.container = d.container.WithExec(args, dagger.ContainerWithExecOpts{})
+
+	var err error
+	// Force execution
+	d.container, err = d.container.Sync(ctx)
+	return err
+}
+
+// TestUsability determines if the Bubblewrap runner can be used
```
mattmoor, line 83: Bubblewrap?

pkg/container/dagger_runner.go
```diff
@@ -0,0 +1,194 @@
...
+		}
+	}
+
+	script := "[ -x /sbin/ldconfig ] && /sbin/ldconfig /lib || true"
+	d.container = d.container.WithExec([]string{"/bin/sh", "-c", script})
+	return nil
+}
+
+// TerminatePod terminates a pod if necessary.  Not implemented
+// for Bubblewrap runners.
```
mattmoor, line 145: Bubblewrap again

pkg/container/dagger_runner.go
```diff
@@ -0,0 +1,194 @@
...
+func (d *daggerRunner) TerminatePod(ctx context.Context, cfg *Config) error {
+	return nil
+}
+
+func (d *daggerRunner) WorkspaceTar(ctx context.Context, cfg *Config) (io.ReadCloser, error) {
+	return nil, nil
+}
+
+// WorkspaceTar implements Runner
```
mattmoor, line 154: Wrong method name

pkg/container/dagger_runner.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	img, err := apko_oci.BuildImageFromLayer(layer, bc.ImageConfiguration(), creationTime, arch, bc.Logger())
+	if err != nil {
+		return "", err
+	}
+
+	ref, err := name.ParseReference("melange:latest")
```
mattmoor, line 184: I could see having a fixed name here creating races in the load/run phase when performing multiple builds locally.

pkg/container/dagger_runner.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	tarPath := filepath.Join(d.tmpDir, imageTarName)
+	if err := tarball.WriteToFile(tarPath, ref, img); err != nil {
```
jonjohnsonjr, line 190: This will be really inefficient. I need to look to see if dagger exposes a better interface.

dlorenc: Yea, I know. I think the only better option is to round trip through a registry. I looked and couldn't find much else.

pkg/container/dagger_runner.go
dlorenc: I think this is just at throwaway name but could be wrong.

pkg/container/dagger_runner.go
dlorenc: OR change the entire setup so dagger also invokes apko for us.

pkg/container/dagger_runner.go
```diff
@@ -0,0 +1,194 @@
...
+	var err error
+	// Force execution
+	d.container, err = d.container.Sync(ctx)
+	return err
+}
+
+// TestUsability determines if the Bubblewrap runner can be used
+// as a container runner.
+func (d *daggerRunner) TestUsability(ctx context.Context) bool {
+	return true
```
joshrwolf, line 86: I think the sdk exposes a simple health check somewhere we could leverage here

PR #909, opened 2024-01-02
meson/configure: don't download subprojects by default

imjasonh:
https://mesonbuild.com/Subprojects.html#commandline-options

> * **--wrap-mode=nodownload**
> Meson will not use the network to download any subprojects or fetch any wrap information. Only preexisting sources will be used. This is useful (mostly for distros) when you want to only use the sources provided by a software release, and want to manually handle or provide missing dependencies.

### DIFF
--- a/pkg/build/pipelines/meson/configure.yaml
+++ b/pkg/build/pipelines/meson/configure.yaml
@@ -18,4 +18,9 @@ pipeline:
   - runs: |
       meson setup . ${{inputs.output-dir}} \
         --prefix=/usr \
+        # Don't download subprojects by default. We want to use only the source
+        # provided by the project, and if any subprojects are needed we should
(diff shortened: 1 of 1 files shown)


pkg/build/pipelines/meson/configure.yaml
```diff
@@ -18,4 +18,9 @@ pipeline:
   - runs: |
       meson setup . ${{inputs.output-dir}} \
         --prefix=/usr \
+        # Don't download subprojects by default. We want to use only the source
```
kaniini, line 21: I believe the comment will separate the shell tokens here.  As an alternative, lets put it above the `runs` statement.

kaniini (changes requested):

kaniini (approved):

PR #910, opened 2024-01-02
Add --test-package-append that you can specify extra test packages for each test.

vaikas:
Add a new flag `--test-package-append` that allows you to omit specifying packages in the `yaml` file, and instead specify
them from the command line. They are handy if you have a common package that all tests require so that you don't have
to keep specifying them in your yaml.

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

### DIFF
--- a/.github/workflows/melange-test-pipelines.yaml
+++ b/.github/workflows/melange-test-pipelines.yaml
@@ -47,6 +47,13 @@ jobs:
         package:
           - php-8.2-msgpack
           - py3-pandas
+          # The ones with -nopkg are packages which do not specify test
+          # packages, and they are added to the test environment by using flags.
+          # They make sure the flag `--test-package-append` works.
+          # We also do not specify the test package in the test file, so that
+          # we test that the main package gets correctly pulled from the file.
+          - php-8.2-msgpack-test-nopkg
+          - py3-pandas-test-nopkg
 
     steps:
       # Grab the melange we uploaded above, and install it.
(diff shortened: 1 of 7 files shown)


vaikas:
Hm, what the heck am I doing wrong here:
https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution#overview

Their example is:
```
if: ${{ ! startsWith(github.ref, 'refs/tags/') }}
```

Yet the yaml file is deemed invalid here:
https://github.com/chainguard-dev/melange/actions/runs/7389375643

```
The workflow is not valid. .github/workflows/melange-test-pipelines.yaml (Line: 80, Col: 13): Unexpected symbol: '${{'. Located at position 13 within expression: ! endsWith( ${{ matrix.package
```

I'm clearly missing something:
https://github.com/chainguard-dev/melange/pull/910/files#diff-1b8e5d0e09c96ec54aca51c55dbbb57a171052b14f3a289ef3f93af244ed4e1aR80
```
        if: ${{ ! endsWith( ${{ matrix.package }}, '-nopkg' ) }}
```
Maybe I can't eval the ${{ matrix.package }} within endsWith?

rawlingsj:
does removing the dollar squgilies from around matrix help?
i.e.

```
if: ${{ ! endsWith(matrix.package, '-nopkg') }}
```

vaikas:
> matrix.package

Thanks! Looks like if I just use `false == ...` it works. Well, fails for different reason, but progress 😆

vaikas:
> > matrix.package
> 
> Thanks! Looks like if I just use `false == ...` it works. Well, fails for different reason, but progress 😆

Thanks @rawlingsj removing the `${{` around matrix. seems to have done the trick!!! 🙇

imjasonh (approved):

PR #935, opened 2024-01-17
Allow vendored pkgconfig deps

jonjohnsonjr:
Some packages vendor multiple versions of the same dependency. For example, neon contains several version of postgres. These versions each had a pkgconfig directory that provided different versions of the same thing. These overlapping provides cause apk to fail when trying to install it.


jonjohnsonjr:
Weird: https://github.com/wolfi-dev/os/pull/11326

imjasonh:
Is this something we can add a test for? 🙏

imjasonh (approved):

PR #939, opened 2024-01-19
add e2e test that packages can be installed with apk

imjasonh:


stormqueen1990 (approved):

imjasonh:
This makes the glibc example build from taking [~9m](https://github.com/chainguard-dev/melange/actions/runs/7576343430/job/20634953127) up to [~11m](https://github.com/chainguard-dev/melange/actions/runs/7585273740/job/20660770948?pr=939), since it installs a bunch of locale subpackages. I don't think it's a huge deal, but it's worth tracking in case this makes CI take a long time.

stormqueen1990 (approved):

PR #942, opened 2024-01-19
Fail if unknown variable is used in substitution

jonjohnsonjr:
Tested this with this diff to crane.yaml:

```
diff --git a/crane.yaml b/crane.yaml
index 3f5c32dc2..9d458809f 100644
--- a/crane.yaml
+++ b/crane.yaml
@@ -1,13 +1,15 @@
 package:
   name: crane
   version: 0.18.0
-  epoch: 0
+  epoch: 1
   description: Tool for interacting with remote images and registries.
   copyright:
     - license: Apache-2.0
   dependencies:
     runtime:
       - ca-certificates-bundle
+    provides:
+      - foo=${{package.fullversion}}
 
 environment:
   contents:
```

```
ℹ️            | error during command execution: failed to apply replacement to provides "foo=${{package.fullversion}}": variable package.fullversion not defined
```

### DIFF
--- a/pkg/build/pipelines/autoconf/configure.yaml
+++ b/pkg/build/pipelines/autoconf/configure.yaml
@@ -16,6 +16,11 @@ inputs:
       The GNU triplet which describes the build system.
     default: ${{host.triplet.gnu}}
 
+  opts:
+    description: |
+      Options to pass to the ./configure command.
+    default: ''
+
 pipeline:
   - runs: |
       cd ${{inputs.dir}}
--- a/pkg/build/pipeline.go
+++ b/pkg/build/pipeline.go
@@ -189,7 +189,7 @@ func validateWith(data map[string]string, inputs map[string]config.Input) (map[s
 	}
 
 	for k, v := range inputs {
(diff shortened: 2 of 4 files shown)


pkg/build/pipelines/autoconf/configure.yaml
```diff
@@ -16,6 +16,11 @@ inputs:
       The GNU triplet which describes the build system.
     default: ${{host.triplet.gnu}}
 
+  opts:
+    description: |
+      Options to pass to the ./configure command.
+    default: ''
+
```
ajayk, line 23: nice , was this intended to be part of the same PR

jonjohnsonjr: Yeah the e2e tests fail without this.

jonjohnsonjr: (If you expand the rest of the file, you can see we are using `${{inputs.opts}}` even though it's not defined.)

imjasonh (approved):
Thank you!!

PR #956, opened 2024-01-23
Take advantage of Octo STS to publish homebrew updates.

mattmoor:
Once the trust policy lands here: https://github.com/chainguard-dev/homebrew-tap/pull/53

This change will enable the release workflow to federate with the Octo STS app to create tokens in accordance with the trust policy and avoid the use of PATs.


mattmoor:
Marking this as a draft until the upstream PR lands.

cpanato (approved):

PR #960, opened 2024-01-25
Fix aws-c-s3 SCA

jonjohnsonjr:
libaws-c-s3.so.1.0.0 has:

```
  SONAME       libaws-c-s3.so.0unstable
```

We were generating:

```
  provides = so:libaws-c-s3.so.0unstable=0unstable
```

This is invalid, apparently, so now we generate:

```
  provides = so:libaws-c-s3.so.0unstable=0
```


imjasonh (approved):

kaniini:
This seems indicative of a packaging error, too.  The SONAME should match the file name that gets installed on disk.

Issue #945, opened 2024-01-21
var-transform with '_' in replace


mesaglio:
Hi! @Elizafox 
I dont know if [this functionality](Elizafox) is not working as spected, or is not for this case.

I tested locally with the [following test](https://github.com/mesaglio/melange/blob/1689798d6c9d95f71b3f10f92c6fb7fb899b8ec4/pkg/renovate/bump/bump_test.go#L31).
This generate this error
```
2024/01/19 11:43:58 INFO   evaluated: http://127.0.0.1:52223/wine/cheese/cheese-7.0.1+1.tar.gz
    bump_test.go:118:
        	Error Trace:	/Users/jmesaglio/Documents/repos/melange/pkg/renovate/bump/bump_test.go:118
        	            				/usr/local/go/src/net/http/server.go:2136
        	            				/usr/local/go/src/net/http/server.go:2938
        	            				/usr/local/go/src/net/http/server.go:2009
        	            				/usr/local/go/src/runtime/asm_arm64.s:1197
        	Error:      	Not equal:
        	            	expected: "/wine/cheese/cheese-7.0.1+1.tar.gz"
        	            	actual  : "/wine/cheese/cheese-7.0.1.tar.gz"

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-/wine/cheese/cheese-7.0.1+1.tar.gz
        	            	+/wine/cheese/cheese-7.0.1.tar.gz
        	Test:       	TestBump_versions/major_minor_patch.yaml#01
```

I undestand the [var transformation](https://github.com/chainguard-dev/melange/blob/main/pkg/renovate/bump/testdata/major_minor_patch.yaml#L11), generate `same-package-version` with `7.0.1` value.
```yaml
var-transforms:
  - from: ${{package.version}}
    match: (\d.\d.\d)
    replace: $1
    to: same-package-version
```

We found this error en melanga package when uses `fetch` node.

For example:
```yaml
package:
  name: temurin-11-jdk
  version: "11.0.21.9"
  epoch: 0
  description: "Adoptium Temurin 11 JDK"
  copyright:
    - license: Apache-2.0
  dependencies:
    runtime:

environment:
  contents:
    keyring:
      - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
    repositories:
      - https://packages.wolfi.dev/os
    packages:
      - wolfi-base

var-transforms:
  - from: ${{package.version}}
[...]

PR #973, opened 2024-02-01
version-transform and var-transform documentation

mesaglio:
This have more documentation 🎊 
- Add other case to `var-transform` (`${1}_`)
- Add `version-transform` to update section

cc @vaikas

### DIFF
--- a/docs/UPDATE.md
+++ b/docs/UPDATE.md
@@ -73,10 +73,31 @@ update:
 
 Some projects create tags than are not compliance with apk format. You can manipulate this with regex on `version-transform` section.
 
+Example:
+
+```patch
+ package:
+   name: owfs
+-  version: 3.2.3
++  version: 3.2p4
+   epoch: 0
+```
+
+With this version
 ```yaml
 update:
   enabled: true
   version-transform:
-    - match: \+(\d+)$
-      replace: -${1}
+    - match: p(\d+)$
+      replace: .${1}
+```
+
+the next update pr will be like:
+
+```patch
+ package:
+   name: owfs
+-  version: 3.2.3
++  version: 3.2.4
+   epoch: 0
 ```
--- a/docs/UPDATE.md
+++ b/docs/UPDATE.md
@@ -68,3 +68,15 @@ update:
     - "ignore_me*"
     - "*ignore_me"
 ```
+
+## Version Transform
+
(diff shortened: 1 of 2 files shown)


docs/UPDATE.md
```diff
@@ -68,3 +68,15 @@ update:
     - "ignore_me*"
     - "*ignore_me"
 ```
+
+## Version Transform
+
+Some projects create tags than are not compliance with apk format. You can manipulate this with regex on `version-transform` section.
+
```
vaikas, line 75: Can you give an example here from the other PR so people don't have to go look around for it?

vaikas (commented):
Thanks!

docs/UPDATE.md
mesaglio: The pr I mentioned in the issue is private, but I create the patch of the changes. What do you think?

vaikas: Sorry, I meant the issue: https://github.com/chainguard-dev/melange/issues/945

Where you could put the relevant details like:

version: "11.0.21.9"

And that needs to be turned into a filename (show the filename example):
`https://github.com/adoptium/temurin11-binaries/releases/download/jdk-${{vars.mangled-name-jdk}}/OpenJDK11U-jdk_x64_linux_hotspot_${{vars.mangled-version-binary}}.tar.gz
`
where you should then what the 11.0.21.9 should turn into for `vars.mangled-name-jdk` as `11.0.21+9` and so forth.

mesaglio: Sorry Im not understand :grimacing:.
This details of `11.0.21+9` is not documented [here](https://github.com/chainguard-dev/melange/blob/main/docs/VAR-TRANSFORMS.md?plain=1#L15), with the openjdk version?
One time you modify with `version-transform` you can manuplate the `package.version` with the `var-transforms`

vaikas (approved):

docs/UPDATE.md
vaikas: I think it's great as is. Thank you so much for doing this!!

mesaglio:
We need other change to merge?

mesaglio:
🏓

vaikas:
No, It's been approved, go ahead and merge it 😄

PR #982, opened 2024-02-05
Fix race condition in log monitoring

jonjohnsonjr:
For some reason we were using os.Pipe instead of io.Pipe for an in-process pipe. I've switched those over to using io.Pipe so we can CloseWithError.

Also, switched to using errgroup.Group instead of manual channel signaling.

Now, when we hit an error, we actually log that error instead of the line just before that error.


imjasonh (dismissed):
❤️

PR #987, opened 2024-02-06
Experimental Dagger Runner

jonjohnsonjr:
This just rebases Dan's PR (https://github.com/chainguard-dev/melange/pull/905) but it seems to kinda work.

For very long builds, this seems to be a big improvement.

To use it, set `--runner=experimentaldagger`.

I reserve the right to remove this in the future, but only when we have a better solution.


pdeslaur (approved):
I want it!

PR #997, opened 2024-02-07
Add --die-with-parent to bwrap flags

jonjohnsonjr:
This prevents us from failing when background processes have not exited.


joshrwolf:
any chance this is going to bite us? https://github.com/containers/bubblewrap/issues/529

jonjohnsonjr:
We already set `--unshare-pid`, so I think we're good unless you mean something else in that thread?

joshrwolf:
> We already set --unshare-pid, so I think we're good unless you mean something else in that thread?

duh, I missed it!

joshrwolf (approved):

imjasonh (approved):

PR #1026, opened 2024-02-15
Add a --trace flag to melange build

jonjohnsonjr:
We could do this globally but I'm going to add this here for now as an experiment. This takes a filename and writes an otel JSON trace to it.


imjasonh (approved):
If we use this in more places (I could see wanting it in tf-apko and wolfictl as well?) we should make it an easily shareable method somewhere.

PR #1052, opened 2024-02-29
Fix job control with interactive bubblewrap

jonjohnsonjr:
This flag breaks job control, which only matters during `--interactive` debugging.


imjasonh (approved):

smoser (commented):
this looks great. thank you.

PR #1058, opened 2024-03-03
add go/bump as a default pipeline

willswire:
### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

This PR adds the `go/bump` [pipelined](https://github.com/wolfi-dev/os/blob/main/pipelines/go/bump.yaml) used in wolfi-dev/os to melange, so that local builds can use this pipeline too. For example, see [here](https://github.com/willswire/kube-webhook-certgen/blob/5dc8bfcb5fb1960d0ce1395a4a51278de57b26c2/melange.yaml#L28).


pkg/build/pipelines/go/bump.yaml
```diff
@@ -0,0 +1,46 @@
...
+    default: false
+    description: Show the difference between the go.mod file before and after the bump
+  tidy-compat:
+    description: "Set the go version for which the tidied go.mod and go.sum files should be compatible"
+    default: ""
+
+pipeline:
+  - runs: |
+      # install the latest version of gobump
+      go install github.com/chainguard-dev/gobump@latest
```
xnox, line 32: Given build systems are expected to be rigid, we do want to use a packaged version of gobump - rather than building new one on every go package; and/or arbitrary versioned one.

Imho this should still be the same as the stock wolfi-dev/os pipeline and add gobump in the environment.

xnox: Fixed this, but will need re-review by someone else.

jonjohnsonjr (approved):

PR #1066, opened 2024-03-05
Update pombump.yaml

pdeslaur:
The default name file name of `pombump.yaml` is causing issues, since our CI build thinks it could be a package


vaikas (approved):
Thanks!

PR #1072, opened 2024-03-07
Add package.major-minor substitution

smoser:
I'm interested in feedback on this. 

I came to wanting a variable for major.minor when looking at [python-3.12](https://github.com/wolfi-dev/os/blob/main/python-3.12.yaml) which repeats the string 3.12 eleven times.

I can accept responses like "3.12 is more easily readable than ${{package.major-minor}}".

interested in thoughts.

### DIFF
--- a/pkg/build/pipeline.go
+++ b/pkg/build/pipeline.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


imjasonh:
Overall this seems fine. It might be possible to do this with a var substitution, but it would be gross to duplicate those 4 lines across 11 packages (and growing).

Could you add a simple test that demonstrates how this works on versions like `1.2.3` and maybe `1.2.3.4`, and especially something like `version: 12345` (I don't know if that's valid in the first place, but that would also be good to test)

smoser:
> Overall this seems fine. It might be possible to do this with a var substitution, but it would be gross to duplicate those 4 lines across 11 packages (and growing).

Yeah, i saw that I could do that, but the duplicated 4 lines were annoying.

> Could you add a simple test that demonstrates how this works on versions like 1.2.3 and maybe 1.2.3.4, and especially something like version: 12345 (I don't know if that's valid in the first place, but that would also be good to test)

yeah, can do that.

smoser:
I'm gonna close this for now, maybe someday I'll re-open.  I wasn't planning on using it right yet, and I don't like having code that no one uses.

PR #1076, opened 2024-03-10
remove files from SBOM

imjasonh:


imjasonh:
The pulumi image's SBOM is currently 18.4 MB. Without files it's 5.4 MB.

```
cosign download attestation \
  --platform=linux/amd64 \
  --predicate-type=https://spdx.dev/Document \
  cgr.dev/chainguard/pulumi | jq -r .payload | base64 -d | jq -r .predicate | jq -r 'del(.files)' | wc -c 
```

Python goes from 2.05 MB to 600KB

xnox:
I like this. Does `wolfi scan` still work correctly? Did it rely on these file manifests to detect which files came from which APK?

jspeed-meyers:
Less code (usually) == better :)

imjasonh:
> I like this. Does `wolfi scan` still work correctly? Did it rely on these file manifests to detect which files came from which APK?

Good questions.

`wolfictl scan` either scans with Syft, or accepts a previously-Syft-produced SBOM: https://github.com/wolfi-dev/wolfictl/blob/main/pkg/cli/scan.go And I can't see anything around there that operates on files at all, so I think we're okay.

`apk audit` also operates on the installed DB, not on the SBOMs (which Alpine packages don't contain at all).

pdeslaur (approved):

PR #1079, opened 2024-03-11
sca: allow co-installable pythons

xnox:

Skip generating python3~N.M dependency on python-3.11 package itself (and similar). Thus allowing to co-install python-3.11 and python-3.12, as they got finally correctly streamed by @smoser 

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

  * built python-3.11 & python-3.12 and they both co-install now
  * dropped manual python3 depends from py3-requests, rebuild it with this melange, and python3~3.12 is correctly added
  * note, the generated dependency is likely to change soon from python3~3.N to python-3.N

### DIFF
--- a/pkg/sca/sca.go
+++ b/pkg/sca/sca.go
@@ -462,7 +462,7 @@ func generatePkgConfigDeps(ctx context.Context, hdl SCAHandle, generated *config
 	return nil
 }
 
-var pythonNamesRegexp = regexp.MustCompile(`^python-3.[0-9]*$`)
+var pythonNamesRegexp = regexp.MustCompile(`^python-3\.\d+$`)
 
 // generatePythonDeps generates a python3~$VERSION dependency for packages which ship
 // Python modules.
--- a/pkg/sca/sca.go
+++ b/pkg/sca/sca.go
@@ -462,6 +462,8 @@ func generatePkgConfigDeps(ctx context.Context, hdl SCAHandle, generated *config
 	return nil
 }
 
+var pythonNamesRegexp = regexp.MustCompile(`^python-3.[0-9]*$`)
+
 // generatePythonDeps generates a python3~$VERSION dependency for packages which ship
 // Python modules.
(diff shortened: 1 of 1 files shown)


imjasonh (approved):
Our SCA code has been woefully under-tested since forever, given how magical and secretly load-bearing it is. If there's a test case you can think of that would be easy to add, I'd really appreciate it. If not, that's okay too, there's a lot we should improve about the testing story here.

pkg/sca/sca.go
```diff
@@ -462,6 +462,8 @@ func generatePkgConfigDeps(ctx context.Context, hdl SCAHandle, generated *config
 	return nil
 }
 
+var pythonNamesRegexp = regexp.MustCompile(`^python-3.[0-9]*$`)
```
ajayk, line 465: minor nit 
```suggestion
var pythonNamesRegexp = regexp.MustCompile(`^python-3\.\d+$`)
```

smoser:
> [smoser](https://github.com/smoser) dismissed [imjasonh](https://github.com/imjasonh)’s [stale review](https://github.com/chainguard-dev/melange/pull/1079#pullrequestreview-1928231924) via [3a76748](https://github.com/chainguard-dev/melange/commit/3a76748cda66d7a828333d03e9e908e2a896b715)

I didn't really dismiss it. I agree with the need for test.  I just applied @ajayk 's suggestion on the regex.

smoser (approved):
I'm +1 on this. I do wish there was an easy way to test.

xnox:
> Our SCA code has been woefully under-tested since forever, given how magical and secretly load-bearing it is. If there's a test case you can think of that would be easy to add, I'd really appreciate it. If not, that's okay too, there's a lot we should improve about the testing story here.

a test case here would be .... just rebuild two pythons and co-install them. But it is very heavy, we can instrument a mock package that is called python-3.11 and ships a site-package, and something called python-soundslikearegularmodule and check that it does generate the expected provides.

Something to improve as we add automatic generation of better provides and depends.

PR #1083, opened 2024-03-12
Add filter-prefix option to update config.

smoser:
There is not currently a way to include only matching versions from the release update service

This adds filter-prefix entry to Update section of melange.yaml. Specific use case for filter-prefix is a streamed package (ruby).

After the corresponding change to make use of this filter lands in wolfi, then we can update ruby-3.2 to enable the release monitor like with:

 > update:
 >  enabled: true
 >  manual: true
 >  filter-prefix: "3.2."
 >  release-monitor:
 >    identifier: 4223

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -459,6 +459,8 @@ type Update struct {
 	GitHubMonitor *GitHubMonitor `json:"github,omitempty" yaml:"github,omitempty"`
 	// The configuration block for transforming the `package.version` into an APK version
 	VersionTransform []VersionTransform `json:"version-transform,omitempty" yaml:"version-transform,omitempty"`
+	// Prefix filter to apply when searching releases
+	FilterPrefix string `json:"filter-prefix,omitempty" yaml:"filter-prefix,omitempty"`
 }
 
 // ReleaseMonitor indicates using the API for https://release-monitoring.org/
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -459,6 +459,8 @@ type Update struct {
 	GitHubMonitor *GitHubMonitor `json:"github,omitempty" yaml:"github,omitempty"`
 	// The configuration block for transforming the `package.version` into an APK version
 	VersionTransform []VersionTransform `json:"version-transform,omitempty" yaml:"version-transform,omitempty"`
+	// Prefix filter to apply when searching releases
+	FilterPrefix string `json:"filter-prefix,omitempty" yaml:"filter-prefix,omitempty"`
```
rawlingsj, line 463: I wonder if putting it under the `update` block will cause confusion with the existing filter-prefix that is under the `github` block?  I.e. would it be better if we move this under the `release-monitor` block?

smoser: Or maybe Deprecate the 'TagFilterPrefix' in github block?  filtering strings isn't really "provider specific".

I'm happy to do whatever you think is right.

smoser:
I was just going through my open PRs. It feels like this is still useful.  @rawlingsj I'm fine to close it or update it.  let me know what you think is right.

smoser:
note to self, if this lands, need update to wolfctl also (as seen  https://github.com/wolfi-dev/wolfictl/pull/692)

PR #1102, opened 2024-03-22
Python/sca updates

smoser:
There are 2 changes here, I'm happy to split them into separate PRs if requested.

1. instead of generating python3~$VERSION generate python-3.$VERSION-base . this way the python library does not depend on /usr/bin/python, but rather /usr/bin/python3.$VERSION
2. look at shebang programs and generate a cmd:<prog> for each of those.

### DIFF
--- a/pkg/sca/sca.go
+++ b/pkg/sca/sca.go
@@ -520,10 +520,8 @@ func generatePythonDeps(ctx context.Context, hdl SCAHandle, generated *config.De
 		}
 	}
 
-	// We use the python3 name here instead of the python-3 name so that we can be
-	// compatible with Alpine and Adelie.  Only Wolfi provides the python-3 name.
-	log.Infof("  found python module, generating python3~%s dependency", pythonModuleVer)
-	generated.Runtime = append(generated.Runtime, fmt.Sprintf("python3~%s", pythonModuleVer))
+	log.Infof("  found python module, generating python-%s-base dependency", pythonModuleVer)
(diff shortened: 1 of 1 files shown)


smoser:
Please give this some thought... While I think both the changes are good, i can see how they both might cause some havoc.

imjasonh (dismissed):

pkg/sca/testdata/shebang-test.yaml
```diff
@@ -0,0 +1,47 @@
...
+      echo hi from bash
+      EOF
+      
+      wbin "python-via-env" <<"EOF"
+      #!/usr/bin/env python3
+      print("hello world")
+      EOF
+      
+      wbin "python-straight" <<"EOF"
+      #!/usr/bin/python3
```
xnox, line 42: hm..... I hoped we generate most of our scripts with python3.12 shebangs.
Or if we do have scripts with python3 shebang + modules => we need a dep on python-3.12 (which would allow executing the module, and has the python3 symlink).

smoser: I think we do generate "most of our scripts" with usr/bin/python3.12 shebang , and we absolutely should.

The change here adds a 'cmd:<thing>' for the thing in the shebang.  So that should work with 'python3.12', or 'python3' or 'bash', as the package that has that program will provide the 'cmd:thing' correctly.  Am I missing something there?

smoser:
I'm splitting the 'cmd:shbang' commit out to #1110  since it seems maybe more involved based on the test case failures.

imjasonh (approved):

PR #1110, opened 2024-03-26
Find shbangs to generate depends

smoser:
If a file in usr/bin or bin contains '#!/usr/bin/env <prog>' or '#!/path/to/prog', then add a 'cmd:<prog>' depends

The go-apk update is to handle empty file reads.

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:


pkg/sca/sca.go
```diff
@@ -546,6 +546,80 @@ func sonameLibver(soname string) string {
...
+	} else if err != nil {
+		return "", err
+	}
+
+	if buf[0] != '#' && buf[1] != '!' {
+		return "", nil
+	}
+	toks := strings.Fields(string(buf[2 : blen-2]))
+
+	// if it is not /usr/bin/env, then use it.
```
imjasonh, line 567: Does it matter if it's `/bin/env` or `/sbin/env` or `/tmp/env` or `/definitely/not/env`? I don't know enough about shebangs or `env` to know if we need to handle those cases.

imjasonh (approved):
This lgtm but @jonjohnsonjr might have a better set of eyes here than me.

pkg/sca/sca.go
smoser: I'm happy to limit it to /usr/bin/env . I don't know that I've ever seen use of 'env' from any other PATH.

I'll update to just do that, and maybe even warn or error if '#!/some/other/path/env'

smoser:
The 2 issues here are:

1. [s3cmd](https://github.com/chainguard-dev/melange/actions/runs/8429532808/job/23084001450) which provides a `python` program to run.  It can't find anything that provides 'cmd:python3':

       ERROR: unable to select packages:
        cmd:python3 (no such package):
           required by: s3cmd-2.4.0-r1[cmd:python3]

2. [ncurses](https://github.com/chainguard-dev/melange/actions/runs/8429532808/job/23084001844) ncurses-dev subpackage gets a 'cmd:sh' depend from usr/bin/ncursesw6-config which cannot be resolved.

       ERROR: unable to select packages:
         cmd:sh (no such package):
           required by: ncurses-dev-6.4_p20231125-r1[cmd:sh]

I thought that melange would automatically add the 'provides cmd:<thing>' for busybox (or bash) and python-3.X packages, but it apparently does not.

The python-3.X shoudl be straight forward enough, but bash or busybox and
/bin/sh are both non-trivial . they both provide a (dynamicly added?) symlink /bin/sh -> themselves.

pkg/sca/sca.go
smoser: that is done. only /usr/bin/env is treated specially.

smoser:
Issues are resolved now by:
a. ignoring '/bin/sh'
b. ignoring /usr/bin/python3 (still allow /usr/bin/python3.x)

jonjohnsonjr (approved):

PR #1120, opened 2024-04-02
Go fips deps

xnox:

Automatically generate runtime deps for go binaries that smell like go-fips ones. Specifically check if CGO is on, and boringcrypto experiment is on, and in that case generate runtime dependencies on openssl libraries, fips config, and fips provider.

This will reduce number of by-hand declared runtime dependencies on the go-fips compiled binaries.

### Functional Changes

It reduces manually specified runtime dependencies in many golang based -fips packages.

Notes:

### SCA Changes

The `openssl-config-fipshardened` is Chainguard-specific, however in practice it means "anything that will get one correctly configured OpenSSL fips provider" with no restriction as to which OpenSSL it is, or which fips provider build it is. Non-chainguard APK distributions can call that thing whatever they want, and add a virtual providers for it.
If this is deemed too specific, maybe we should add a more generic provides name; and use that in SCA; and have that match in Chainguard i.e. something fake like `so:openssl-fips-provider` and then use that everywhere.

W.r.t. detecting correct toolchain this should work for many go-fips builds from Chainguard, Microsoft, RHEL, Canonical. Thus it is universall-ish enough.

Notes:

### Linter

Notes:

Test case added for new behavior using a binary that mimics a fips go binary close enough.

Developed on x86_64, should work on arm64 too.

### DIFF
--- a/pkg/sca/testdata/go-fips-bin/go-fips-bin.yaml
+++ b/pkg/sca/testdata/go-fips-bin/go-fips-bin.yaml
@@ -1,0 +1,30 @@
+# SPDX-FileCopyrightText: 2022 Chainguard, Inc
+# SPDX-License-Identifier: Apache-2.0
+#
+# This is a sample configuration file to demonstrate how to build a software
+# project using melange's built-in go/install pipeline.
+#
+# For more information about melange's built-in golang support check out:
+#
+#
+# For an equivalent pipeline that uses go/install to build the same project
+# please see go-install.yaml in this directory.
+package:
+  name: go-fips-bin
+  version: v0.0.1
+  epoch: 0
+  description: "A tiny fips-like go binary"
+
+environment:
+  contents:
+    keyring:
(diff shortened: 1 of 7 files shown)


xnox:
TODO: somebody on a Mac, please try `make test` and let me know if it works correctly.

pkg/sca/testdata/go-fips-bin/go-fips-bin.yaml
```diff
@@ -0,0 +1,30 @@
+# SPDX-FileCopyrightText: 2022 Chainguard, Inc
+# SPDX-License-Identifier: Apache-2.0
+#
+# This is a sample configuration file to demonstrate how to build a software
+# project using melange's built-in go/install pipeline.
+#
+# For more information about melange's built-in golang support check out:
+#
```
vaikas, line 8: super nit: looks like this is missing a link?

vaikas (approved):

PR #1142, opened 2024-04-17
pipelines/go: add back symbols tables

xnox:
Symbols tables in go binaries can take up about 10% of the binary size. However, they are useful for cryptography audits as well as for operational govulncheck. So add them back in.

### Functional Changes

Notes: Rebuilt a few public and fips go packages with this pipeline change in place

### SCA Changes

Notes:

### Linter

Notes:


luhring (approved):
Nice! Thanks!

luhring (approved):

jonjohnsonjr (approved):

PR #1146, opened 2024-04-19
verify SPDX SBOMs using spdx-tools-java

imjasonh:
This runs the official SPDX validator on SBOMs built in our e2e tests, using the package we build in Wolfi.

https://github.com/spdx/tools-java

There's more to do after this, downstream of this, but this is a start.


mattmoor (approved):
You could also use Wolfi-act?

PR #1148, opened 2024-04-20
Fix sca detection case for env with multiple arguments.

dlorenc:


rawlingsj (approved):

rawlingsj:
seems the `Build packages (gdk-pixbuf) ` check is stuck but not been able to find it to cancel / retrigger, wonder if we need to close and re-open the PR?

dlorenc:
I think Jason just fixed that check, rebased.

hectorj2f (approved):

PR #1160, opened 2024-04-23
ensure shbang check only checks valid shbangs

joshrwolf:
the previous implementation was matching on invalid shbangs (`##`) due to the conditional `&&`

it was also erroneously checking and logging errors on directories, so I restructured the caller a little bit as well


xnox (approved):

xnox:
@joshrwolf note to actually fix this; there needs to be a release cut; and wolfictl release cut; and images updated too.

smoser (approved):
thank you.

PR #1184, opened 2024-04-29
sbom: set supplier in addition to originator

imjasonh:
Supplier is required now in ntia-conformance-checker 2.0.0 with https://github.com/spdx/ntia-conformance-checker/pull/176


jspeed-meyers (approved):
LGTM. I apologize for not realizing that this particular change would be an issue. 😢

PR #1187, opened 2024-05-02
melange bump: don't skip main git checkout when mangled package versio…

rawlingsj:
…n is used

There are cases where we use a mangled package version var when doing a git checkout of the main source repo.  Let's update the git commit in these cases.  The check was originally added in https://github.com/chainguard-dev/melange/pull/975 but only identified the main repo using `${{package.version}}`, we should include `${{vars.mangled-package-version}}`


rawlingsj:
fyi @xnox

xnox:
Scope of change:

```
$ echo $(git grep 'tag:.*vars.mangled-package-version' | wc -l)
29
```

PR #1188, opened 2024-05-03
Allow symlinks to provide cmd:

smoser:
Previously code that added automatic 'cmd:' provides ignored symlinks.  That doesn't seem right.

Possible fall out of this is busybox and /bin/sh related things as well as python -> python3 -> python3.12

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

### DIFF
--- a/pkg/sca/sca.go
+++ b/pkg/sca/sca.go
@@ -121,10 +121,6 @@ func generateCmdProviders(ctx context.Context, hdl SCAHandle, generated *config.
 		}
 
 		mode := fi.Mode()
-		if !mode.IsRegular() {
-			return nil
-		}
-
 		if mode.Perm()&0555 == 0555 {
 			if isInDir(path, pathBinDirs) {
 				basename := filepath.Base(path)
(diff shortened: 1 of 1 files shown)


smoser:
the first hunk is Providers.  that will make busybox , dash-binsh and bash-binsh provide 'cmd:sh'. (and google-cloud-sdk provide 'gcloud').

The second hunk is shebang (Depends).  That would make gcloud realize it dependend on sh (depend = cmd:sh) .

We can't do part 2 without part 1 (or gcloud wou'dnt be installable because nothing provides cmd:sh).  We could do part 1 without part 2.

smoser:
@xnox , a request for symlinks as 'cmd:' providers was made https://github.com/chainguard-dev/customer-issues/issues/1416.  The gist of that is that 'google-cloud-sdk' provides /usr/bin/gcloud, but 'apk add cmd:gcloud' will not find anything to install.

I think its hard to disagree that it should.

We generally avoided this previously because of 'python3'  (and python) and 'sh'.

alpine uses provider priority for has a solution for 'sh'. that can be seen [here](https://apk.dag.dev/https/dl-cdn.alpinelinux.org/alpine/edge/main/aarch64/APKINDEX.tar.gz@etag:36363334623538382d3733313331/APKINDEX?full=false&search=binsh&depend=&provide=).

 * yash-binsh - provider prority 50
 * dash-binsh - provider priority 60
 * busybox-binsh - provider priority 100

I have not played with the effects of this on apko or seen what happens if you do 'apk add busybox' and then 'apk add dash-binsh'.

I know you have played here some so would appreciate your thoughts.

jonjohnsonjr (approved):

xnox:
we can try this and see if it sticks.

xnox (approved):
This is great, especially now that https://github.com/chainguard-dev/melange/pull/1164 is fixed too.

PR #1198, opened 2024-05-07
Extract `melange sign` to a library

tcnghia:


k4leung4 (commented):
should we add unit test?

tcnghia:
> should we add unit test?

I don't have enough knowledge of how to do the key verification here to write a meaningful unit test for this. Do you have a suggestion on how I can do so?

k4leung4:
> > should we add unit test?
> 
> I don't have enough knowledge of how to do the signature verification here to write a meaningful unit test for this. Do you have a suggestion on how I can do so?

i have even less of an idea then you
its definitely no worst than before, but having library code that arent tested doesnt seem right.

k4leung4 (dismissed):

tcnghia:
> > > should we add unit test?
> > 
> > 
> > I don't have enough knowledge of how to do the signature verification here to write a meaningful unit test for this. Do you have a suggestion on how I can do so?
> 
> i have even less of an idea then you its definitely no worst than before, but having library code that arent tested doesnt seem right.

@jonjohnsonjr  would RSAVerifySHA1Digest from go-apk work to verify package digest? I could try using that

k4leung4 (approved):

PR #1213, opened 2024-05-16
Only read the first line for shbang.

smoser:
I noticed this when I was playing with a shbang program like:

     #!/usr/bin/env busybox sh
     echo hello world

Melange would give a warning like this:

     Error reading shbang from usr/bin/hello-busybox:
        a shbang of only '/usr/bin/env' with multiple arguments
           (6 /usr/bin/env busybox sh echo "hello world")

### DIFF
--- a/pkg/sca/sca.go
+++ b/pkg/sca/sca.go
@@ -601,7 +601,7 @@ func sonameLibver(soname string) string {
 	return libver
 }
 
-func getShbang(fp fs.File) (string, error) {
+func getShbang(fp io.Reader) (string, error) {
 	// python3 and sh are symlinks and generateCmdProviders currently only considers
 	// regular files. Since nothing will fulfill such a depend, do not generate one.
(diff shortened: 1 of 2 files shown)


imjasonh:
Is there a test you could add to demonstrate the problem, and that this is the fix?

Do you have an idea if there are packages in the wild that will produce different `provides`es with this change? Or is this just about avoiding the misleading warning?

smoser:
> Is there a test you could add to demonstrate the problem, and that this is the fix?
> 
> Do you have an idea if there are packages in the wild that will produce different `provides`es with this change? Or is this just about avoiding the misleading warning?

i could add a test.l  I'm pretty sure this only affects the error path, in which there will be no provides generated.

smoser:
@imjasonh got around to adding a test, and i think fixed another bug. we'll see if it passes the _other_ tests I didn't write (or run) just now.

imjasonh (approved):

PR #1214, opened 2024-05-16
delete k8s runner impl

imjasonh:
We don't plan to use this runner impl in the immediate future, instead using melange bundles via wolfictl. If we decide to pursue this in the future, we have Git history.


luhring (dismissed):
😱

joshrwolf (dismissed):

joshrwolf (approved):

siennathesane:
Well, just saw this after trying to find the k8s runner configs... I was definitely using this but alright.

PR #1222, opened 2024-05-22
sca: allow building packages with version-less command provides

xnox:
Both apk and go-apk allow co-installing packages which have
versionless provides. This is useful for commands which per-se have
identical or unrelated version strings, but provide the same or
similar enough behaviour.

Eg. /usr/bin/pip from py3.10-pip py3.11-pip py3.12-pip of same, or
even different versions, should be co-installable.

Signed-off-by: Dimitri John Ledkov <dimitri.ledkov@chainguard.dev>


xnox (commented):
NACK for now (not needed for now) - we will just split usr/bin/pip into separate conflicting packages.

PR #1230, opened 2024-05-22
Add hidden gcptrace flag

jonjohnsonjr:
This also hoists the trace logic from the build subcommand up into the root Command so that it's available for every subcommand.


luhring (dismissed):

imjasonh (dismissed):

pkg/cli/commands.go
```diff
@@ -44,12 +61,77 @@ func New() *cobra.Command {
...
+			return errors.Join(errs...)
+		},
 	}
 	cmd.PersistentFlags().Var(&level, "log-level", "log level (e.g. debug, info, warn, error)")
 	cmd.PersistentFlags().BoolVar(&gcplog, "gcplog", false, "use GCP logging")
 	_ = cmd.PersistentFlags().MarkHidden("gcplog")
+	cmd.PersistentFlags().BoolVar(&gcptrace, "gcptrace", false, "use GCP tracing")
+	_ = cmd.PersistentFlags().MarkHidden("gcptrace")
+	cmd.PersistentFlags().StringVar(&traceFile, "trace", "", "where to write trace output")
+	_ = cmd.PersistentFlags().MarkHidden("trace")
```
luhring, line 134: If we're marking this as hidden, could we have some documentation somewhere to show that it exists and how to use it? It's an incredible developer feature, IMHO

jonjohnsonjr: I think basically nobody but me uses it, and I'd like to be able to break it in the future.

imjasonh: I'd like to explicitly not commit to supporting this right now, which means (unfortunately) not telling anybody outside of this completely private circle of trust that it exists.

If it proves as useful as we think it is, and word catches on, we should un-hide it and document it.

I think a more ideal outcome would be that all the GCP trace exporting happens in a different, non-melange-CLI caller that wraps melange-as-a-library, since running the melange CLI in GCP directly as a long-term strategy is Bonkers™️.

luhring: I agree with that more ideal outcome. And I agree that we shouldn't overcommit to supporting something externally.

I'm okay with leaving it hidden. I'd like to make a non-blocking request that we at least document this internally, if not now then soon. I'd love to move away from things that only one person gets to know about, even when the thing is not yet considered "stable".

PR #1240, opened 2024-05-30
Speed up presubmit

jonjohnsonjr:
wolfictl scan is very slow to start, and glibc has like a million subpackages, so the presubmit jumped from 15 to 30 minutes when we started wolfictl scanning every package.

image

### DIFF
--- a/.github/workflows/wolfi-presubmit.yaml
+++ b/.github/workflows/wolfi-presubmit.yaml
@@ -106,12 +106,13 @@ jobs:
 
       - name: Test installable and Scan for CVEs
         run: |
-          for f in packages/x86_64/${{ matrix.package }}-*.apk; do
-            docker run --rm -v $(pwd):/work cgr.dev/chainguard/wolfi-base apk add --allow-untrusted /work/$f
+            docker run --rm -v $(pwd):/work --workdir /work cgr.dev/chainguard/wolfi-base apk add --allow-untrusted packages/x86_64/${{ matrix.package }}-*.apk
+
+            # There is a huge fixed cost for every wolfictl scan invocation for grype DB init.
+            # Do this outside of the loop in one invocation with every package.
             wolfictl scan \
             --advisories-repo-dir 'data/wolfi-advisories' \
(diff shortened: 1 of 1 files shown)


.github/workflows/wolfi-presubmit.yaml
```diff
@@ -108,10 +108,13 @@ jobs:
         run: |
           for f in packages/x86_64/${{ matrix.package }}-*.apk; do
             docker run --rm -v $(pwd):/work cgr.dev/chainguard/wolfi-base apk add --allow-untrusted /work/$f
-            wolfictl scan \
```
xnox, line 111: Oh if one can scan all packages in one go, then for loop should be removed altogether no?

xnox (changes requested):

.github/workflows/wolfi-presubmit.yaml
jonjohnsonjr: I don't think they can be _installed_ in one go safely.

jonjohnsonjr: Let's try it.

luhring: @xnox Do we need scanning here in the first place? I'm guessing this was added for a reason, just not sure why yet

jonjohnsonjr: Nice it worked!

xnox (approved):
Whilst we may have conflicting packages in the future in general this should just work for the small number of packages under test and is the quickest way to ensure they all work.

.github/workflows/wolfi-presubmit.yaml
xnox: @luhring yes to ensure that any changes of melange pipelines do not break scanning the binaries. Like when i changed how SBOMs look and what they contain. Because in theory any changes of melange, should still produce scanable / parsable binaries, without any change of CVE detection. As discussed on when SBOM changes with external references got added.

PR #1259, opened 2024-06-06
Add a new property to maven pombump pipeline that defaults to pom.xml and allows an override

krishjainx:
We add a new property that defaults to pom.xml and allow an override so we can call multiple uses: maven/pombump  and pass in the somewhere-else/pom.xml for example. With Maven what we want to do is the same with go/bump where if upstream does update it’s dependencies then the update bot will automatically remove the pom bump entry. There’s some docs here on our internal wiki on Bumping dependencies in Maven (PomBump)  - https://eng.inky.wtf/docs/squads/lifecycle/maintainance/cves/cve-remediation-via-dep-bumps/#bumping-dependencies-in-maven-pombump. Some examples in Wolfi https://github.com/search?q=repo%3Awolfi-dev%2Fos+maven%2Fpombump&type=code. 

As a note:  if maven properties properties are being adjusted, you should use properties-file instead of patch-file when invoking maven/pombump. I'd also recommend renaming the files from pombump-deps-* to pombump-properties-* for consistency. To test locally I did so by passing a flag. This modified pipeline would help avoid creating patches to bump pom.xml dependency versions and make life easier.

```
MELANGE_EXTRA_OPTS="--pipeline-dirs pipelines ./foo" make package/bar
```

and added this updated pombump yaml to the local ./foo folder.

I tested with enterprise-packages/request-1277 v0.1.79 (will use this modified pipeline for updating https://github.com/chainguard-dev/enterprise-packages/pull/4582 to use maven/pombump and get it merged).

it's being used as 

```
  - uses: maven/pombump

  - uses: maven/pombump
    with:
      pom: management-api-server/pom.xml
      properties-file: pombump-properties-server.yaml
 ```

instead of something like

```
 - uses: patch
    with:
      patches: upgrade-deps.patch

```
      
 to patch pom.xml and management-api-server/pom.xml in this instance.
 
 
 cc @rawlingsj


EyeCantCU (approved):
Awesome work. This looks great, and also fixes an issue where a properties file couldn't be used independently of a patch file which is terrific. After this has been merged, we'll need to cut a melange release, update melange in wolfictl, cut a tools release, and then bump image digests in os/ent

PR #1263, opened 2024-06-06
Add a pip install melange build pipeline

dakaneye:
### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

### DIFF
--- a/pkg/build/pipelines/python/pipinstall.yaml
+++ b/pkg/build/pipelines/python/pipinstall.yaml
@@ -1,0 +1,15 @@
+name: Pip Install a Python package
+
+needs:
+  packages:
+    - busybox
+
+pipeline:
+  - runs: |
+      if ! [ -x "$(command -v python)" ]; then
+        echo 'Error: Python is not installed.'
+        exit 1
+      fi
+
+  - runs: |
+      pip install --prefix=/usr --root="${{targets.contextdir}}" .


pkg/build/pipelines/python/pipinstall.yaml
```diff
@@ -0,0 +1,15 @@
+name: Pip Install a Python package
+
+needs:
+  packages:
+    - busybox
```
vaikas, line 5: Should this also require `pip` as a dependency? Like here:
https://github.com/chainguard-dev/melange/blob/main/pkg/build/pipelines/go/build.yaml#L5

Now, having said that, I wonder how this should be handled with the python versioning. I know @smoser has been working on that, so tagging him here.

dakaneye: I thought about this, but there are 3 different pip packages, perhaps we add that as an input, defaulting to the most recent one?

xnox:
@dakaneye we have this in-flight https://github.com/wolfi-dev/os/pull/20318 =)

pkg/build/pipelines/python/pipinstall.yaml
smoser: I'm guessing that @dakaneye found https://github.com/wolfi-dev/os/pull/20318/files after doing this.  (thanks for being thorough).

I think right now I'd not add this pipeline.  We may or may not end up adding one in the future, but our use of pip and python package installation is going through some changes.

smoser (changes requested):

PR #1274, opened 2024-06-13
Remove ability to set logging policy 

krishjainx:
cc @imjasonh 

(will lint in time)

### DIFF
--- a/docs/md/melange.md
+++ b/docs/md/melange.md
@@ -14,9 +14,8 @@ toc: true
 ### Options
 
 ```
-  -h, --help                 help for melange
-      --log-level string     log level (e.g. debug, info, warn, error) (default "info")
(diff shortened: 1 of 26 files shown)


imjasonh:
So, this diff is actually instructive about why this wasn't doing what we intended. The purpose of `--log-policy` was to pass those values to [apko's log writer](https://github.com/chainguard-dev/apko/blob/5cc150ddfcabe9fbe7ee700d731ff55721a27989/pkg/log/log.go#L28) which understands them. It looks like removing the value of `--log-writer` shows it was never used by anything.

FWIW I'm still fine moving forward with this PR and remove this from Melange. But, another option, would be to figure out where we were supposed to be passing these to apko's `log.Writer` and fixing it.

If you feel like you can find the right spot for it, that's great. If not, I'm also fine removing this "feature" from Melange (and Apko in the future)

imjasonh (approved):

PR #1277, opened 2024-06-13
git-checkout - try harder if getting hash from tag fails.

### LINKED ISSUE #1272: git-checkout pipeline fails with some repos

smoser:
If the remote repository has a branch and a tag with the provided name (in there exists a branch in the remot with the same name as the provided tag (inputs.tag), then git rev-parse will fail with:

    fatal: Needed a single revision

If that happens, then try to get a value with:

    git ls-remote --tags origin ${tag}

Fixes #1272.

### DIFF
--- a/pkg/build/pipelines/git-checkout.yaml
+++ b/pkg/build/pipelines/git-checkout.yaml
@@ -39,65 +39,110 @@ inputs:
 
 pipeline:
   - runs: |
-      if [ -z "${{inputs.branch}}" ] && [ -z "${{inputs.tag}}" ]; then
-        echo "Warning (git-checkout): you have not specified a branch or tag."
-      fi
+      #!/bin/sh
+      set -e
 
-      git_clone_flags=""
-      if [ "${{inputs.recurse-submodules}}" == "true" ]; then
-        git_clone_flags="--recurse-submodules"
-      fi
+      msg() { echo "[git checkout]" "$@" 1>&2; }
+      fail() { msg FAIL "$@"; exit 1; }
+      vr() { msg "execute:" "$@"; "$@"; }
 
-      [ -n '${{inputs.branch}}' ] && clone_target='--branch ${{inputs.branch}}'
-      [ -n '${{inputs.tag}}' ] && clone_target='--branch ${{inputs.tag}}'
+      main() {
+          local repo=$1 dest=${2:-.} depth=${3:-"-1"} branch=$4
+          local tag=$5 expcommit=$6 recurse=${7:-false}
+          msg "repo='$repo' dest='$dest' depth='$depth' branch='$branch'" \
+              "tag='$tag' expcommit='$expcommit' recurse='$recurse'"
 
-      workdir=$(mktemp -d)
-      mkdir -p '${{inputs.destination}}'
-      clone_fullpath=$(realpath '${{inputs.destination}}')
+          case "$recurse" in
+              true|false) :;;
+              *) fail "recurse must be true or false, not '$recurse'"
+          esac
 
-      git config --global --add safe.directory $workdir
-      git config --global --add safe.directory $clone_fullpath
-      if [ "${{inputs.depth}}" == "-1" ]; then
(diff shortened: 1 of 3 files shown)


smoser:
this doesn't really work.  when you clone with --branch=tag and there is a branch named 'tag', it will clone the branch. :-(

smoser:
I'm not sure I have the "unpeel" part correct. Anyone have an example on that ?

Apologies for the full rewrite.  Its a lot easier to test this this way as you can just put the script into a file and then execute it with arguments.

pkg/build/pipelines/git-checkout.yaml
```diff
@@ -39,65 +39,108 @@ inputs:
...
-        if [[ '${{inputs.expected-commit}}' != "${unpeeled_commit}" ]]; then
-          echo "Error (git-checkout): expect commit ${{inputs.expected-commit}}, got ${unpeeled_commit}"
-          exit 1
-        fi
-      else
-        echo "Error (git-checkout): no branch or tag provided"
-      fi
+      #!/bin/bash
+      set -e
+      set -o pipefail
```
smoser, line 44: I just realized `set -o pipefail` and `#!/bin/bash`. I think the only bashism here is `pipefail`.

The previous version of the script had several bashism  (`[[`) but did not declare bash anywhere).
That is interesting.  are pipelines allowed to assume bash?

pkg/build/pipelines/git-checkout.yaml
```diff
@@ -39,65 +39,108 @@ inputs:
...
+              [ "$tagobj" != "$expcommit" ] &&
+                  msg "tag object hash was $tagobj"
+              fail "Expected commit $expcommit for $tag, found $foundcommit"
+          fi
+
+          msg "Warning: The provided expected-commit ($expcommit)"
+          msg "was the hash of the annotated tag object for $tag."
+          msg "Update to set expected-commit to $foundcommit"
+          return 0
+
```
jonjohnsonjr, line 139:
```suggestion

          return 0
```

pkg/build/pipelines/git-checkout.yaml
smoser: Jon pointed me at https://github.com/chainguard-dev/melange/blob/ee5792a7313b7d062c3cf5960fbeaaf90350940c/pkg/build/pipeline.go#L325-L332 .

So the answer to my question is 'no', they cannot assume bash.
The interesting fact is though, that in almost all scenarioes (/bin/sh is busybox, and busybox's /bin/sh implements both `[[` and `set -o pipefail`).  So we're lucky that the previous version works, if it was ever run with dash it would not.

I went ahead and pushed a commit to drop pipefail and still check the rc value of the left side of the `tar | tar`

pkg/build/pipelines/git-checkout.yaml
smoser: I think i did what you were asking.

smoser:
fwiw, I did run a test of  several different wolfi packages. I looked through the list of 'git-checkout' and found different configured stanzas.  here is the shell test i ran:

```bash
#!/bin/bash
set -o pipefail
build() {
    local name="$1" reason="$2"
    echo "$(date -R)" "building $name : ${reason}"
    ( echo "$(date -R): start" make "package/$name" 
      echo "make package/$name"
      make "package/$name"
      rc=$?
      echo "$(date -R): finished with exit $rc in ${SECONDS}s"
      exit $rc
    ) 2>&1 | tee ~/"logs/${name}.log"
}
fails=0
while read pname reason ; do
    build "$pname" "$reason" || fails=$((fails+1))
done <<EOF
acme.sh                standard
aom                    destination
apache-arrow           working-directory
argon2                 explicit depth=1
jvmkill                branch
py3-archspec           recurse-submodules
ruby3.2-yajl-ruby      branch no expected
EOF

echo "finished in ${SECONDS}s with $fails fails"
exit $fails
```

imjasonh (approved):

PR #1288, opened 2024-06-14
unbreak build at head from log policy removal

k4leung4:
```
# chainguard.dev/melange/pkg/cli
Error: pkg/cli/compile.go:108:11: undefined: build.WithLogPolicy
```


smoser (approved):
I wonder how this landed?

Issue #1299, opened 2024-06-17
Make  python/import pipeline support multiple pipeline args


pnasrat:
Currently pkg/build/pipelines/python/import.yaml uses python3, this should be configurable

smoser:
PR https://github.com/chainguard-dev/melange/pull/1332/files

PR #1300, opened 2024-06-17
python/import pipeline allow setting python binary

pnasrat:
See: #1299
Signed-off-by: Pris Nasrat <pris.nasrat@chainguard.dev>

### Functional Changes

Notes:

Will build and run against python packages using this pipeline before
requesting review

### DIFF
--- a/pkg/build/pipelines/python/import.yaml
+++ b/pkg/build/pipelines/python/import.yaml
@@ -5,6 +5,10 @@ needs:
     - wolfi-base
 
 inputs:
+  python:
+    description: Which python to use
+    required: true
+    default: python3
   import:
     description: |
       The package to import.
@@ -17,7 +21,7 @@ inputs:
 pipeline:
   - runs: |
       if [ -n "${{inputs.from}}" ]; then
-        python3 -c "from ${{inputs.from}} import ${{inputs.import}}"
+        ${{inputs.python}} -c "from ${{inputs.from}} import ${{inputs.import}}"
       else
-        python3 -c "import ${{inputs.import}}"
+        ${{inputs.python}} -c "import ${{inputs.import}}"
       fi


smoser (commented):
I verified this works with one of the new packages (setup tools) like this:

```diff
diff --git a/py3-setuptools.yaml b/py3-setuptools.yaml
index 86881e803..b8d41506a 100644
--- a/py3-setuptools.yaml
+++ b/py3-setuptools.yaml
@@ -47,6 +47,12 @@ subpackages:
         with:
           python: python${{range.key}}
       - uses: strip
+    test:
+      pipeline:
+        - uses: python/import
+          with:
+            import: setuptools
+            python: python${{range.key}}
 
   - name: py3-supported-${{vars.pypi-package}}
     description: meta package providing ${{vars.pypi-package}} for supported python versions.
```

We can make it find the right python in more cases like this:
```diff
diff --git a/pkg/build/pipelines/python/import.yaml b/pkg/build/pipelines/python/import.yaml
index ee3a40f..9850e02 100644
--- a/pkg/build/pipelines/python/import.yaml
+++ b/pkg/build/pipelines/python/import.yaml
@@ -8,7 +8,7 @@ inputs:
   python:
     description: Which python to use
     required: true
-    default: python3
+    default: DEFAULT
   import:
     description: |
       The package to import.
@@ -20,8 +20,29 @@ inputs:
 
 pipeline:
   - runs: |
+      python="${{inputs.python}}"
+      if [ "$python" = "DEFAULT" ]; then
+        set +x
+        nfound=0
+        for python in /usr/bin/python3.*; do
+          [ -x "$python" ] && nfound=$((nfound+1))
+        done
+        if [ $nfound -eq 0 ]; then
+          if [ ! -x /usr/bin/python3 ]; then
+            echo "no python3.X found and no python3 - specify 'python'"
+            exit 1
+          fi
+          python=/usr/bin/python3
+        elif [ $nfound -gt 1 ]; then
+          echo "found $nfound python3.X programs - must specify 'python'"
+          exit 1
+        fi
+        echo "using python $python"
+        set -x
+      fi
+
       if [ -n "${{inputs.from}}" ]; then
-        ${{inputs.python}} -c "from ${{inputs.from}} import ${{inputs.import}}"
+        $python -c "from ${{inputs.from}} import ${{inputs.import}}"
       else
-        ${{inputs.python}} -c "import ${{inputs.import}}"
+        $python -c "import ${{inputs.import}}"
       fi
```

The simplicity of the original is nice, but removing the need for the caller to specify 'python' has some benefits, and potentially makes existing packages work without change.

I'm just putting that here for consideration, I'm actually OK with the existing PR.

pnasrat:
@smoser I think I would prefer on doing the simple explicit thing in my PR for now, given setting the python version only needed for versioned python packages for now.

smoser:
> @smoser I think I would prefer on doing the simple explicit thing in my PR for now, given setting the python version only needed for versioned python packages for now.

ACK.

Issue #1306, opened 2024-06-18
Python test pipelines should test every python version


xnox:
See https://github.com/wolfi-dev/os/pull/20318/files#r1643611760

We should add import tests where possible.

pnasrat:
See #1299 and my draft PR #1300

pnasrat:
We probably want subpackage test block in the range for import tests for the new style py3 version packaging, then setting the python binary to use rather than the main package test

pnasrat:
Also follow up to change the numpy e2e test and remove the env changes in #1308

PR #1308, opened 2024-06-18
melange numpy test include python-3.12

pnasrat:
Needed for default python until multiversion python supported see #1306

### DIFF
--- a/e2e-tests/numpy-test.yaml
+++ b/e2e-tests/numpy-test.yaml
@@ -8,11 +8,16 @@ pipeline:
 
 # Test python/import, and python/test pipelines
 test:
+  environment:
+    # TODO(pnasrat): fix to use multiple python
+    contents:
+      packages:
+        - python-3.12
   pipeline:
     # Test import with command (python -c "import numpy")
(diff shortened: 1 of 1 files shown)


imjasonh (approved):

e2e-tests/numpy-test.yaml
```diff
@@ -8,11 +8,16 @@ pipeline:
 
 # Test python/import, and python/test pipelines
 test:
+  environment:
+    # TODO(pnasrat): fix to use multiple python
+    contents:
+      packages:
+        - python-3.12
```
smoser, line 15: You should not need this, and it might actually just complicate debugging of a future error case.

Its nice to be explicit, but I think this will cause issue in teh future when numpy moves to 3.13 (default version) and then python-3.12 will be installed along side of python-3.13-base, the 'import' below will just give an error like

```
$ python3.12 -c 'import numpy'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'numpy'
```

PR #1314, opened 2024-06-21
Fix/e2e tests

smoser:
Itemized commits here, so feel free to read them.
Summary of things here:
1. make all yaml files in e2e-tests get built with melange via make target 'test-e2e'
2. fix a problem in the git-checkout-test that was there (depth was wrong)
3. cleanup git-checkout pipeline and test
4. Run test-e2e from single pipeline without matrix (locally 'test-e2e' runs the 6 tests to completion in 15 seconds total.


imjasonh (dismissed):

pkg/build/pipelines/git-checkout.yaml
```diff
@@ -67,9 +67,10 @@ pipeline:
               msg "Warning: no expected-commit"
 
           local flags="" depthflag="" dest_fullpath="" workdir=""
-          local remote="origin" rcfile="" rc=""
-          [ -n "$branch" ] && flags="--branch=$branch"
-          [ -n "$tag" ] && flags="--branch=$tag"
+          local remote="origin" rcfile="" rc="" quiet="--quiet"
+          flags="--config=advice.detachedHead=false"
+          [ -n "$branch" ] && flags="$flags --branch=$branch"
+          [ -n "$tag" ] && flags="$flags --branch=$tag"
```
xnox, line 73: Back in the day, prior to smoser/xnox changes here, branch and tag were mutually exclusive. With this change, when both are specified (which is rare) the clone flags will be `--branch=$branch --branch=$tag`, i don't know if it is expected and intentional, but `--branch=$tag` will win.

Separately we have had bugs with repositories which contain branches and tags of identical name. Is there something better that we can do here? i.e. can we do `--branch=refs/heads/$branch` and `--branch=refs/tags/$tag` ? Such that nothing is ever ambigious?

smoser: > Back in the day, prior to smoser/xnox changes here, branch and tag were mutually exclusive.

I thought they were.  And I actually thought that I had written something to error if they were both given.
But now
a. don't see what I thought I wrote to stop that.
b. I don't see where it did that before.  https://github.com/chainguard-dev/melange/commit/8b8c714c3af15d8df48942b8c425894181ad9a62 is the last commit before I joined. Full file [here](https://github.com/chainguard-dev/melange/blob/8b8c714c3af15d8df48942b8c425894181ad9a62/pkg/build/pipelines/git-checkout.yaml) 

It does look to me like "`--branch=$branch`" would win before. 

I'm up for making it fail if they're both given. thoughts?

smoser: here is what I would do:
```diff
diff --git a/pkg/build/pipelines/git-checkout.yaml b/pkg/build/pipelines/git-checkout.yaml
index 8153651..47d35df 100644
--- a/pkg/build/pipelines/git-checkout.yaml
+++ b/pkg/build/pipelines/git-checkout.yaml
@@ -61,6 +61,8 @@ pipeline:
 
           if [ -z "$branch" ] && [ -z "$tag" ]; then
               msg "Warning: you have not specified a branch or tag."
+          elif [ -n "$branch" ] && [ -n "$tag" ]; then
+              fail "both branch ($branch) and tag ($tag) are specified."
           fi
 
           [ -n "$expcommit" ] ||
```

smoser: I just looked at wolfi and the only occurences of ' tag:' and 'branch:' are:
```
$ git grep -l branch: | xargs grep -l " tag:"
aws-for-fluent-bit.yaml
cilium-1.15.yaml
```

They don't use them in the same git checkout stanza.

xnox (approved):

xnox:
@jonjohnsonjr force merge is required, as github pipelines are being changed; also which pipelines are now required need update too.

PR #1315, opened 2024-06-24
lint: support linting existence of info dirs

imjasonh:
Also:
- support linting remote APKs
- remove some unnecessary regexps, just use `strings.HasPrefix` instead
- warn on lint findings, remove unnecessary newlines

There's more lint cleanups I'd love to do, and put linting on track to be a blocking check. That work never ended up happening before. 😕


pkg/cli/lint.go
```diff
@@ -89,29 +90,26 @@ func Lint() *cobra.Command {
...
 	g, ctx := errgroup.WithContext(ctx)
+	g.SetLimit(runtime.GOMAXPROCS(0))
 
 	for _, pkg := range pkgs {
 		p := pkg
-
-		g.Go(func() error {
-			return o.run(ctx, p)
-		})
+		g.Go(func() error { return o.run(ctx, p) })
```
jdolitsky, line 97: oh hell yea

jdolitsky (dismissed):

PR #1322, opened 2024-06-24
rewrite linting

imjasonh:
Instead of `--fail-on-lint-warning` this adds flags for `--lint-require` and `--lint-warn`, which configures lint checks to warn and fail on (currently the default is only to require the new `infodir` linter)

Over time we can add more required checks, starting with them as warnings while we see how noisy that is.

Build configs can `disable` checks (they can no longer `enable` them), in which case a required check becomes a warning instead.

Linting already-built packages has no way to disable a check, since all we have is the package contents, and not the build context.


main.go
```diff
@@ -15,19 +15,13 @@
...
 
 	"chainguard.dev/melange/pkg/cli"
 )
 
 func main() {
-	ctx, done := signal.NotifyContext(context.Background(), os.Interrupt)
-	defer done()
-
-	if err := cli.New().ExecuteContext(ctx); err != nil {
+	if err := cli.New().Execute(); err != nil {
```
jonjohnsonjr, line 24: Why drop this?

pkg/cli/build.go
```diff
@@ -211,11 +213,12 @@ func Build() *cobra.Command {
 	cmd.Flags().BoolVar(&debugRunner, "debug-runner", false, "when enabled, the builder pod will persist after the build succeeds or fails")
 	cmd.Flags().BoolVarP(&interactive, "interactive", "i", false, "when enabled, attaches stdin with a tty to the pod on failure")
 	cmd.Flags().BoolVar(&remove, "rm", false, "clean up intermediate artifacts (e.g. container images)")
-	cmd.Flags().BoolVar(&failOnLintWarning, "fail-on-lint-warning", false, "turns linter warnings into failures")
```
jonjohnsonjr, line 214: Let's hide this flag with at TODO to drop it (and maybe log a warning if it's true) so we can update `melange` without having to fix the flags up.

pkg/cli/lint.go
```diff
@@ -47,69 +41,32 @@ func Lint() *cobra.Command {
...
-			// Enable non-default lints
-			for _, e := range enabled {
-				if !slices.Contains(disabled, e) {
-					linterSet[e] = struct{}{}
-				}
+			g, ctx := errgroup.WithContext(ctx)
+			g.SetLimit(runtime.GOMAXPROCS(0))
+
+			errs := []error{}
+			var mu sync.Mutex
```
jonjohnsonjr, line 48: I have a slight preference for avoiding the mutex by doing `make([]error, len(args))` and having each goroutine set `errs[i] = err` then letting `errors.Join` filter out the nils.

pkg/cli/lint.go
```diff
@@ -47,69 +41,32 @@ func Lint() *cobra.Command {
...
-
-			o.linters = linters
-
-			return o.RunAllE(ctx, args...)
+			return errors.Join(errs...)
 		},
 	}
 
-	cmd.Flags().StringSliceVar(&enabled, "enabled", []string{}, "enable additional, non-default lints, `--disabled` overrides this")
-	cmd.Flags().StringSliceVar(&disabled, "disabled", []string{}, "disable linters enabled by default or passed in `--enabled`")
```
jonjohnsonjr, line 86: Same comment re: hiding these and logging a warning.

pkg/linter/linter.go
```diff
@@ -512,105 +492,39 @@ func pythonTestPostLinter(_ LinterContext, fsys fs.FS) error {
...
-		if !present {
-			linters = append(linters, l)
-		}
-	}
-
-	return linters
-}
-
-func (lctx LinterContext) lintPackageFs(warn func(error), linters []string, linterClass linter_defaults.LinterClass) error {
+func lintPackageFs(pkgname string, fsys fs.FS, linters []string) error {
```
jonjohnsonjr, line 495: `lintPackageFS` I think is more idiomatic?

pkg/linter/linter.go
```diff
@@ -248,126 +230,126 @@ func srvLinter(_ LinterContext, path string, _ fs.DirEntry) error {
...
+		}
 
-	// No debug sections allowed
-	if file.Section(".debug") != nil || file.Section(".zdebug") != nil {
-		return fmt.Errorf("ELF file is not stripped")
-	}
+		file, err := elf.NewFile(readerAt)
+		if err != nil {
+			// We don't particularly care if this fails otherwise.
+			fmt.Printf("WARNING: Could not open file %q as executable: %v\n", path, err)
```
jonjohnsonjr, line 338: If this is easy enough to disable, I'd rather return an error here than pollute stdout. I think this is a holdover from when we didn't have the check right above that skips non-elf-magic.

pkg/linter/linter.go
```diff
@@ -512,105 +492,39 @@ func pythonTestPostLinter(_ LinterContext, fsys fs.FS) error {
...
-	return lctx.lintPackageFs(warn, linters, linter_defaults.LinterClassBuild)
+	if err := lintPackageFs(packageName, fsys, warn); err != nil {
+		log.Warnf("package linter warning in %s: %v", packageName, err)
+	}
+	return lintPackageFs(packageName, fsys, require)
 }
 
 // Lint the given APK at the given path
-func LintApk(ctx context.Context, path string, warn func(error), linters []string) error {
+func LintApk(ctx context.Context, path string, require, warn []string) error {
```
jonjohnsonjr, line 527: `LintAPK` is idiomatic.

pkg/linter/linter.go
```diff
@@ -512,105 +492,39 @@ func pythonTestPostLinter(_ LinterContext, fsys fs.FS) error {
...
-		}
-	}
-
-	return nil
+	return errors.Join(errs...)
 }
 
 // Lint the given build directory at the given path
-func LintBuild(packageName string, path string, warn func(error), linters []string) error {
+func LintBuild(ctx context.Context, packageName string, path string, require, warn []string) error {
```
jonjohnsonjr, line 516: WDYT about moving the warn vs require distinction outside of this function?

Then we'd just call it twice like:

```go
if err := linter.LintBuild(ctx, pkg, path, warns); err != nil {
    log.Warnf("package linter warning in %s: %v", packageName, err)
}
if err := linter.LintBuild(ctx, pkg, path, requires); err != nil {
  return err
}
```

Or is it better for the require vs warn handling to be the responsibility of this function/package?

Genuinely asking, I haven't thought too deeply about this.

pkg/linter/linter.go
```diff
@@ -18,153 +18,136 @@ import (
...
 	"net/http"
 	"os"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"strings"
 
 	"chainguard.dev/apko/pkg/apk/expandapk"
-	linter_defaults "chainguard.dev/melange/pkg/linter/defaults"
+	"github.com/charmbracelet/log"
```
jonjohnsonjr, line 33: I think we're accidentally pulling this in, see my suggestion below about moving the warn vs require distinction outside of this file.

main.go
imjasonh: Doing notify handling ourselves broke ctrl+C, just relying on cobra's makes it work again.

pkg/cli/build.go
imjasonh: Smart smart.

pkg/linter/linter.go
imjasonh: I don't really love this either. I don't feel strongly, it's only done twice.

pkg/cli/lint.go
imjasonh: Ehhh, that feels clever for a very very small performance gain. If you find out it's slowing us down we can do your thing.

jonjohnsonjr (approved):

PR #1328, opened 2024-06-26
debug: Populate history file via mounts

jonjohnsonjr:
My initial implementation of this was very naive and just used single-quote strings for history, which is almost certainly wrong most of the time.

This just writes the file to the workspace directory, which is less error prone.


smoser (commented):
This works for me . thank you.

The history isn't terribly useful for `git-checkout` , (commands separated over multiple lines, it writes its own HEREDOC ... ).

but, this fixes what was broken.

smoser (approved):

PR #1333, opened 2024-06-28
fix(cargo/build): test for non-zero length

jalmeroth:
Hi :wave:,

I tried to use the `cargo/build` pipeline for the first time and I noticed that the pipeline seems to test the wrong condition. This PR fixes this behaviour.

Greets,
Jan


EyeCantCU (approved):
Thank you!

PR #1343, opened 2024-07-03
support var-transforms in subpackage names

imjasonh:


pkg/config/config.go
```diff
@@ -940,6 +940,10 @@ func ParseConfiguration(ctx context.Context, configurationFilePath string, opts
 
 	// Mutate config properties with substitutions.
 	configMap := buildConfigMap(&cfg)
+	if err := cfg.PerformVarSubstitutions(configMap); err != nil {
+		return nil, fmt.Errorf("applying variable substitutions: %w", err)
+	}
```
luhring, line 945: Should we update the tests to make sure this is happening correctly?

imjasonh: Oh sure!

luhring (approved):
🎉

PR #1360, opened 2024-07-09
python/pipelines - resolve symlink to full path.

smoser:
Running 'python3 -m installer my.whl' (as gets done in build-wheel) will write shebang lines with 'python3'.

The melange SCA is ignoring symlinks in shebang at the moment, so it will not create any dependencies for that.

The larger problem here is that the python package that installs /usr/bin/foo with a shebang '#!/usr/bin/python3' will get a dependency in the SCA on python3.12-base (due to the site-packages python files), but the 'python3' may end up being python3.11. That situation would end up with broken /usr/bin/foo.

I chose to take the least invasive path here, and only change behavior when the 'python3' in the PATH is a symlink.

I *also* left the build and install pipelines calling 'python' rather than 'python3'.  I do not like that, but did not want to change it here.

### DIFF
--- a/pkg/build/pipelines/python/build-wheel.yaml
+++ b/pkg/build/pipelines/python/build-wheel.yaml
@@ -5,6 +5,7 @@ needs:
     - busybox
     - py3-build
     - py3-installer
+    - python3
 
 pipeline:
   - runs: |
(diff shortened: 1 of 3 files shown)


dakaneye (approved):

smoser:
https://github.com/wolfi-dev/os/pull/23627 uses/needs this.

Issue #1359, opened 2024-07-09
`PATH` is not set in `build --debug`


smoser:
I've bumped into a lot of different times when running `melange build --debug` but never really understood it.
Now I understand whats going on.  PATH is not actually set in the environment, it gets set by the shell builtin, but not exported.

Here is an example.

```yaml
$ cat test-me.yaml 
package:
  name: test-me
  version: 1.0
  epoch: 0

environment:
  contents:
    packages:
      - busybox
      - python3

pipeline:
  - runs: |
      echo this is building something
      cat >my.py <<"EOF"
      import os, sys
      print("inside %s: PATH=%s" % (sys.executable, os.environ["PATH"]))
      EOF
      python3 my.py
      exit 1
```

```bash
$ melange build --debug test-me.yaml \
   --repository-append=https://packages.wolfi.dev/os \
   --keyring-append=https://packages.wolfi.dev/os/wolfi-signing.rsa.pub \
   --arch=x86_64 --debug --interactive
2024/07/09 16:03:19 INFO ImgRef = /tmp/melange-guest-350654200
2024/07/09 16:03:19 WARN + export 'PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin'
2024/07/09 16:03:19 WARN + export 'HISTFILE=~/.ash_history'
2024/07/09 16:03:19 WARN + '[' -d /home/build ]
2024/07/09 16:03:19 WARN + cd /home/build
2024/07/09 16:03:19 WARN + echo this is building something
2024/07/09 16:03:19 WARN + cat
2024/07/09 16:03:19 INFO this is building something
2024/07/09 16:03:19 WARN + python3 my.py
2024/07/09 16:03:19 INFO inside /usr/bin/python3: PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
2024/07/09 16:03:19 WARN + exit 1
...
## above is output, here is interactive shell, run the same 'my.py'
$ python3 ./my.py 
Traceback (most recent call last):
  File "/home/build/./my.py", line 2, in <module>
    print("inside %s: PATH=%s" % (sys.executable, os.environ["PATH"]))
                                                  ~~~~~~~~~~^^^^^^^^
  File "<frozen os>", line 714, in __getitem__
KeyError: 'PATH'

```

```
$ melange version
  __  __   _____   _          _      _   _    ____   _____
[...]

PR #1362, opened 2024-07-09
Fix env overrides for interactive builds

jonjohnsonjr:
Fixes https://github.com/chainguard-dev/melange/issues/1359

For reasons I don't understand, the bubblewrap runner would not have $PATH set in interactive builds. The way we were setting PATH was to prepend it to the shell script we execute for each step, which is kind of janky. I've shifted things around to make that overriding of the environment happen in go via arguments rather than at runtime.

This does break the ability for pipeline-specified environment variables to be evaluated at runtime, but that is actually terrifying and I don't think we want that.

### DIFF
--- a/pkg/build/pipeline.go
+++ b/pkg/build/pipeline.go
@@ -21,6 +21,7 @@ import (
 	"maps"
 	"os"
 	"os/signal"
+	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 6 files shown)


smoser (approved):
This is somewhat the path I started down, and then I said to myself "Jon can do this faster". Thanks!

Less janky shell is better.

PR #1386, opened 2024-07-19
Feat/qemu runners

89luca89:
This PR introduces the ability to use `qemu` as runner for melange.

The (micro) VM is build starting from the apko rootfs, with the following package injected in:

- linux (the kernel)
- net-utils
- mount
- ssh-{server,server-config}

# How it works

The apko layer (melange-guest-xxxxx) is used as a base to build the `initramfs` for the VM.
We don't need an *actual* disk, we just start the VM with an SSH server.

during the VM creation, an ephemeral keypair for SSH is created, as commands will be issued this way to the VM.

Every additional mountpoint is passed using 9pfs + fstab, so they appear transparently inside the VM.

Exception is done for the workspace, as it's fetched by tar+scp of `/home/build/melange-out`

# Results

Performance hit was high when using 9pfs to mount the workspace, but using the tar+scp method yields much better results.

For now the biggest time loss is in the creation of the initramfs via cpio+gzip (can take up to 20->30s)

Screenshot from 2024-07-19 11-04-35

As shown in the screenshot, the compilation of the linux kernel takes about 1m hit, but about ~30s of those are the initramfs creation.

Also as shown we have a *big* impact on IO using 9pfs, so we should avoid it when possible.

## TODO/Future improvements

- [x] support arm builds
- [x] macos support 
- [x] default to a precompiled kernel for the VM if the host does not supply one
- [x] improve cpio performance
- [x] rootless builds inside the VM (not sure if necessary due to the sandboxing nature)
- [x] replace systemd with a minimal busybox init

### Functional Changes


pkg/container/qemu_runner.go
```diff
@@ -0,0 +1,820 @@
...
+	}
+
+	clog.FromContext(ctx).Info("qemu: kernel not found, downloading...")
+	// download mainline kernel from ubuntu
+	response, err := http.Get("https://dl-cdn.alpinelinux.org/alpine/edge/main/x86_64/linux-virt-6.6.41-r0.apk")
+	if err != nil {
+		clog.FromContext(ctx).Errorf("qemu: can't download kernel - %v", err)
+		return "", err
+	}
+	defer response.Body.Close()
```
mattmoor, line 557: Instead of hardcoding this (which will break when Alpine withdraws this kernel, and won't work on aarch64) perhaps we can allow folks to set an environment variable for where to pick up the kernel (if set)?

89luca89: Makes sense yes, implemented it so that one can pass a file path to a prebuild vmlinuz or an APK url
else it will use a sensible default

pkg/container/qemu_runner.go
```diff
@@ -0,0 +1,879 @@
...
+	server := host + ":" + port
+
+	signer, err := ssh.ParsePrivateKey(cfg.SSHKey)
+	if err != nil {
+		clog.FromContext(ctx).Errorf("Unable to parse private key: %v", err)
+		return err
+	}
+
+	// Create SSH client configuration
+	config := &ssh.ClientConfig{
```
smoser, line 748: We could specify the cipher and potentially increase the ssh performance. [this](https://www.linkedin.com/pulse/bench-marking-ssh-ciphers-transfer-speed-phil-cryer-eakxc/) says that aes128-ctr is the fastest of those enabled, and reports it is better than 2x the performance of the default (chacha20).

pkg/container/qemu_runner.go
```diff
@@ -0,0 +1,879 @@
...
+	err := sendSSHCommand(ctx,
+		user,
+		"localhost",
+		cfg.SSHPort,
+		cfg,
+		nil,
+		nil,
+		nil,
+		nil,
+		[]string{"cd /home/build && tar cvzf melange-out.tar.gz melange-out"},
```
smoser, line 223: It feels like we can probably read and write without the temporary file.
ie, you could just execute `tar -C /home/build -czf - melange-out`
and then dump that into the local 'melange-out.tar.gz' file.

I'd think somehow we could even create the workspaceTar with the ssh connection's reader.

pkg/container/qemu_runner.go
```diff
@@ -0,0 +1,879 @@
...
+		nil,
+		nil,
+		workspaceTar,
+		[]string{"cat /home/build/melange-out.tar.gz"},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return os.Open(filepath.Join(cfg.WorkspaceDir, "melange-out.tar.gz"))
```
smoser, line 254: does the output have to be compressed?

it feels like compression and then decompression is going to be slower than skipping it on a virtio network.

pkg/container/qemu_runner.go
```diff
@@ -0,0 +1,879 @@
...
+	// but still useful otherwise
+	baseargs = append(baseargs, "-display", "none")
+	baseargs = append(baseargs, "-no-reboot")
+	baseargs = append(baseargs, "-no-user-config")
+	baseargs = append(baseargs, "-nodefaults")
+	baseargs = append(baseargs, "-parallel", "none")
+	baseargs = append(baseargs, "-serial", "none")
+	baseargs = append(baseargs, "-vga", "none")
+	// use -netdev + -device instead of -nic, as this is better supported by microvm machine type
+	baseargs = append(baseargs, "-netdev", "user,id=id1,hostfwd=tcp::"+cfg.SSHPort+"-:22")
```
smoser, line 369: https://pkg.go.dev/github.com/mdlayher/vsock would maybe allow us to use a vsock for host<->guest . I guess you still need user networking for host -> internet.

smoser (commented):
ok, i left a bunch of comments. 
I probably should have first said "This is neat!".  I'm really surprised how little there is to this.  Great work.

pkg/container/qemu_runner.go
89luca89: Thanks @smoser addressed this :+1:

pkg/container/qemu_runner.go
89luca89: Not having proper scp and rsync I thought using a single connection to move a single file would be better
as throughput would be better than latencies in these type of connections (localhost)

pkg/container/qemu_runner.go
89luca89: That's something to explore thanks
This would remove that ~5s delay of the `sshd` starting up and would help a bit for smaller packages in proportion

89luca89:
@mattmoor @dustinkirkland @imjasonh 

Moved the init to the following subpackage: https://github.com/wolfi-dev/os/pull/24903
This frees melange from fiddling with the rootfs, and rely on apko for the cpio build.

Also now, the ssh key is shared using a network mount (9p) and set up in the init itself.

Some fiddling of the rootfs is needed in case we need to specify kernel modules:
`QEMU_KERNEL_IMAGE=/path/to/vmlinuz` points to a `vmlinuz` file, but most of them (if not compiled with `make kvm_guest`) will miss the virtio and 9p drivers by default
in those case we specify the `QEMU_KERNEL_MODULES=/path/to/lib/modules/` and we inject that kernel's modules into the apko's generated v1.Layer, then use apko's new cpio build on that.

Also introduced some `ssh` improvements suggested by @smoser  :+1:

89luca89:
> @mattmoor @dustinkirkland @imjasonh
> 
> Moved the init to the following subpackage: [wolfi-dev/os#24903](https://github.com/wolfi-dev/os/pull/24903) This frees melange from fiddling with the rootfs, and rely on apko for the cpio build.
> 
> Also now, the ssh key is shared using a network mount (9p) and set up in the init itself.
> 
> Some fiddling of the rootfs is needed in case we need to specify kernel modules: `QEMU_KERNEL_IMAGE=/path/to/vmlinuz` points to a `vmlinuz` file, but most of them (if not compiled with `make kvm_guest`) will miss the virtio and 9p drivers by default in those case we specify the `QEMU_KERNEL_MODULES=/path/to/lib/modules/` and we inject that kernel's modules into the apko's generated v1.Layer, then use apko's new cpio build on that.
> 
> Also introduced some `ssh` improvements suggested by @smoser 👍

Continuing on this, I've added the ability to specify an external disk to use for the compilation (/home/build) in case it's needed
It supports both CLI flags and Yaml resources 

```yaml
resources:
  cpu: 4
  memory: 12Gi
  disk: 100Gi
```

Disk is allocated using the `fallocate` syscall so it's fast to create, and the `melange-microvm-init` will detect it and `mkfs.ext4` it

This allows to compile build packages like `gcc`, as demonstrated here:
Screenshot from 2024-07-26 11-47-47

Without disk it fails for out-of-space
Also a comparison of performance hit vs bubblewrap

pkg/container/qemu_runner.go
```diff
@@ -0,0 +1,760 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// WorkspaceTar implements Runner
+func (bw *qemu) WorkspaceTar(ctx context.Context, cfg *Config) (io.ReadCloser, error) {
```
smoser, line 210: this never returns anything other than nil?
i'm guessing this is just in progress.

89luca89: for some runners (see also bwrap) this is a noop
in this case it's not properly a noop, but does not return a tar indeed

89luca89:
> > @mattmoor @dustinkirkland @imjasonh
> > Moved the init to the following subpackage: [wolfi-dev/os#24903](https://github.com/wolfi-dev/os/pull/24903) This frees melange from fiddling with the rootfs, and rely on apko for the cpio build.
> > Also now, the ssh key is shared using a network mount (9p) and set up in the init itself.
> > Some fiddling of the rootfs is needed in case we need to specify kernel modules: `QEMU_KERNEL_IMAGE=/path/to/vmlinuz` points to a `vmlinuz` file, but most of them (if not compiled with `make kvm_guest`) will miss the virtio and 9p drivers by default in those case we specify the `QEMU_KERNEL_MODULES=/path/to/lib/modules/` and we inject that kernel's modules into the apko's generated v1.Layer, then use apko's new cpio build on that.
> > Also introduced some `ssh` improvements suggested by @smoser 👍
> 
> Continuing on this, I've added the ability to specify an external disk to use for the compilation (/home/build) in case it's needed It supports both CLI flags and Yaml resources
> 
> ```yaml
> resources:
>   cpu: 4
>   memory: 12Gi
>   disk: 100Gi
> ```
> 
> Disk is allocated using the `truncate` syscall so it's fast to create, and the `melange-microvm-init` will detect it and `mkfs.ext4` it
> 
> This allows to compile build packages like `gcc`, as demonstrated here: Screenshot from 2024-07-26 11-47-47
> 
> Without disk it fails for out-of-space Also a comparison of performance hit vs bubblewrap

Expanding on this, I've made so that disks are used by default.

Using `os.Truncate` files are sparse, and they only occupy the space needed when filled.

Implemented some mount optimizations in the init (here: https://github.com/wolfi-dev/os/pull/24903) so that it's mounted fast and with performance in mind.

pkg/container/qemu_runner.go
89luca89: In the meantime I've improved the sshd startup time, improving the entropy (passing the host's one) and properly checking for the server to come up (instead of using ssh client that is way slower)

pkg/container/qemu_runner.go
```diff
@@ -0,0 +1,744 @@
...
+	diskFile, err := generateDiskFile(ctx, cfg.Disk)
+	if err != nil {
+		clog.FromContext(ctx).Errorf("qemu: could not generate additional disks: %v", err)
+		return err
+	}
+
+	// append raw disk, init will take care of formatting it if present.
+	baseargs = append(baseargs, "-object", "iothread,id=io1")
+	baseargs = append(baseargs, "-device", "virtio-blk-pci,drive=disk0,iothread=io1")
+	baseargs = append(baseargs, "-drive", "if=none,id=disk0,cache=none,format=raw,aio=threads,werror=report,rerror=report,file="+diskFile)
```
smoser, line 370: per qemu-system-x86 [man page](https://manpages.ubuntu.com/manpages/noble/man1/qemu-system.1.html) :

> The  most explicit way to describe disks is to use a combination of -device to specify the
> hardware device and -blockdev to describe the backend. The device defines what  the  guest
> sees and the backend describes how QEMU handles the data. It is the only guaranteed stable
> interface for describing block devices and as such is recommended for management tools and
> scripting.
>
> The  -drive option combines the device and backend into a single command line option which
> is a more human friendly. There is however no interface stability guarantee although  some
> older board models still need updating to work with the modern blockdev forms.

So it would be better here to use a `-device` and `-blockdev`.  Unfortunately I don't have an example of that.

pkg/container/qemu_runner.go
```diff
@@ -0,0 +1,744 @@
...
+
+	if cfg.Memory != "" {
+		memKb, err := convertMemoryToKB(cfg.Memory)
+		if err != nil {
+			return err
+		}
+
+		baseargs = append(baseargs, "-m", fmt.Sprintf("%dk", memKb))
+	} else {
+		baseargs = append(baseargs, "-m", getAvailableMemoryKB())
```
smoser, line 311: this is kind of weird.
`"-m", getAvailableMemoryKB()` reads like "pass all the memory in the system".
versus:

`"-m", getAvailableMemoryKB()/2` or something.

pkg/container/qemu_runner.go
```diff
@@ -0,0 +1,744 @@
...
+	if err != nil {
+		clog.FromContext(ctx).Errorf("could not prepare rootfs: %v", err)
+		return err
+	}
+
+	baseargs := []string{}
+	// load microvm profile and bios, shave some milliseconds from boot
+	// using this will make a complete boot->initrd (with working network) In ~700ms
+	// instead of ~900ms.
+	if _, err := os.Stat("/usr/share/qemu/bios-microvm.bin"); err == nil {
```
smoser, line 294: this is a rathole, but on ubuntu 24.04

```
$ ls -l /usr/share/qemu/bios-microvm.bin
ls: cannot access '/usr/share/qemu/bios-microvm.bin': No such file or directory
$ dpkg -S /usr/share/seabios/bios-microvm.bin
seabios: /usr/share/seabios/bios-microvm.bin
```

pkg/container/qemu_runner.go
smoser: +1 on the hosts entropy.

pkg/container/qemu_runner.go
```diff
@@ -0,0 +1,744 @@
...
+	publicKey, err := ssh.NewPublicKey(&privateKey.PublicKey)
+	if err != nil {
+		clog.FromContext(ctx).Errorf("qemu: ssh keygen failed: %v", err)
+		return nil, err
+	}
+
+	return ssh.MarshalAuthorizedKey(publicKey), nil
+}
+
+func randpomPortN() (int, error) {
```
smoser, line 665: rand**p**omPortN ? spelling?

pkg/container/qemu_runner.go
89luca89: Will test

`	baseargs = append(baseargs, "-blockdev", "file,node-name=disk0,locking=off,cache.direct=off,discard=ignore,aio=threads,filename="+diskFile)`

seems like it should cover the current command

pkg/container/qemu_runner.go
```diff
@@ -0,0 +1,744 @@
...
+	// but still useful otherwise
+	baseargs = append(baseargs, "-display", "none")
+	baseargs = append(baseargs, "-no-reboot")
+	baseargs = append(baseargs, "-no-user-config")
+	baseargs = append(baseargs, "-nodefaults")
+	baseargs = append(baseargs, "-parallel", "none")
+	baseargs = append(baseargs, "-serial", "none")
+	baseargs = append(baseargs, "-vga", "none")
+	// use -netdev + -device instead of -nic, as this is better supported by microvm machine type
+	baseargs = append(baseargs, "-netdev", "user,id=id1,hostfwd=tcp::"+cfg.SSHPort+"-:22")
```
smoser, line 339: For future, and micro optimization.  In order to avoid the "find a random , hopefully  unused port" thing, you can do 
1. pick a random 127.0.0.0/8 address - you can use port 22 on any one of the 16 million (2^24) host addresses on the localhost network. 
2. use port 0 and then ask the qemu monitor which one it got.

pkg/container/qemu_runner.go
smoser: If this is non-trivial, then we can leave that to future.

smoser (commented):
I think this is really close.  I'm personally happy to get it in and iterate. 

if any of these are sticking points we could address later.

mattmoor:
@89luca89 had a chance to try this on my M2 and I needed to pass `-machine virt` as well, but after I added that it worked for me!

```
2024/07/30 13:12:21 INFO scanning for shbang deps...
2024/07/30 13:12:21 INFO   installed-size: 25914332
2024/07/30 13:12:21 INFO   data.tar.gz digest: d118b375c998165869a496ccbb304903064e584f4f43e335753eae6ea07b9498
2024/07/30 13:12:21 INFO wrote packages/aarch64/ko-0.15.4-r5.apk
2024/07/30 13:12:22 INFO generating apk index from packages in packages/aarch64
2024/07/30 13:12:22 INFO processing package packages/aarch64/ko-0.15.4-r5.apk
2024/07/30 13:12:22 INFO loaded 2/2 packages from index packages/aarch64/APKINDEX.tar.gz
2024/07/30 13:12:22 INFO updating index at packages/aarch64/APKINDEX.tar.gz with new packages: [ko-0.15.4-r5]
2024/07/30 13:12:22 INFO signing apk index at packages/aarch64/APKINDEX.tar.gz
2024/07/30 13:12:22 INFO signing index packages/aarch64/APKINDEX.tar.gz with key local-melange.rsa
2024/07/30 13:12:22 INFO appending signature to index packages/aarch64/APKINDEX.tar.gz
2024/07/30 13:12:22 INFO writing signed index to packages/aarch64/APKINDEX.tar.gz
2024/07/30 13:12:22 INFO signed index packages/aarch64/APKINDEX.tar.gz with key local-melange.rsa
2024/07/30 13:12:22 INFO qemu: sending shutdown signal
```

smoser:
Some of my feedback is addressed https://github.com/89luca89/melange/pull/1

pkg/container/qemu_runner.go
```diff
@@ -0,0 +1,744 @@
...
+		err = session.Setenv(k, v)
+		if err != nil {
+			return err
+		}
+	}
+
+	session.Stdin = stdin
+	session.Stderr = stderr
+	session.Stdout = stdout
+	err = session.Run("set -e;" + strings.Join(command, " "))
```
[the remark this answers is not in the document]
89luca89: Resolved

pkg/container/qemu_runner.go
89luca89: Resolved

pkg/container/qemu_runner.go
89luca89: Resolved

89luca89:
> @89luca89 had a chance to try this on my M2 and I needed to pass `-machine virt` as well, but after I added that it worked for me!
> 
> ```
> 2024/07/30 13:12:21 INFO scanning for shbang deps...
> 2024/07/30 13:12:21 INFO   installed-size: 25914332
> 2024/07/30 13:12:21 INFO   data.tar.gz digest: d118b375c998165869a496ccbb304903064e584f4f43e335753eae6ea07b9498
> 2024/07/30 13:12:21 INFO wrote packages/aarch64/ko-0.15.4-r5.apk
> 2024/07/30 13:12:22 INFO generating apk index from packages in packages/aarch64
> 2024/07/30 13:12:22 INFO processing package packages/aarch64/ko-0.15.4-r5.apk
> 2024/07/30 13:12:22 INFO loaded 2/2 packages from index packages/aarch64/APKINDEX.tar.gz
> 2024/07/30 13:12:22 INFO updating index at packages/aarch64/APKINDEX.tar.gz with new packages: [ko-0.15.4-r5]
> 2024/07/30 13:12:22 INFO signing apk index at packages/aarch64/APKINDEX.tar.gz
> 2024/07/30 13:12:22 INFO signing index packages/aarch64/APKINDEX.tar.gz with key local-melange.rsa
> 2024/07/30 13:12:22 INFO appending signature to index packages/aarch64/APKINDEX.tar.gz
> 2024/07/30 13:12:22 INFO writing signed index to packages/aarch64/APKINDEX.tar.gz
> 2024/07/30 13:12:22 INFO signed index packages/aarch64/APKINDEX.tar.gz with key local-melange.rsa
> 2024/07/30 13:12:22 INFO qemu: sending shutdown signal
> ```

Fixed now :+1: 



> Some of my feedback is addressed [89luca89#1](https://github.com/89luca89/melange/pull/1)

Thanks a lot!

pkg/container/qemu_runner.go
89luca89: Removed

pkg/container/qemu_runner.go
89luca89: Implemented this method instead :+1:

pkg/container/qemu_runner.go
smoser: > Sadly -blockdev is really poorly documented :/

Yeah, my source of documentation has always been either source code or "what does libvirt do here".

I've just been bit by the -drive option before, having it not work in between versions.
its fine as TODO.

pkg/container/qemu_runner.go
smoser: We are intentionally passing '-m' a value that indicates all the system memory.  Am I reading that right?

Even in a single hardware machine that is provisioned to do nothing other than that, we'd be over-committing the memory, right?

mattmoor:
@89luca89 needs a rebase too

pkg/container/qemu_runner.go
89luca89: if nothing is specified it uses available memory
thi shouldn't be a problem at least on linux as qemu only uses whatever it needs, it does not block-allocate all the ram

Else we do 1/2 of available memory? wdyt?

89luca89:
> @89luca89 needs a rebase too

Rebased :+1:  I think we will need to cut a new apko release to include your PR as right now I'm using a `replace` to access it

mattmoor:
@89luca89 you shouldn't need a release or replace to `go get chainguard.dev/apko@main` to point at a commit for now, but I'd be cool with a release as well!

pkg/container/qemu_runner.go
smoser: I guess we can try whatever and see how it works.

It becomes weird when the only way you can *not* use all the system memory is for the melange yaml file you're using to specify; Generally speaking, that configuration in the melange yaml is saying "I need more resources" but here its the only way to turn things down.

I guess really we should do something like:

```go
requested := cfg.Memory
system := getAvailableMemoryKB()
if requested > 0 && requested > system  {
   panic("BAD")
}
baseargs = append(baseargs, "-m", system)
```

smoser:
@89luca89 ,

What more do you think you need here?

pkg/container/qemu_runner.go
89luca89: Perfect, defaulted to 1/4th of available memory, and check if cfg.mem > available mem and error out :+1:

89luca89:
> @89luca89 you shouldn't need a release or replace to `go get chainguard.dev/apko@main` to point at a commit for now, but I'd be cool with a release as well!

done :+1: 

> @89luca89 ,
> 
> What more do you think you need here?

Just working on adding a test in the CI, code wise I think we're good to go just working on the tests :+1:

pkg/container/qemu_runner.go
```diff
@@ -0,0 +1,761 @@
...
+		"Gi": 1024 * 1024 * 1024,        // Gibibytes
+		"Ti": 1024 * 1024 * 1024 * 1024, // Tebibytes
+		"K":  1024,                      // Kilobytes (KB)
+		"M":  1024 * 1024,               // Megabytes (MB)
+		"G":  1024 * 1024 * 1024,        // Gigabytes (GB)
+		"T":  1024 * 1024 * 1024 * 1024, // Terabytes (TB)
+		"k":  1024,                      // Kilobytes (KB)
+		"m":  1024 * 1024,               // Megabytes (MB)
+		"g":  1024 * 1024 * 1024,        // Gigabytes (GB)
+		"t":  1024 * 1024 * 1024 * 1024, // Terabytes (TB)
```
smoser, line 708: oh yeah, you should fix the comments or the numbers.
Either comment 'Kibibytes' or change it to 1000.

mattmoor (approved):
@smoser let me know if you have any more comments here, otherwise I’ll do one final test locally with this, and adjust the branch protections to land it in a few hours.

thanks @89luca89 !  This is awesome 🤩

smoser (dismissed):
I'm happy with this. Thank you @89luca89 . 
Well done.

smoser:
> [Matt] @smoser let me know if you have any more comments here, otherwise I’ll do one final test locally with this, and adjust the branch protections to land it in a few hours.

It has my approval, but I think @luca was wanting to add some tests.

> [Luca] Just working on adding a test in the CI, code wise I think we're good to go just working on the tests 👍

@89luca89 did you do that?

mattmoor (dismissed):

mattmoor:
Hmm, I'm seeing two things trying to use the latest bits on my M1:
1. The following error:
```
2024/08/02 13:18:58 ERRO qemu: failed to run qemu command: exit status 1 - qemu-system-aarch64: -netdev user,id=id1,hostfwd=tcp:127.84.156.78:36564-:22: Could not set up host forwarding rule 'tcp:127.84.156.78:36564-:22'
```
2. I see it leaving around `.img` files in the working directory (dirtying the workspace) 😞

89luca89:
> > [Matt] @smoser let me know if you have any more comments here, otherwise I’ll do one final test locally with this, and adjust the branch protections to land it in a few hours.
> 
> It has my approval, but I think @luca was wanting to add some tests.
> 
> > [Luca] Just working on adding a test in the CI, code wise I think we're good to go just working on the tests 👍
> 
> @89luca89 did you do that?

Yep added the qemu build tests into the CI



> Hmm, I'm seeing two things trying to use the latest bits on my M1:
> 
> 1. The following error:
> 
> ```
> 2024/08/02 13:18:58 ERRO qemu: failed to run qemu command: exit status 1 - qemu-system-aarch64: -netdev user,id=id1,hostfwd=tcp:127.84.156.78:36564-:22: Could not set up host forwarding rule 'tcp:127.84.156.78:36564-:22'
> ```
> 
> 2. I see it leaving around `.img` files in the working directory (dirtying the workspace) 😞

1. I think that's a regression introduced by trying to use addresses in the 127.0.0.0/8 instead of always using `localhost`, I'll revert that change

2. seems like it's treated like the workspace (same life cycle) so in case of error it's left there, I'll find a way to cleanup things :+1:

smoser:
> I think that's a regression introduced by trying to use addresses in the 127.0.0.0/8 instead of always using localhost, I'll revert that change

thats a total bummer. 

I'm fine with the random port thing.  At some point we'll update to specify port 0 and then get the info back from the qemu monitor on what port it did use.

smoser:
> seems like it's treated like the workspace (same life cycle) so in case of error it's left there, I'll find a way to cleanup things 👍

melange _does_ have a more general issue here, in that it leaves workspace `melange-guest` and `melange-workspace` files in TMPDIR .  This was discussed recently on slack.  I think the right solution for _those_ is to implement "keep" with values "none|fail|always". It feels like these disk files are just an extension of that.  qemu runner's workspace is the disk image, until we can manage to get rid of the image files.  

Would it make sense to just put the disk image files inside the melange-workspace dir and let it be cleaned up by that?

89luca89:
> Would it make sense to just put the disk image files inside the melange-workspace dir and let it be cleaned up by that?

Don't know if that's a problem, usually this will fill TMPDIR which limits our disk growth to the tmpfs size
Also there is a phase where we actually copy the workspace dir from the shared netmount to the disk, this would create a loop


Right now the disks are correctly cleared up (in case of failure) together with the rest

imjasonh:
I'd like to get this merged and start testing it out. Are there any blocking concerns that should block that?

mattmoor:
Needs a rebase, but +1 from me!  I think we may need to confirm it works on M1?

89luca89:
> Needs a rebase, but +1 from me! I think we may need to confirm it works on M1?

did the rebase, can't test on Apple Mseries, @imjasonh can you?

mattmoor:
Testing this now

mattmoor:
It works, I am going to change the presubmits and merge this!

89luca89:
> It works, I am going to change the presubmits and merge this!

Thanks
Can I suggest a squash+merge? there are a touch too many commits here :joy:

mattmoor (approved):

mattmoor:
lol, yeah I squash merged it

89luca89:
Thanks!

PR #1394, opened 2024-07-19
sca: Allow non-executable .so to provide/vendor

jonjohnsonjr:
For executable files, we expect the executable bit to be set; however, the exec bit doesn't actually need to be set for shared objects. When we are computing provided/vendored .so files, we were ignoring any that didn't have the exec bit set, which meant a lot of python packages were unexpectedly pulling in dependencies when they should have been self-provided.

### DIFF
--- a/pkg/sca/sca.go
+++ b/pkg/sca/sca.go
@@ -268,10 +268,6 @@ func generateSharedObjectNameDeps(ctx context.Context, hdl SCAHandle, generated 
 			return nil
 		}
 
-		if mode.Perm()&0555 != 0555 {
-			return nil
-		}
-
 		basename := filepath.Base(path)
 
 		// most likely a shell script instead of an ELF, so treat any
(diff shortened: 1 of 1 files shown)


jonjohnsonjr:
X-ref: https://github.com/wolfi-dev/os/pull/24325

If I drop the `no-depends`:

```diff
diff --git a/airflow.yaml b/airflow.yaml
index 0c08c1ead..5711a33c3 100644
--- a/airflow.yaml
+++ b/airflow.yaml
@@ -3,11 +3,6 @@ package:
   version: 2.9.3
   epoch: 0
   description: Platform to programmatically author, schedule, and monitor workflows
-  options:
-    #  There is a dependency on libarrow.so although it
-    #  is provided in the virtual environment. Enabling no-depends
-    #  works around this
-    no-depends: true
   copyright:
     - license: Apache-2.0
```

Then melange erroneously treated these as `depend` before and correctly `# vendored` (and self-satisfied) after:

```diff
14,18d13
< depend = so:libarrow.so.1700
< depend = so:libarrow_acero.so.1700
< depend = so:libarrow_dataset.so.1700
< depend = so:libarrow_flight.so.1700
< depend = so:libarrow_substrait.so.1700
28d22
< depend = so:libparquet.so.1700
36a31,34
> # vendored = so:libarrow.so.1700=1700
> # vendored = so:libarrow_acero.so.1700=1700
> # vendored = so:libarrow_dataset.so.1700=1700
> # vendored = so:libarrow_flight.so.1700=1700
39a38
> # vendored = so:libarrow_substrait.so.1700=1700
63a63
> # vendored = so:libparquet.so.1700=1700
```

PR #1404, opened 2024-07-29
Bump apko dependency

mattmoor:
I believe this is needed to pick up the key discovery so we can drop the hard-coded keys in Makefiles 🤞


cpanato:
```
Error return value of `auth.DefaultAuthenticators.AddAuth` is not checked (errcheck)
```

mattmoor:
@cpanato Yeah, we added an `error` return.  I fixed it up.

imjasonh (approved):

cpanato (approved):

k4leung4 (approved):

Issue #1407, opened 2024-07-29
bubblewrap environment does not include HOME in test, docker runner does


smoser:
As seen https://github.com/wolfi-dev/os/commit/9a36346964d080e69196746ea5b804d5363b3280
I have had to add 'HOME' to the environment for running of tests in order for them to run on bwrap runner.

The build environment does get HOME=/home/build but test environment it is not set.

Given the yaml below, with my melange:
```
$ melange version | grep '^[A-Z]'
GitVersion:    v0.11.1-31-g4b59659
GitCommit:     4b59659dbfa5c422a79e059d4585806abaa0d4e1
GitTreeState:  clean
BuildDate:     '2024-07-29T15:47:22Z'
GoVersion:     go1.22.5
Compiler:      gc
Platform:      linux/amd64
```

Running this `test-me.yaml`:

```yaml
package:
  name: test-me
  version: 1.0
  epoch: 0

environment:
  contents:
    packages:
      - busybox

pipeline:
  - runs: |
      showhome=${{targets.contextdir}}/usr/bin/show-home
      mkdir -p $(dirname "$showhome")
      printf '#!/bin/sh\necho HOME="$HOME"\n' > "$showhome"
      chmod 755 "$showhome"

      echo === build says ===
      "$showhome"

test:
  pipeline:
    - name: Run the showhome
      runs: |
        out=$(show-home)
        echo "got: $out"
        case "$out" in
          *HOME=) echo "home not set"; exit 1;;
          *HOME=) echo "got home=${out#HOME=}";;
        esac
```

We see:

    $ make MELANGE_EXTRA_OPTS=--runner=docker test/test-me
    /home/smoser/go/bin/melange test test-me.yaml \
         --repository-append /home/smoser/src/wolfi/os/packages \
         --keyring-append local-melange.rsa.pub --arch x86_64 \
         --pipeline-dirs ./pipelines/ \
         --repository-append https://packages.wolfi.dev/os \
         --keyring-append https://packages.wolfi.dev/os/wolfi-signing.rsa.pub \
         --test-package-append wolfi-base \
         --debug --runner=docker --source-dir ./test-me/
    ...
    2024/07/29 12:44:40 INFO got: HOME=/root

But with bwrap, we see:

    $ make MELANGE_EXTRA_OPTS=--runner=bubblewrap test/test-me
    melange test test-me.yaml --repository-append /home/smoser/src/wolfi/os/packages \
[...]

PR #1408, opened 2024-07-29
Add HOME=/root to default test environment.

smoser:
HOME was set in the test environment by docker runner, but not by the bwrap runner.  The docker runner was mostly happenstance be explicit and set it to /root.

Fixes https://github.com/chainguard-dev/melange/issues/1407


pkg/build/test.go
```diff
@@ -426,6 +426,10 @@ func (t *Test) TestPackage(ctx context.Context) error {
 		return fmt.Errorf("unable to build workspace config: %w", err)
 	}
 
+	if _, ok := cfg.Environment["HOME"]; !ok {
+		cfg.Environment["HOME"] = "/root"
+	}
+
```
jdolitsky, line 432: Can we do this inside of `buildWorkspaceConfig` instead?

smoser: ack. yeah, that does seem better.

imjasonh (approved):

Issue #1409, opened 2024-07-30
incorrect ruby-3.1 dependency generated for vendored jruby gems


xnox:
Files of the form

```
usr/share/logstash/vendor/bundle/jruby/3.1.0/gems
```

Appear to generate runtime dependency on `ruby-3.1` when jruby is used instead.

Also bundled gems possibly need not to generate public dependency on `ruby-3.x` at all. In case they vendor in ruby interpreter too, like this package does.

PR #1410, opened 2024-07-30
fix ruby sca

xnox:
- **sca: add ruby test case**
  

- **sca: Use glob implementation for ruby versions**
  Find exact ruby gem directories for SCA dependency generation. This
  avoids generating ruby-M.N dependency for private jruby vendored gems.
  
  Fixes: #1409


EyeCantCU (approved):
This is a lot better. Thank you so much!

PR #1412, opened 2024-08-01
update config: provide configuration to describe polling and schedules

rawlingsj:
### Functional Changes

Notes: This is additional schema changes but no implementation as it is expected that other tooling works with these config options.


pkg/config/schema.json
```diff
@@ -327,9 +327,6 @@
           },
           "type": "array"
         },
-        "os-release": {
```
rawlingsj, line 330: I didn't touch this but after running `make generate` this and the `OSRelease` below were removed.

pkg/config/config.go
```diff
@@ -492,6 +492,10 @@ type Update struct {
 	VersionTransform []VersionTransform `json:"version-transform,omitempty" yaml:"version-transform,omitempty"`
 	// ExcludeReason is required if enabled=false, to explain why updates are disabled.
 	ExcludeReason string `json:"exclude-reason,omitempty" yaml:"exclude-reason,omitempty"`
+	// Polling indicates that polling should be used to find latest versions
+	Polling *Polling `json:"polling,omitempty" yaml:"polling"`
```
luhring, line 496: Just curious, should this have `omitempty` for YAML too?

rawlingsj: yes! nice catch :)

luhring (approved):

PR #1414, opened 2024-08-05
refactor runner implementations

imjasonh:
This refactors and centralizes how runner implementations are registered and selected. Existing runners (docker, bubblewrap, experimentaldagger) are unaffected, and can still be selected automatically or with the `--runner` flag.

This enables other callers of melange's Go library to register their own implementations without having to modify melange's source code. Registering can be done in `func init` or in `func main`, any time before `GetRunner` is called.


pkg/cli/build.go
```diff
@@ -229,31 +226,6 @@ func buildCmd() *cobra.Command {
...
-		case "docker":
-			return docker.NewRunner(ctx)
-		case "experimentaldagger":
-			return dagger.NewRunner(ctx)
-		default:
-			return nil, fmt.Errorf("unknown runner: %s", runner)
-		}
-	}
-
-	switch runtime.GOOS {
```
89luca89, line 246: I think with the new implementation we're missing this, so e2e are failing because we don't have a `default` now, but it expect explicitly to be set

PR #1421, opened 2024-08-06
update config: replace recently added polling with git struct

rawlingsj:
`Polling` was recently added in https://github.com/chainguard-dev/melange/pull/1412, there's no implementation yet for this.  After adding support for native `Git` it seems polling would add confusion and not needed.  Let's just use Git.

We are also adding an interface that defines common methods for retrieving version filtering and stripping parameters for GitHub and Git implementations.

### Functional Changes


pkg/config/config.go
```diff
@@ -538,17 +599,11 @@ type VersionTransform struct {
...
-type Polling struct {
-	// Git should be used to find the latest version
-	Git bool `json:"git,omitempty" yaml:"git,omitempty"`
-	// The reason polling is being used
-	Reason string `json:"reason,omitempty" yaml:"reason,omitempty"`
+	// The reason scheduling is being used
+	Reason  string `json:"reason,omitempty" yaml:"reason,omitempty"`
+	Daily   bool   `json:"daily,omitempty" yaml:"daily,omitempty"`
+	Weekly  bool   `json:"weekly,omitempty" yaml:"weekly,omitempty"`
+	Monthly bool   `json:"monthly,omitempty" yaml:"monthly,omitempty"`
```
imjasonh, line 606: Should this be something like `period: daily|weekly|monthly` (a string enum thing)?

Will something ever be both `daily: true` and `monthly: true` ? 🤔

imjasonh (approved):
one non-blocking comment

pkg/config/config.go
rawlingsj: yeah sounds good to me!

rawlingsj:
@imjasonh this is ready for another look when you get the chance.

imjasonh (approved):

PR #1433, opened 2024-08-14
Centralize sca options handling

jonjohnsonjr:
Returning early is nice to avoid doing extra work, but it confuses the control flow and makes it too easy to miss places where we should be skipping certain provides. This centralizes where we check for no-provides, no-commands, and no-depends so that we can't do this accidentally again.

### DIFF
--- a/pkg/sca/sca.go
+++ b/pkg/sca/sca.go
@@ -24,6 +24,7 @@ import (
 	"io/fs"
 	"path/filepath"
 	"regexp"
+	"slices"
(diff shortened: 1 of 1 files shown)


pnasrat (commented):
LGTM

Although what triggered the issue I had was in part my config, this is clearer, fixes edge cases I was also hitting, and allows for more extensible handling of future options eg to filter out specific provides.

pnasrat (approved):

PR #1439, opened 2024-08-19
Add test to catch duplicate package names

jonjohnsonjr:
We had a test for duplicate subpackages names, but it didn't catch the case where a subpackage had the same name as the main package.


pnasrat (dismissed):
LGTM

k4leung4 (approved):

PR #1443, opened 2024-08-19
upgrade to golang 1.23

k4leung4:


.github/workflows/verify.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@aaa42aa0628b4ae2578232a66b541047968fac86 # v3.1.0
         with:
-          version: v1.59
+          version: v1.60
```
p5, line 32:
```suggestion
          version: ${{ steps.setup-go.outputs.go-version }}
```
Untested, but you could likely use the actions/setup-go output in this input. 
actions/setup-go exposes a `go-version` output ([here](https://github.com/actions/setup-go/blob/main/action.yml#L23)).

This means, you can change this and the step above to something like:
```yaml
      - uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5.0.2
        id: setup-go
        with:
          go-version-file: go.mod
check-latest: true

      - name: golangci-lint
        uses: golangci/golangci-lint-action@aaa42aa0628b4ae2578232a66b541047968fac86 # v3.1.0
        with:
          version: ${{ steps.setup-go.outputs.go-version }}
```

k4leung4: thanks
though this is asking for the version of golangci and not go itself

k4leung4: the failure is due to the version of go in alpine is not at 1.23 yet;
https://github.com/chainguard-dev/melange/blob/main/melange.yaml#L23

p5: Ah, very sorry.  I misunderstood, but it looks obvious now I'm looking at it.  :facepalm:

cpanato (approved):
thanks

PR #1450, opened 2024-08-21
if-arch: only run the pipeline if runtime arch matches

imjasonh:
Split out of #1449 

We need this first so we can migrate existing packages to `if-arch` before removing `if` and `assertions`.


xnox:
do we really need this? I'd rather just use shell and compare the `${{build.arch}}` string and exit 0, or continue executing pipeline. etc.

Such that we can kill all non-determinism.

xnox:
@imjasonh do you want me to go rip all of these things out, in all the places?

PR #1459, opened 2024-08-28
More SBOM logic improvements

luhring:
This PR continues to improve the state of melange's SBOM code:

1. Fixes the issue where our external references used `PACKAGE_MANAGER` instead of `PACKAGE-MANAGER`.
2. Breaks apart "SBOM generation" logic from "writing the SBOM back into the APK's filesystem" logic. The hope is we can then use this seam to implement tests in the near future.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -497,6 +497,10 @@ func (b *Build) IsBuildLess() bool {
 // ConfigFileExternalRef calculates ExternalRef for the melange config
 // file itself.
 func (b *Build) ConfigFileExternalRef() (*purl.PackageURL, error) {
+	// TODO(luhring): This is now the second implementation of finding the commit
+	//  for the config file (the first being the "detectedCommit" logic. We should
+	//  unify these.
+
 	// configFile must exist
 	configpath, err := filepath.Abs(b.ConfigFile)
 	if err != nil {
@@ -508,6 +512,11 @@ func (b *Build) ConfigFileExternalRef() (*purl.PackageURL, error) {
 	if err != nil {
 		return nil, nil
 	}
+
+	// TODO(luhring): This is brittle and assumes a specific git remote configuration.
+	//  We should consider a more general approach, and this may be moot when we
+	//  unify our git state detection mechanisms.
+
 	// If no remote origin, skip (local git repo)
 	remote, err := r.Remote("origin")
 	if err != nil {
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -871,8 +871,9 @@ func (b *Build) BuildPackage(ctx context.Context) error {
 		sp := sp
 
 		log.Infof("generating SBOM for subpackage %s", sp.Name)
-		if err := sbom.Generate(ctx, &sbom.Spec{
-			Path:            filepath.Join(b.WorkspaceDir, "melange-out", sp.Name),
+
+		apkFSPath := filepath.Join(b.WorkspaceDir, "melange-out", sp.Name)
+		if err := sbom.GenerateAndWrite(ctx, apkFSPath, &sbom.Spec{
 			PackageName:     sp.Name,
 			PackageVersion:  fmt.Sprintf("%s-r%d", b.Configuration.Package.Version, b.Configuration.Package.Epoch),
 			License:         b.Configuration.Package.LicenseExpression(),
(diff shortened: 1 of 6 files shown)


pkg/sbom/implementation.go
```diff
@@ -75,11 +75,12 @@ func encodeInvalidRune(r rune) string {
...
-// checkEnvironment returns a bool indicating if Spec's Path exists. If the path
-// does not exist, it returns false and a nil error. If an error occurs while
-// checking the directory, it returns false and the error.
-func checkEnvironment(spec *Spec) (bool, error) {
-	dirPath, err := filepath.Abs(spec.Path)
+// checkEnvironment returns a bool indicating if the specified path for an APK's
+// filesystem exists. If the path does not exist, it returns false and a nil
+// error. If an error occurs while checking the directory, it returns false and
+// the error.
+func checkEnvironment(apkFSPath string) (bool, error) {
```
vaikas, line 82: super nit and certainly not your doing, but checkEnvironment seems a bit confusing vs. checkPathExists or something like that. Environment has so many loaded meanings.

pkg/sbom/implementation.go
```diff
@@ -155,7 +156,7 @@ func addPackage(doc *spdx.Document, p *pkg) {
 	}
 
 	// Add the purl to the package
-	const extRefCatPackageManager = "PACKAGE_MANAGER"
+	const extRefCatPackageManager = "PACKAGE-MANAGER"
```
vaikas, line 159: nit, why not hoist to package level constant for easier discoverability?

pkg/sbom/implementation.go
```diff
@@ -95,14 +96,14 @@ func checkEnvironment(spec *Spec) (bool, error) {
 	return true, nil
 }
 
-// generateAPKPackage generates the sbom package representing the apk
-func generateAPKPackage(spec *Spec) (pkg, error) {
+// generateAPKPackage generates the sbom package representing the apk.
+func generateAPKPackage(spec *Spec) (*pkg, error) {
```
vaikas, line 100: nit: If I saw this in other code I'd assume it creates the actual APKPackage, again, not your doing, but I'd consider renaming this to generateAPKSBOM or something like that.

pkg/build/build.go
```diff
@@ -497,6 +497,10 @@ func (b *Build) IsBuildLess() bool {
 // ConfigFileExternalRef calculates ExternalRef for the melange config
 // file itself.
 func (b *Build) ConfigFileExternalRef() (*purl.PackageURL, error) {
+	// TODO(luhring): This is now the second implementation of finding the commit
```
vaikas, line 500: Are these TODOs worthy of an issue? If I were looking at the second implementation, would I know of this one?

vaikas (commented):
super nitty nits, but are there tests that exist? Could we add some if they don't?

pkg/sbom/implementation.go
luhring: I agree, the name is vague and unhelpful — time to change it...

pkg/sbom/implementation.go
luhring: I definitely can — I thought about this, and wasn't sure of our stylistic preference. A theme from our recent style guide discussions is to keep value definitions as local as possible to where they're used. But this seems like a general SPDX-ism that's perhaps worthy of hoisting

pkg/sbom/implementation.go
luhring: Agreed 1000%

pkg/build/build.go
luhring: >Are these TODOs worthy of an issue?

Yes! I'll open one and link it here...

>If I were looking at the second implementation, would I know of this one?

Good point — do you have a specific solution in mind? I can add a similar comment to the other implementation, for example.

pkg/sbom/implementation.go
luhring: 7209756

pkg/sbom/implementation.go
luhring: 9f3a18f

pkg/build/build.go
luhring: c7f9e86 adds a cross-referencing comment

luhring: https://github.com/chainguard-dev/melange/issues/1460

luhring:
> super nitty nits, but are there tests that exist? Could we add some if they don't?

@vaikas There are no tests, and there haven't ever been as far as I can tell. 😞 

We can add an E2E test that asserts the SBOM in an APK against a goldenfile or something. From what I can tell there's not yet a great place to put in tests for our SBOM logic, because of how tangled the data flow is throughout these giant structs that depend on mutations using data from all sorts of places. But I think this PR makes progress toward a more "testable" state, FWIW.

vaikas (approved):
Thank you!

jdolitsky (approved):

jdolitsky (approved):

Issue #1172, opened 2024-04-24
pkg-config versions may not align with apk versions


jonjohnsonjr:
We ran into this with bash-dev which has:

```
curl -sL https://packages.wolfi.dev/os/aarch64/bash-dev-5.2.21-r1.apk | tar -Oxz usr/lib/pkgconfig/bash.pc | grep Version
Version: 5.2-release
```

Which makes `apk add` fail (but for some reason only with a local `.apk`???) which means we can't bump bash in wolfi.

We kind of try to do this with https://github.com/chainguard-dev/melange/blob/0a50317b8fe3d7fd4c3cd6cdcb658752a6f992d5/pkg/sca/sca.go#L387 but that only works for suffixes that also happen to be valid apk suffixes.

I'm not sure how we want to handle this, maybe we just ignore versions that aren't valid? Do we try to trim suffixes until it's a valid version? Who knows.

xnox:
IMHO the version for pc that we put in APK metadata should be the package metadata, not contents of pc Version.

Whilst it feels like the soname, it really isn't.

kaniini:
Yeah I agree, it was a mistake to do this in abuild, too.  I think we should push to just change it to $pkgver everywhere.

PR #1467, opened 2024-09-02
sca: make pc: provides/vendored use full package version

xnox:
.pc file version can be an aribitrary string value, without any
relation to soname, nor semver.

This yields the very opaque errors

    provides = pc:gdal=3.8.5dev
    datahash = c4638ec49b51aac1fc8f1d948b06a5b6c7a9f41dca2f980629b1f12f8e8be711
    ERROR: ./packages/x86_64/gdal-3.8-dev-3.8.5-r0.apk: package file format error

Which pointlessly waste engineering time.

Node we do not generate depends on pc: provides, as currently
`wolfictl text` transitive dependencies checks are unable to resolve
these.

Fixes: https://github.com/chainguard-dev/melange/issues/1172

Note there is sense that this may have been a missfeature in abuilds
too.

### DIFF
--- a/pkg/sca/sca.go
+++ b/pkg/sca/sca.go
@@ -23,7 +23,6 @@ import (
 	"io"
 	"io/fs"
 	"path/filepath"
-	"regexp"
 	"slices"
 	"strings"
 	"unicode"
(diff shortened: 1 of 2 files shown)


jonjohnsonjr (approved):
Thank you for being brave enough to do this. I don't think it's quite correct to always do this but it's probably more correct than what we were doing before.

PR #1472, opened 2024-09-03
feat(pipelines/split): Support overriding source package

EyeCantCU:
Currently, our split pipelines only support splitting targets from the main package in destdir. There are cases where this behavior could be useful for subpackages

I.E., minizip is built as part of the zlib packaging process as a subpackage. As a result, it includes development headers. As we can now override the source package, we can provide a minizip-dev subpackage without bundling it in zlib-dev

Example:

```
- name: minizip-dev
  pipeline:
    - uses: split/dev
      with:
        package: minizip
```

### DIFF
--- a/pkg/build/pipelines/split/debug.yaml
+++ b/pkg/build/pipelines/split/debug.yaml
@@ -15,7 +15,7 @@ inputs:
 pipeline:
   - runs: |
       PACKAGE_DIR="${{targets.destdir}}"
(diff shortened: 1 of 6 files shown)


jonjohnsonjr (approved):

EyeCantCU:
I need to clean this up, I'd rather it take the package name, and when we pass ./melange-out instead of /home/build/melange-out, it breaks everything

jdolitsky (approved):

PR #1481, opened 2024-09-08
pipelines/ruby: remove signing_key by default

Dentrax:
If it exist, remove the signing_key from gemspec by default since it leads to build failures as the key is not available in the build environment and is not needed.


mamccorm (approved):

developer-guy:
as its default: https://guides.rubygems.org/specification-reference/#signing_key

PR #1482, opened 2024-09-08
bail out on patch application failure

kouhaidev:

bail out on patch failure.

when building x86_64 images on apple silicon mac, melange may fail because of patch application failure. 

silent failure seems like a bad idea! if patch application fails, then the pipeline should then fail, notifying the melange user so that the build can get a proper fix.

```
2024/09/08 16:05:02 INFO running step "Apply patches"
2024/09/08 16:05:02 WARN qemu-x86_64-static: QEMU internal SIGSEGV {code=MAPERR, addr=0x20}
2024/09/08 16:05:02 WARN Segmentation fault (core dumped)
2024/09/08 16:05:02 INFO running step "Run autoconf make"
2024/09/08 16:05:02 INFO make: Entering directory '/home/build'
2024/09/08 16:05:02 INFO /usr/bin/make -C deps -f Makefile all
2024/09/08 16:05:02 WARN qemu-x86_64-static: QEMU internal SIGSEGV {code=MAPERR, addr=0x20}
2024/09/08 16:05:03 INFO make[1]: Entering directory '/home/build/deps'
2024/09/08 16:05:03 INFO /usr/bin/make -C redis -f Makefile all
```

### Functional Changes

~~- [ ] This change can build all of Wolfi without errors (describe results in notes)~~

Notes:

### SCA Changes

~~- [ ] Examining several representative APKs show no regression / the desired effect (details in notes)~~

Notes:

### Linter

~~- [ ] The new check is clean across Wolfi~~
~~- [ ] The new check is opt-in or a warning~~

Notes:

### DIFF
--- a/pkg/build/pipelines/patch.yaml
+++ b/pkg/build/pipelines/patch.yaml
@@ -33,5 +33,5 @@ pipeline:
       fi
 
       grep -v -E '^(#|$)' $series | (while read patchfile; do
-        patch '-p${{inputs.strip-components}}' < $patchfile
+        patch '-p${{inputs.strip-components}}' < $patchfile || exit 1
       done)


lf-:
O_o there's not set -e in the shell setup in the first place? nonetheless this should be fixed tbh

xnox:
@smoser should we generate list of patches to apply, and then apply them with the for loop? or should we `set -e` in the subshell? is there a way to always enforce `set -e` in the subshells? not sure why we do things like this.

smoser:
This is fixed now through https://github.com/chainguard-dev/melange/pull/2423 .
Sorry for the delinquency.

PR #1483, opened 2024-09-09
pipelines/ruby: add new pipelines to ruby packages

Dentrax:
This PR adds new pipelines for Ruby:
* `ruby/test`
* `ruby/require`

Example usages:

```yaml
- uses: ruby/test
  with:
    command: ruby -e "require 'json'"

- uses: ruby/require
  with:
    require: json

- uses: ruby/require
  with:
    requires: |
      require 'json'
```


pkg/build/pipelines/ruby/test.yaml
```diff
@@ -0,0 +1,14 @@
+name: Test a Ruby package
```
EyeCantCU, line 1: I think we should get rid of this. We're not doing anything inherently specific to Ruby

pkg/build/pipelines/ruby/require.yaml
```diff
@@ -0,0 +1,85 @@
+name: Test a Ruby package require, with optional load clause
+
+needs:
+  packages:
+    - wolfi-base
```
EyeCantCU, line 5: Wolfi base is implicit. We don't need to define it

pkg/build/pipelines/ruby/require.yaml
```diff
@@ -0,0 +1,85 @@
+name: Test a Ruby package require, with optional load clause
+
+needs:
+  packages:
```
EyeCantCU, line 4: We should add Ruby here, passed as an input. See go/build for what I'm talking about

pkg/build/pipelines/ruby/require.yaml
```diff
@@ -0,0 +1,85 @@
...
+        # test that otherthing can be required from asdf
+        require 'asdf'
+        require 'bark' # this is like woof
+
+      full-line and inline comments are supported via '#'
+    required: false
+
+pipeline:
+  - runs: |
+      set +x
```
EyeCantCU, line 28: I don't think this applies to the rest of the pipeline

EyeCantCU (changes requested):
This is great. Left a few comments. Can we also document the usage of the pipeline in melange's docs?

pkg/build/pipelines/ruby/require.yaml
Dentrax: I just checked the other Ruby pipelines and it seems we didn't include the `Ruby`, was wondering if there was any reason. 🤔

pkg/build/pipelines/ruby/test.yaml
Dentrax: It seems we have `name:` field defined on each pipeline, do you say we should omit this field?

pkg/build/pipelines/ruby/require.yaml
Dentrax: This is a copy-paste thing from `python/import` pipeline actually. I'm checking the CI logs [here](https://github.com/chainguard-dev/melange/actions/runs/10778428673/job/29889753721?pr=1483#step:9:1075) but couldn't see the instructions either.

pkg/build/pipelines/ruby/test.yaml
EyeCantCU: Sorry, the pipeline. It looks like it does the same thing a `runs` would do

pkg/build/pipelines/ruby/require.yaml
EyeCantCU: I don't think it'd ever been considered that pipeline should fetch the toolchain. `go/build` didn't offer this functionality until a few months ago either. We should provide because it's helpful, and we can implicitly pull in ruby without defining it in the environment

pkg/build/pipelines/ruby/require.yaml
EyeCantCU: Yeah, happy to be wrong here, but I don't believe it's transient to the rest of the pipeline. Will let someone that knows for sure chime in

pkg/build/pipelines/ruby/require.yaml
Dentrax: That makes sense, I've added Ruby here and couldn't pass the CI error: https://github.com/chainguard-dev/melange/actions/runs/10788637898/job/29919880644?pr=1483#step:9:1071 - for some reason, it doesn't find `ruby-3.2` even its there.

pkg/build/pipelines/ruby/require.yaml
smoser: it does "apply", in that it turns off '-x' option (making it not spew bash debug to stderr).

smoser (changes requested):
with regard to the error you're seeing [here]:
```
 2024/09/10 09:27:54 INFO running step "Test a Ruby package require, with optional load clause"
2024/09/10 09:27:54 WARN /bin/sh: ruby-3.2: not found
2024/09/10 09:27:54 INFO ruby-3.2 -e "require 'json'": FAIL
2024/09/10 09:27:54 INFO ERROR: failed to test package. the test environment has been preserved:
2024/09/10 09:27:54 INFO   workspace dir: /tmp/melange-workspace-1987892197
2024/09/10 09:27:54 INFO   guest dir: /tmp/melange-guest-1360058723
```

A couple things:
1. ruby-3.2 does not provide an executable named 'ruby-3.2'.  The package is
named ruby-3.2 but it only provides /usr/bin/ruby.  So thats why you get
'ruby-3.2' command not found.

2. If the 'inputs.ruby' is providing the _package_ rather than an executable
name (in python it provides an executable name),  to install, _and_ we are assuming that a ruby (ruby-3.X) package will always install /usr/bin/ruby, then you can ditch the searching for a binary (lines 43-57 entirely), and make 'RUBY='ruby' on line 29.

3. in e2e-tests/ruby3.2-json-test.yaml you should actually drop the 'packages:' section, because the 'needs' should fulfill that.

Dentrax:
Thank you so much @smoser! I just updated the PR and hit a new issue:

```
ruby-3.2-3.2.4-r1.apk disqualified because ruby-3.3-3.3.5-r1.apk already provides cmd:erb
```

That's probably due to I set `${{inputs.ruby}}=ruby` so it installs the `ruby-3.3` (latest) whereas I have `ruby-3.2` in the test pipeline. But I'd like to set `ruby-3.2` as a default package since most of Wolfi packages using this version. That's why I had to set `default: ruby-3.2`.

pkg/build/pipelines/ruby/require.yaml
smoser: So , yeah... several of the pipelines that I've added have this flaw in that they don't declare `needs` that it'd be nice if they did. [`py/pip-build-install`](https://github.com/wolfi-dev/os/blob/main/pipelines/py/pip-build-install.yaml) for example needs bot a pip and a python, but doesn't declare either.

The reason is that it takes `python` as an input (the python version to use).

I don't think  you can use a input to a pipeline in its `needs`. It would be cleaner here if you could.  This specific case would be nice in that you could then:
```
needs:
  packages:
    - ruby-${{inputs.version}}
```

And then have a inputs that declared `3.2` as the default or something. 

If that happens to work, (including with a `range` section like we're doing in python) then I guess I'd suggest doing that, but other wise I'd not bother.

Another option would be to have a pipeline per ruby-version (ruby/require-3.2, ruby/require-3.3).  I suspect that would work, and you could reduce copy/paste code between ruby/require-3.2 and ruby/require-3.3 with nested pipelines.  But if you were doing per-ruby versions i'd want that in wolfi-dev/os rather than melange proper.

So...
that was all long winded way of saying "I think this is fine".

pkg/build/pipelines/ruby/require.yaml
```diff
@@ -0,0 +1,85 @@
...
+        command="$1"
+        if $RUBY -e "$command"; then
+          echo "$RUBY -e \"$command\": PASS"
+        else
+          echo "$RUBY -e \"$command\": FAIL"
+          return 1
+        fi
+      }
+
+      if [ "$RUBY" = "ruby" ]; then
```
smoser, line 43: I think the default 'DEFAULT' or some other non-usable string is prefereable here.
The caller could very legitmately have `ruby` in the PATH and want that to be tested.  At least I can't think of a good reason to "fix" it for them.

Did you have one?

Dentrax:
Thank you @smoser! I made some changes according to your reviews here, could you PTAL when possible? It seems the CI is now green but not sure whether it's false-positive.

pkg/build/pipelines/ruby/test.yaml
smoser: so drop this?

pkg/build/pipelines/ruby/require.yaml
```diff
@@ -0,0 +1,88 @@
...
+  packages:
+    - ruby-${{inputs.version}}
+
+inputs:
+  ruby:
+    description: Which Ruby to use
+    default: DEFAULT
+  version:
+    description: Which Ruby version to use
+    default: 3.2
```
smoser, line 13: maybe if you're letting it declare version, then don't let it declare 'ruby' ?
because what happens if i do:
```
ruby: ruby3.3
version: ruby3.2
```

I tihnk one or the other.
Either take `ruby` and just don't bother with `needs`.
or take version , set needs, and use either explicitly use `/usr/bin/ruby${{inputs.version}}` or just assume in PATH as `ruby${{inputs.version}}`.

thoughts?

smoser: I'm kind of thinking out loud here.  I think its wierd to take both, but it was _very_ useful to in the python ones to use `DEFAULT` and pick the right /usr/bin/python3.XX when there was only one, and error in other cases.

smoser (changes requested):

Dentrax:
Thank you so much for the reviews @smoser! I have pushed a few commits, could you please take to a look again to see whether your concerns clarified here?

PR #1495, opened 2024-09-16
cleanup: remove some direct imports of charm log

imjasonh:
These seem to be relatively common cases where someone writes `log.Warn` and goimports chooses to import charmbracelet/log. This shouldn't be necessary when you call `log := clog.FromContext(ctx)` first.

For CLI users the result is not really any different (we use charm logging via clog anyway), but for places where this code might get embedded in another service where a different logger is used, it can be surprising.


89luca89 (approved):
lgtm!

Not sure what is going on with the CI but seems more like a gnu.org problem?

PR #1496, opened 2024-09-16
keygen: reject bit size < 2048

imjasonh:


xnox (changes requested):
That's way too much beyond what all standards bodies request and recommend.

2048 should be the current minimum for RSA.

xnox (approved):

PR #1522, opened 2024-09-27
Make installing -dev package "just work" and pull in all other dev libraries it needs

xnox:
Developer impact:

I want to compile and link against curl, i installed curl-dev and nothing works, i even installed openssl-dev and everything still does not work:

```
diff --git a/scans/curl.yaml.scan b/scans/curl.yaml.scan
index 98c235989..5260fd010 100644
--- a/scans/curl.yaml.scan
+++ b/scans/curl.yaml.scan
@@ -13,6 +13,11 @@ depend = brotli-dev
 depend = libpsl-dev
 depend = nghttp2-dev
 depend = openssl-dev
+depend = pc:libbrotlidec
+depend = pc:libnghttp2
+depend = pc:libpsl
+depend = pc:openssl
+depend = pc:zlib
 depend = so:libcurl.so.4
 provides = cmd:curl-config=8.10.1-r1
 provides = pc:libcurl=8.10.1-r1
```

With this change curl-dev automatically gains depends on openssl zlib libpsl libnghttp2 libbrotlidec dev packages to ensure that building and linking against curl-dev "just works".

- **generateRuntimePkgConfigDeps: only do so for public .pc, not vendored**
  The existing neom test case of vendored .pc files verifies this, and
  fails, when this feature is turned on.
  
  The diff is ugly, but it moves and indents the `if
  generateRuntimePkgConfigDeps{}` to be nested in the `if isInDir(){}`
  first condition for public files.
  

- **Enable pc file dependencies**
  Downloaded and extracted 1,030 .pc files from wolfi (base on 24th
  September file listing of all packages from @dustin).
  
  Executing `pkgconf --exist` on them revieled that 57 of them declare
  dependencies that do not, in fact exist.
  
  It is safe to turn pc: depends, as for vast majority of packages they
  will work correctly.
  
  For the 57 broken ones, there is now test/pkgconf pipeline, and one
  genuinely has to fix those. By either packaging libraries that are
  declared as required, ensuring those dependant libraries ship
  pkgconfig file, or patch/remove said broken pkgconfig file.
  
  Out of the sample of 1,030 .pc files, the 788 of them do have
  requirements declared. Thus they will need rebuilds to gain the
  correct dependencies. These however are in just 469 binary packages,
  some of which come from the same origin. Thus roughly at most 400 or
  so origins to rebuild.
  
  Furthermore wolfi pre-submit, install apk check will catch the broken
  -dev packages that cannot be installed.
  
  With this feature enabled, build log has:
  
  ```
  2024/09/27 23:44:24 INFO scanning for pkg-config data...
  2024/09/27 23:44:24 INFO   found pkg-config libarchive for usr/lib/pkgconfig/libarchive.pc
  ...
  2024/09/27 23:44:24 INFO   found pkg-config dependency (requires private) libcrypto for usr/lib/pkgconfig/libarchive.pc
  ...
  2024/09/27 23:44:24 INFO   runtime:
  2024/09/27 23:44:24 INFO     bzip2-dev
  2024/09/27 23:44:24 INFO     lz4-dev
  ...
  2024/09/27 23:44:24 INFO     pc:libcrypto
  ...
  2024/09/27 23:44:24 INFO     so:libarchive.so.13
  2024/09/27 23:44:24 INFO     xz-dev
  2024/09/27 23:44:24 INFO     zlib-dev
  2024/09/27 23:44:24 INFO     zstd-dev
  2024/09/27 23:44:24 INFO   provides:
  2024/09/27 23:44:24 INFO     pc:libarchive=3.7.6-r0
  ```
  
  These extra new dependency generated `pc:libcrypto` which upon
  installing libarchive-dev allows one to use it straight away for both
  dynamic and static linking.
  
  Note all of our public pc files already generated provides:
  
  ```
  $ apk info --provides openssl-dev
  openssl-dev-3.3.2-r0 provides:
  pc:libcrypto=3.3.2
  pc:libssl=3.3.2
  pc:openssl=3.3.2
  ```

### DIFF
--- a/pkg/sca/sca.go
+++ b/pkg/sca/sca.go
@@ -441,28 +441,28 @@ func generatePkgConfigDeps(ctx context.Context, hdl SCAHandle, generated *config
 		if isInDir(path, []string{"usr/local/lib/pkgconfig/", "usr/local/share/pkgconfig/", "usr/lib/pkgconfig/", "usr/lib64/pkgconfig/", "usr/share/pkgconfig/"}) {
 			log.Infof("  found pkg-config %s for %s", pcName, path)
 			generated.Provides = append(generated.Provides, fmt.Sprintf("pc:%s=%s", pcName, hdl.Version()))
+
+			if generateRuntimePkgConfigDeps {
+				// TODO(kaniini): Capture version relationships here too.  In practice, this does not matter
+				// so much though for us.
+				for _, dep := range pkg.Requires {
+					log.Infof("  found pkg-config dependency (requires) %s for %s", dep.Identifier, path)
+					generated.Runtime = append(generated.Runtime, fmt.Sprintf("pc:%s", dep.Identifier))
+				}
+
+				for _, dep := range pkg.RequiresPrivate {
+					log.Infof("  found pkg-config dependency (requires private) %s for %s", dep.Identifier, path)
+					generated.Runtime = append(generated.Runtime, fmt.Sprintf("pc:%s", dep.Identifier))
+				}
+
+				for _, dep := range pkg.RequiresInternal {
+					log.Infof("  found pkg-config dependency (requires internal) %s for %s", dep.Identifier, path)
+					generated.Runtime = append(generated.Runtime, fmt.Sprintf("pc:%s", dep.Identifier))
+				}
+			}
 		} else {
 			log.Infof("  found vendored pkg-config %s for %s", pcName, path)
 			generated.Vendored = append(generated.Vendored, fmt.Sprintf("pc:%s=%s", pcName, hdl.Version()))
-		}
-
-		if generateRuntimePkgConfigDeps {
-			// TODO(kaniini): Capture version relationships here too.  In practice, this does not matter
-			// so much though for us.
-			for _, dep := range pkg.Requires {
-				log.Infof("  found pkg-config dependency (requires) %s for %s", dep.Identifier, path)
-				generated.Runtime = append(generated.Runtime, fmt.Sprintf("pc:%s", dep.Identifier))
-			}
-
-			for _, dep := range pkg.RequiresPrivate {
-				log.Infof("  found pkg-config dependency (requires private) %s for %s", dep.Identifier, path)
-				generated.Runtime = append(generated.Runtime, fmt.Sprintf("pc:%s", dep.Identifier))
-			}
-
-			for _, dep := range pkg.RequiresInternal {
-				log.Infof("  found pkg-config dependency (requires internal) %s for %s", dep.Identifier, path)
-				generated.Runtime = append(generated.Runtime, fmt.Sprintf("pc:%s", dep.Identifier))
-			}
 		}
 
 		return nil
--- a/pkg/sca/sca.go
+++ b/pkg/sca/sca.go
@@ -380,8 +380,10 @@ func generateSharedObjectNameDeps(ctx context.Context, hdl SCAHandle, generated 
 	return nil
 }
 
-// TODO(kaniini): Turn this feature on once enough of Wolfi is built with provider data.
-var generateRuntimePkgConfigDeps = false
+// TODO(xnox): Note remove this feature flag, once successful
+// note this can generate depends on pc: files that do not exist in
+// wolfi, however package install tests will catch that in presubmit
+var generateRuntimePkgConfigDeps = true
 
 // generatePkgConfigDeps generates a list of provided pkg-config package names and versions,
 // as well as dependency relationships.


mattmoor:
Very cool.

A drive-by comment: historically every time we touch SCA we've broken....  something 😅   Now that we have world builds, were you able to use that to help/do any of the validation here?

I definitely appreciate the thoroughness outlined in your PR description, so thank you for that! 🙏

xnox:
> Very cool.
> 
> A drive-by comment: historically every time we touch SCA we've broken.... something 😅 Now that we have world builds, were you able to use that to help/do any of the validation here?
> 
> I definitely appreciate the thoroughness outlined in your PR description, so thank you for that! 🙏

for that I need to substitute a different sdk image / wolfictl /melage used during the build. So far I have used world rebuilds with custom packages / custom extra package repositories. But not with customized wolfictl / melange.

xnox:
@mattmoor so we don't need to rebuild the packages to see what updated depends/provides they will have.

As long as one has every apk access, one can run `melange scan foo.yaml` and as long as all packages are in packages/arch/foo-version-repoch.apk one can get updated PKGINFO as generated by an old and new melange.

This opens up a trove of information, list of packages that would gain `depends = pc:foo` and ability to check if those foo files exist. majority are fine, but here is the finite list of things that would literary break installability of dev packages:

```
$ diff -u provides-no-version.list existing.list  | grep '^-'
--- provides-no-version.list	2024-10-08 00:35:32.070613079 +0100
-provides = pc:compositeproto
-provides = pc:damageproto
-provides = pc:dri2proto
-provides = pc:dri3proto
-provides = pc:fixesproto
-provides = pc:glproto
-provides = pc:inputproto
-provides = pc:kbproto
-provides = pc:lept
-provides = pc:libpng
-provides = pc:ncurses
-provides = pc:presentproto
-provides = pc:randrproto
-provides = pc:recordproto
-provides = pc:renderproto
-provides = pc:resourceproto
-provides = pc:scrnsaverproto
-provides = pc:shared-mime-info
-provides = pc:skalibs
-provides = pc:videoproto
-provides = pc:xextproto
-provides = pc:xf86driproto
-provides = pc:xf86vidmodeproto
-provides = pc:xineramaproto
```

this is a list of new generated dependencies on pkg-config files, that are declared as required and yet such pkg-config file is not shipped by any package in wolfi. It feels like it is an incomplete configuration, build or installation of portions of X11 / X.org stack of packages (all the proto ones). And some are surprising, as if our shared-mime-info is missbuilt.

Working with dustin on getting the above list fixed.

For those that are curious dustin pushed the output of new and old scans to => https://github.com/dustinkirkland/wolfi-os/tree/xnox/scans and @justinvreeland those old and new appear to indicate some interesting changes for `so:` stuff, sort of wish I updated melange and/or rerun those for the changes you are doing over at https://github.com/chainguard-dev/melange/pull/1545

(cc @smoser @jonjohnsonjr @imjasonh @joshrwolf @philroche i think you will like all of the above)

mattmoor:
👨‍🍳 💋

philroche:
@xnox Thanks for the ping. Can you add some details on what we are fixing and why?

You noted that there were misconfigured package definitions depending on package configs that do not exist or rather there is no package that provides that package config - is it this misconfiguration that you aim to fix?

xnox:
A full rescan shows that these dependencies will be added:

https://github.com/xnox/os/commit/1fad4996464cf384eeb3a9a93cda1730b3a2c93e

All of them exists and are available.

It is not entirely deterministic depends, because there are multiple providers of the same .pc files:

```
$ git grep -h provides scans/ | cut -d= -f1-2 | grep pc: | sort | uniq -c | grep -v '1 provides'
      2 provides = pc:egl
      2 provides = pc:fontsproto
      2 provides = pc:gdal
      2 provides = pc:gi-docgen
      2 provides = pc:gl
      2 provides = pc:glesv1_cm
      2 provides = pc:glesv2
      2 provides = pc:kadm-client
      2 provides = pc:kadm-server
      2 provides = pc:krb5
      2 provides = pc:krb5-gssapi
      2 provides = pc:libecpg
      2 provides = pc:libecpg_compat
      2 provides = pc:libisoburn-1
      2 provides = pc:libpgtypes
      2 provides = pc:libpq
      2 provides = pc:libtirpc
      2 provides = pc:libudev
      2 provides = pc:mimalloc
      4 provides = pc:pybind11
      2 provides = pc:simdjson
      2 provides = pc:SPIRV-Tools
      2 provides = pc:SPIRV-Tools-shared
      2 provides = pc:tbb
      2 provides = pc:udev
      2 provides = pc:xkbfile
      2 provides = pc:xproto
      2 provides = pc:xscrnsaver
```

Some are intentional (udev & eudev) but some may need a closer look in case we have naming clashes here. This may need upstream bug reports.

xnox:
Given analysis in https://github.com/chainguard-dev/melange/pull/1522#issuecomment-2405614496 I believe this is ready to land, and is safe to do so, and it will simplify "apk add foo-dev" which will pull in all the right "bar-dev baz-dev" that are also required.

dustinkirkland:
That would be amazing :-)

dustinkirkland:
I support and agree with the idea, design, and end results.  I defer to @jonjohnsonjr and @joshrwolf on the code review of sca.go.

kaniini:
This design is basically the same as we are doing in Alpine, so it's certainly a valid approach.  One thing to be aware of is that you probably want to also add virtual dependencies for CMake modules, since CMake decided to NIH the pkg-config concept...

pkg/sca/sca.go
```diff
@@ -380,8 +380,10 @@ func generateSharedObjectNameDeps(ctx context.Context, hdl SCAHandle, generated
 	return nil
 }
 
-// TODO(kaniini): Turn this feature on once enough of Wolfi is built with provider data.
-var generateRuntimePkgConfigDeps = false
+// TODO(xnox): Note remove this feature flag, once successful
+// note this can generate depends on pc: files that do not exist in
+// wolfi, however package install tests will catch that in presubmit
+var generateRuntimePkgConfigDeps = true
```
smoser, line 386: I'd just drop the flag.
If we need to revert it, we'll revert it and this flag will come back.

smoser (approved):
This LGTM.

xnox:
> This design is basically the same as we are doing in Alpine, so it's certainly a valid approach. One thing to be aware of is that you probably want to also add virtual dependencies for CMake modules, since CMake decided to NIH the pkg-config concept...

i was going to do cmake modules separately and later; we still have cmake bootstrap cycles and not enough things that can provide cmake modules actually do yet. Also not sure about which way the NIH - i somehow suspect Cmake modules were there first =) however CMake is a programming language, not just declarative files so it definitely is not the most straight forward one.

xnox:
@kaniini thank you for your code implementation, it was and still is absolutely correct and didn't need any changes, so happy to finally getting around to rolling this out.

pkg/sca/sca.go
xnox: ack.

xnox:
There is small fallout from this change, but which is also expected. if generated  "depend = pc:foo" has more than one package that provides such pc file, an explicit choice needs to be made as to which one it is desired for a given -dev package to depend on. So far it has been obvious, but it might be a case when any might want to be desired to be supported. In such cases we might need to suppress such dependency.

PR #1526, opened 2024-10-01
Properly detect .so files as deps

justinvreeland:
Fixes: https://github.com/wolfi-dev/os/issues/29828

Currently sca only adds files ending in '.so.' as valid deps. Some libraries, libgallium in mesa for example, don't have a numbered suffix and are being dropped by sca.

```
ldd usr/lib/libEGL.so.1.0.0 | grep gallium
	libgallium-24.2.3.so => not found
```

### Functional Changes

Notes:

This isn't a functional change as far as I'm ware. I didn't find issues building other packages.

### SCA Changes

Notes:

Previously mesa-egl was missing the dep on so:libgallium. With this change it is not. Other packages missing deps should now have them once this lands and they're rebuilt.

### DIFF
--- a/pkg/sca/sca.go
+++ b/pkg/sca/sca.go
@@ -310,7 +310,7 @@ func generateSharedObjectNameDeps(ctx context.Context, hdl SCAHandle, generated 
 			if lib == "libcuda.so.1" {
 				continue
 			}
-			if strings.Contains(lib, ".so.") {
+			if strings.Contains(lib, ".so") {
 				log.Infof("  found lib %s for %s", lib, path)
 				generated.Runtime = append(generated.Runtime, fmt.Sprintf("so:%s", lib))
 			}
--- a/pkg/sca/e2e_test.go
+++ b/pkg/sca/e2e_test.go
@@ -74,6 +74,7 @@ func TestAnalyze(t *testing.T) {
 				"so:liblcms2-e69eef39.so.2.0.16",
 				"so:liblzma-13fa198c.so.5.4.5",
 				"so:libm.so.6",
+				"so:libopenblas64_p-r0-0cf96a72.3.23.dev.so",
 				"so:libopenjp2-eca49203.so.2.5.0",
 				"so:libpng16-78d422d5.so.16.40.0",
 				"so:libpthread.so.0",
@@ -131,6 +132,8 @@ func TestAnalyze(t *testing.T) {
 				"so:libm.so.6",
 				"so:libmount.so.1",
 				"so:libssl.so.3",
+				"so:libsystemd-core-256.so",
+				"so:libsystemd-shared-256.so",
 				"so:libudev.so.1",
 			},
 			Provides: []string{


pkg/sca/sca.go
```diff
@@ -310,7 +310,7 @@ func generateSharedObjectNameDeps(ctx context.Context, hdl SCAHandle, generated
 			if lib == "libcuda.so.1" {
 				continue
 			}
-			if strings.Contains(lib, ".so.") {
+			if strings.Contains(lib, ".so") {
 				log.Infof("  found lib %s for %s", lib, path)
```
xnox, line 314: Need to check history, I do hope we do not have plugins depending on plugins which would be unresolved.

justinvreeland: Not sure what you mean here. If you can link against it some package should be providing the .so for it so this shouldn't cause problems there?   There's certainly a few existing packages missing deps.

xnox: two private .so, can link against each other, and are not marked as provides, but instead are vendored.

generating public depends on either, will be uninstallable.

justinvreeland: I don't think that clarifies things. Why would that be uninstallable? Can apk not resolve mutual dependencies? What mechanism is stopping the "private" so from being mentioned in the corresponding provides? Is this the check? Just that it's not suffixed with a number?

xnox: please install melange this this patch, and rebuild airflow, and check if it is still installable. And if its provides and depends make sense.

it ships these lovely things:

```
/opt/airflow/venv/lib/python3.12/site-packages/shapely/_geos.cpython-312-x86_64-linux-gnu.so:
	linux-vdso.so.1 (0x00007fffa7bdd000)
	libgeos_c-1e72bdae.so.1.17.4 => /opt/airflow/venv/lib/python3.12/site-packages/shapely/../shapely.libs/libgeos_c-1e72bdae.so.1.17.4 (0x0000724f5ac0d000)
	libpthread.so.0 => /lib/libpthread.so.0 (0x0000724f5ac07000)
	libc.so.6 => /lib/libc.so.6 (0x0000724f5aa12000)
	libgeos-e4f0e0c1.so.3.11.4 => /opt/airflow/venv/lib/python3.12/site-packages/shapely/../shapely.libs/libgeos-e4f0e0c1.so.3.11.4 (0x0000724f5a6ef000)
	libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x0000724f5a473000)
	libm.so.6 => /lib/libm.so.6 (0x0000724f5a38f000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x0000724f5a361000)
	/lib64/ld-linux-x86-64.so.2 (0x0000724f5ac7c000)
```

xnox: I agree htat shared library dependencies should be generate on libpthread, libc, libstdc++, libm nad libgcc_s. But not on libgeos_c-1e.... nor libgeos-e4f0.....

justinvreeland: Installs fine and I don't see any difference in the deps.

```
diff ./.PKGINFO  (curl -sL https://packages.wolfi.dev/os/x86_64/airflow-2.10.2-r0.apk | tar -Oxz .PKGINFO | psub)
--- ./.PKGINFO
+++ /home/user/tmp/.psub.XXXXglfsKC
@@ -2,11 +2,11 @@
 pkgname = airflow
 pkgver = 2.10.2-r0
 arch = x86_64
-size = 985378385
+size = 972759948
 origin = airflow
 pkgdesc = Platform to programmatically author, schedule, and monitor workflows
 url = 
-commit = 75a8e679c8411609e517525cc5c2337a72dd1745
+commit = 1720bc7fe9d8dd57c145c948bce0a9b93bc3614e
 builddate = 1726879781
 license = Apache-2.0
 depend = python3
@@ -51,4 +51,4 @@
 # vendored = so:libssl-3e69114b.so.1.1=1.1
 # vendored = so:libssl.so.3=3
 # vendored = so:libyajl-d141338e.so.2.1.0=2.1.0
-datahash = cc57f096c13ad62c1c2f5ed00221d5733977b5c37a39dfe342993f3a469f8b81
+datahash = 6c828b19ed510c45c591e80acff8194c6c2e0ecfcdae8ae50d197cb0ed351903
```


Mesa:

```
diff ./.PKGINFO  (curl -sL https://packages.wolfi.dev/os/x86_64/mesa-egl-24.2.3-r1.apk | tar -Oxz .PKGINFO | psub)
--- ./.PKGINFO
+++ /home/user/tmp/.psub.XXXXtxkLnD
@@ -2,18 +2,17 @@
 pkgname = mesa-egl
 pkgver = 24.2.3-r1
 arch = x86_64
-size = 443390
+size = 443605
 origin = mesa
 pkgdesc = mesa egl
 url = 
-commit = 75a8e679c8411609e517525cc5c2337a72dd1745
+commit = 711c1d76cacd50ae9505b2c1d18b7fbcaf4ef4d3
 builddate = 1727005161
 license = MIT AND SGI-B-2.0 AND BSL-1.0
 depend = so:libX11-xcb.so.1
 depend = so:libc.so.6
 depend = so:libdrm.so.2
 depend = so:libexpat.so.1
-depend = so:libgallium-24.2.3.so
 depend = so:libgbm.so.1
 depend = so:libglapi.so.0
 depend = so:libm.so.6
@@ -29,4 +28,4 @@
 depend = so:libxcb.so.1
 depend = so:libxshmfence.so.1
 provides = so:libEGL.so.1=1
-datahash = 3dee3bcc60aed1bc623c15518d1c99251e013772d46e427c7f2ac7ba6e5fe8ef
+datahash = 739adba2046c8367cea68acd0217e65de6a1344369bc5aa7c2d2211557d3f154
```

xnox: great

xnox (approved):
here be dragons; and hopefullly presubmit install test catch any edge cases;

cause the only thing i can think of has been accounted for.

xnox:
pandas test still fails. not sure if this is due to python3.13 churn; or if some update is needed.

justinvreeland:
> pandas test still fails. not sure if this is due to python3.13 churn; or if some update is needed.

I'm looking at it now locally I think it's just update churn.

xnox (approved):

PR #1531, opened 2024-10-02
fix py3-pandas-test

imjasonh:
This test currently fails with `/bin/sh: python3: not found`

Instead, we have `/usr/bin/python3.12` (someday `3.13`, etc.)


imjasonh:
Superceded by https://github.com/chainguard-dev/melange/pull/1532

PR #1537, opened 2024-10-07
[Draft] sca: check if runtime dependencies are vendored

justinvreeland:
This fixes and issue with 288fb15cf152df1de3ab94d5b4c4202acd005d7d that too broadly added dependencies on vendored libraries.

Libreoffice for example picked up a dep to an internal library without the corresponding provides.

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

### DIFF
--- a/pkg/sca/sca.go
+++ b/pkg/sca/sca.go
@@ -247,8 +247,9 @@ func generateSharedObjectNameDeps(ctx context.Context, hdl SCAHandle, generated 
 
 				for _, soname := range sonames {
 					log.Infof("  found soname %s for %s", soname, path)
-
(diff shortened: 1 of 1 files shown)


justinvreeland:
@xnox I'm going through some packages now. Do you have any specific recommendations to check?

xnox:
This is not that something one can guess.

The check like this is good, but should be allowed to be customizable.

Specifically one has to allow specifying file paths that are considered in the library path.

Such that those dependencies are ignored, and yet all others are resolved correctly.

I will pull up examples from other distros.

xnox:
Broadly need to deploy this, to stop breakages.

xnox (approved):

PR #1545, opened 2024-10-07
[Draft] Don't check if .so files are executable

justinvreeland:
### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:


xnox:
i like that you are scoping this to libraries only.

PR #1550, opened 2024-10-09
sca: revert melange changes for naked sca

justinvreeland:
These changes surface worse problems.


xnox (approved):
this gets us back to where we were

PR #1553, opened 2024-10-10
nit: correct errror messages under applySubstitutionsForPriorities

ajayk:


ajayk:
@jonjohnsonjr always time travels back and fixes things in the past https://github.com/chainguard-dev/melange/pull/1448

PR #1564, opened 2024-10-15
update config: add version filter prefix and contains to release monitor config block so implementations can perform the same behaviour as git and github configs

rawlingsj:
NOTE:
This renames two exported interfaces so they are normalized naming
```
	GetFilterPrefix() string
	GetFilterContains() string
```

This helps keep the existing UX consistent but allows the new release monitoring config to avoid mentioning `tag`s as that would be confusing.

Example of how things will look while keeping existing UX in YAML

Git - unchanged
```yaml
update:
  enabled: true
  git:
    tag-filter-prefix: v17.2
    tag-filter-contains: foo
```

GitHub - unchanged
```yaml
update:
  enabled: true
  git:
    tag-filter-prefix: v17.2
    tag-filter-contains: foo
```

Release monitor - changed: new config will use version-filter-prefix and not tag-filter-prefix.
```yaml
update:
  enabled: true 
  release-monitor:
    identifier: 38
    version-filter-prefix: v17.2
    version-filter-contains: foo
```

### Functional Changes

### DIFF
--- a/docs/UPDATE.md
+++ b/docs/UPDATE.md
@@ -30,6 +30,8 @@ update:
     identifier: 38 # Mandatory, ID number for release monitor
     strip-prefix: v # Optional, if the version obtained from the update service contains a prefix which should be ignored
     strip-suffix: ignore_me # Optional, if the version obtained from the update service contains a suffix which should be ignored
+    version-filter-prefix: v17.2 # Optional, filter to apply when searching versions with a prefix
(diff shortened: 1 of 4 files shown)


rawlingsj:
Thanks, gonna switch to draft as want to tinker with it a bit more, will regen the schema once happy.

pnasrat (approved):
LGTM

PR #1567, opened 2024-10-16
Do not add a cmd:awk dependency as nothing will ever provide cmd:awk

smoser:
Do not add a cmd:awk dependency as nothing will ever provide cmd:awk

awk is only provided by busybox or gawk.  Both of those do so with a
symlink.  That means that we have to ignore shbang that have awk or
we'll get a 'cmd:awk' dep that will never be provided.

Sad face.

Somehow we need to fix the symlink provider thing, but for
now, just this.

This was noticed in wolfi-dev/os/check.yaml . Melange said:

    Added shbang dep cmd:awk for usr/bin/checkmk

### DIFF
--- a/pkg/sca/sca.go
+++ b/pkg/sca/sca.go
@@ -608,7 +608,7 @@ func sonameLibver(soname string) string {
 func getShbang(fp io.Reader) (string, error) {
 	// python3 and sh are symlinks and generateCmdProviders currently only considers
 	// regular files. Since nothing will fulfill such a depend, do not generate one.
-	ignores := map[string]bool{"python3": true, "python": true, "sh": true}
+	ignores := map[string]bool{"python3": true, "python": true, "sh": true, "awk": true}
 
 	buf := make([]byte, 80)
 	blen, err := io.ReadFull(fp, buf)


imjasonh (approved):

dustinkirkland:
Could teach apk to understand and respect symlinks for executables in the $PATH?  I sure wish "apk add cmd:awk" worked as one would have expected...

jonjohnsonjr:
IIRC @xnox said that was _not_ desirable behavior.

dustinkirkland:
Why?  Because of the ambiguity of cmd:awk either being busybox or gawk?  Or something else?

xnox:
> Could we teach apk to understand and respect symlinks for executables in the $PATH?  I sure wish "apk add cmd:awk" worked as one would have expected...

We can, but they all must be generated without version, and then all of them will be coinstallable.

Or none should provide it.

It is ambiguous which thing should provide awk, and we may not prevent co-installing gnu-awk & busybox.

Such resolution is possible in APK & APKO, but we currently do not generate such resolutions.

I.e. symlinks need to be weak provides.

PR #1572, opened 2024-10-18
fix(bubblewrap_runner): run as build 1000 by default

### LINKED ISSUE #1570: Bubblewrap runner does not run as 1000 (build)
### Context

The Bubblewrap pipeline is run with same uid of the user that run `melange`, even though in differnt user namespaces. Only that uid is mapped inside the Bubblewrap user namespace, and not 1000 as standard and as already set in passwd via apko.

```
$ id -u
1001
$ melange build --runner=bubblewrap .. &
$ cat /proc/$(pgrep mycommand_in_bubblewrap)/uid_map
      1001       1001          1
```

Considering that my uid on the root user namespace is 1001, when I run `melange build --runner=bubblewrap` the pipeline runs commands as 1001 in a different user namespace but mapping 1001 of that user ns to 1001 of the host user namespace of the process that executed melange.

### Expected behaviour

To have the pipeline be run in a different user namespace, **but** with the id of the user that run `melange` in that process's user namespace, be mapped to a consistent uid in the Bubblewrap user namespace.

Indeed, the passwd entry is also created with name build and uid and gid 1000. 

### How to reproduce

Test pipeline:

```yml
$ cat <<EOF >test.yaml
package:
  name: test
  version: 0.1.0
  epoch: 0

environment:
  contents:
    packages:
      - build-base
      - busybox

pipeline:
  - runs: |
      echo "running as $(id)"
      echo "build user is $(id build)"
      echo "running sleep..."
      sleep infinity
      exit 1
EOF
```
```shell
$ melange build test.yaml --runner bubblewrap [...]

...
2024/10/18 19:14:58 INFO running as uid=1001 gid=1001 groups=65534(nobody),65534(nobody),1001
2024/10/18 19:14:58 INFO build user is uid=1000(build) gid=1000(build) groups=1000(build)
2024/10/18 19:14:58 INFO running sleep...
```

The running user in Bubblewrap is the very same user in the host considering the different user namespaces:

```shell
$ ls -l /proc/self/ns/user
lrwxrwxrwx. 1 massimiliano.giovagnoli massimiliano.giovagnoli 0 Oct 18 19:10 /proc/self/ns/user -> 'user:[4026531837]'
$ ls -l /proc/`pgrep -f "^sleep\ infinity$"`/ns/user
lrwxrwxrwx. 1 massimiliano.giovagnoli massimiliano.giovagnoli 0 Oct 18 19:20 /proc/855437/ns/user -> 'user:[4026533321]'
$ cat /proc/self/uid_map
         0          0 4294967295
$ cat /proc/`pgrep -f "^sleep\ infinity$`/uid_map
      1001       1001          1
$ ps -C sleep -o pid,cmd,user
    PID CMD                         USER
 855437 sleep infinity              massimiliano.giovagnoli # indeed
```

### Additional context

Moreover, if it's run with Docker runner, the behaviour is even different by default:

```shell
2024/10/18 19:37:40 INFO running as uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
2024/10/18 19:37:40 INFO build user is uid=1000(build) gid=1000(build) groups=1000(build)
2024/10/18 19:37:40 INFO running sleep infinity now...
```

```shell
$ ls -l /proc/`pgrep -f "^sleep\ infinity$"`/ns/user
lrwxrwxrwx. 1 root root 0 Oct 18 19:42 /proc/857591/ns/user -> 'user:[4026531837]'
$ ls -l /proc/self/ns/user
lrwxrwxrwx. 1 massimiliano.giovagnoli massimiliano.giovagnoli 0 Oct 18 19:45 /proc/self/ns/user -> 'user:[4026531837]'
```

Meaning that we can potentially have different results based on fileystem based on the runner.

Also, the CI runs Melange as `sudo melange` also with Bubblewrap, leading to always be uid 0 in the pipeline.

maxgio92:
Bubblewrap pipeline is run with same uid of the user that run melange by default, mapped for the child user namespace, leading to possible consistency problems on the resulting package filesystem, while standardizing uid to 1000 or 0.
The classic example is the difference between running melange with sudo and without.

Furthermore, passwd database is already set for the uid 1000 with name build, instead of the user that is run as, when not explicitely specified. The same applies to the gid map.

### Additional context

Fixes #1570

### Functional changes

### SCA Changes

Notes:

### DIFF
--- a/pkg/container/bubblewrap_runner.go
+++ b/pkg/container/bubblewrap_runner.go
@@ -34,7 +34,10 @@ import (
 
 var _ Debugger = (*bubblewrap)(nil)
 
-const BubblewrapName = "bubblewrap"
+const (
+	BubblewrapName = "bubblewrap"
+	buildUserID    = "1000"
+)
 
 type bubblewrap struct {
 }
@@ -85,16 +88,18 @@ func (bw *bubblewrap) cmd(ctx context.Context, cfg *Config, debug bool, envOverr
 		"--chdir", runnerWorkdir,
 		"--clearenv")
 
-	// Set caps as for the current thread bounding capability set.
-	baseargs = append(baseargs, "--cap-add", "all")
-
-	baseargs = append(baseargs, "--unshare-user")
+	// If we need to run as an user, we run as that user.
 	if cfg.RunAs != "" {
+		baseargs = append(baseargs, "--unshare-user")
 		baseargs = append(baseargs, "--uid", cfg.RunAs)
 		baseargs = append(baseargs, "--gid", cfg.RunAs)
-	} else {
-		baseargs = append(baseargs, "--uid", "0")
-		baseargs = append(baseargs, "--gid", "0")
+		// Else if we're not using melange as root, we force the use of the
+		// Apko build user. This avoids problems on machines where default
+		// regular user is NOT 1000.
+	} else if os.Getuid() > 0 {
+		baseargs = append(baseargs, "--unshare-user")
+		baseargs = append(baseargs, "--uid", buildUserID)
+		baseargs = append(baseargs, "--gid", buildUserID)
 	}
 
 	if !debug {
--- a/pkg/container/bubblewrap_runner.go
+++ b/pkg/container/bubblewrap_runner.go
@@ -85,6 +85,9 @@ func (bw *bubblewrap) cmd(ctx context.Context, cfg *Config, debug bool, envOverr
 		"--chdir", runnerWorkdir,
 		"--clearenv")
 
+	// Set caps as for the current thread bounding capability set.
+	baseargs = append(baseargs, "--cap-add", "all")
+
 	baseargs = append(baseargs, "--unshare-user")
 	if cfg.RunAs != "" {
 		baseargs = append(baseargs, "--uid", cfg.RunAs)
--- a/pkg/container/bubblewrap_runner.go
+++ b/pkg/container/bubblewrap_runner.go
@@ -34,7 +34,11 @@ import (
 
 var _ Debugger = (*bubblewrap)(nil)
 
-const BubblewrapName = "bubblewrap"
+const (
+	BubblewrapName = "bubblewrap"
+	DefaultUID     = "1000"
+	DefaultGID     = "1000"
+)
 
 type bubblewrap struct {
 }
@@ -88,6 +92,11 @@ func (bw *bubblewrap) cmd(ctx context.Context, cfg *Config, debug bool, envOverr
 	if cfg.RunAs != "" {
 		baseargs = append(baseargs, "--unshare-user")
 		baseargs = append(baseargs, "--uid", cfg.RunAs)
+	} else {
+		// TODO: we may want to inherit the apko.ImageConfiguration.Accounts to match passwd.
+		baseargs = append(baseargs, "--unshare-user")
+		baseargs = append(baseargs, "--uid", DefaultUID)
+		baseargs = append(baseargs, "--gid", DefaultGID)
 	}
 
 	if !debug {
--- a/pkg/container/bubblewrap_runner.go
+++ b/pkg/container/bubblewrap_runner.go
@@ -34,11 +34,7 @@ import (
 
 var _ Debugger = (*bubblewrap)(nil)
 
-const (
-	BubblewrapName = "bubblewrap"
-	DefaultUID     = "1000"
-	DefaultGID     = "1000"
-)
+const BubblewrapName = "bubblewrap"
 
 type bubblewrap struct {
 }
@@ -89,14 +85,13 @@ func (bw *bubblewrap) cmd(ctx context.Context, cfg *Config, debug bool, envOverr
 		"--chdir", runnerWorkdir,
 		"--clearenv")
 
+	baseargs = append(baseargs, "--unshare-user")
 	if cfg.RunAs != "" {
-		baseargs = append(baseargs, "--unshare-user")
 		baseargs = append(baseargs, "--uid", cfg.RunAs)
+		baseargs = append(baseargs, "--gid", cfg.RunAs)
 	} else {
-		// TODO: we may want to inherit the apko.ImageConfiguration.Accounts to match passwd.
-		baseargs = append(baseargs, "--unshare-user")
-		baseargs = append(baseargs, "--uid", DefaultUID)
-		baseargs = append(baseargs, "--gid", DefaultGID)
+		baseargs = append(baseargs, "--uid", "0")
+		baseargs = append(baseargs, "--gid", "0")
 	}
 
 	if !debug {
--- a/pkg/container/bubblewrap_runner_test.go
+++ b/pkg/container/bubblewrap_runner_test.go
@@ -1,0 +1,56 @@
+// Copyright 2024 Chainguard, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package container
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/chainguard-dev/clog/slogtest"
+)
+
+func TestBubblewrapCmd(t *testing.T) {
+	tests := []struct {
+		name         string
+		config       *Config
+		expectedArgs string
+	}{
+		{
+			name:         "With default UID and GID",
+			config:       new(Config),
+			expectedArgs: fmt.Sprintf("--unshare-user --uid %s --gid %s", DefaultUID, DefaultGID),
+		},
+		{
+			name:         "With config RunAs",
+			config:       &Config{RunAs: "65536"},
+			expectedArgs: fmt.Sprintf("--unshare-user --uid %s", "65536"),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := slogtest.Context(t)
+			args := make([]string, 0)
+
+			cmd := new(bubblewrap).cmd(ctx, tt.config, false, nil, args...)
+			if cmd.Args == nil {
+				t.Fatalf("cmd.Args should not be nil")
+			}
+			if !strings.Contains(strings.Join(cmd.Args, " "), tt.expectedArgs) {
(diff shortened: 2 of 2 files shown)


maxgio92:
Package `perl-yaml-syck` requires uid 0 for file permissions. This only works while running `melange` with `sudo`.
This is one big difference from development environment and the CI I think.
We could have other packages that relies on the fact that in Bubblewrap the UID is mapped to the same of the user who runs `melange`.

pkg/container/bubblewrap_runner.go
```diff
@@ -88,6 +92,11 @@ func (bw *bubblewrap) cmd(ctx context.Context, cfg *Config, debug bool, envOverr
 	if cfg.RunAs != "" {
 		baseargs = append(baseargs, "--unshare-user")
 		baseargs = append(baseargs, "--uid", cfg.RunAs)
+	} else {
```
89luca89, line 95: We should do this only if `cfg.RunAs` is not `0` (root)
so if cfg.RunAs != "" and not zero, we use `--uid` and `--gid`, else we completely skip the `--uid` thing

check how I manage it here: https://github.com/chainguard-dev/melange/blob/3d4bb7ed71361e157eab839cb6d650067afdb032/pkg/container/qemu_runner.go#L118

89luca89 (changes requested):

pkg/container/bubblewrap_runner.go
maxgio92: So this would standardize the uid to 0 instead of 1000, in both cases rootless, gotcha.
It should remove the inconsistency of the result also, between local and CI that uses sudo with Bubblewrap.

maxgio92:
We still may need some capabilities that are not inherited across the child user namespace.
They were when no unshare was requested to bwrap, resulting in one level less in the user namespace hierarchy.

pkg/container/bubblewrap_runner.go
89luca89: So doing a little test, this should be:

```go
baseargs = append(baseargs, "--unshare-user")
if cfg.RunAs != "" {
		baseargs = append(baseargs, "--uid", cfg.RunAs)
		baseargs = append(baseargs, "--gid", cfg.RunAs)
} else {
		baseargs = append(baseargs, "--uid", "0")
		baseargs = append(baseargs, "--gid", "0")
}
```

The inconsistency was that we were skipping the `--unshare-user` when runas is NOT specified
This would fallback to whatever user you are on the host, so in your example 1001

To have consistency we should respect runas when set, and map use the root user in the sandbox if not specified

maxgio92: Addressed

pkg/container/bubblewrap_runner.go
```diff
@@ -85,9 +85,13 @@ func (bw *bubblewrap) cmd(ctx context.Context, cfg *Config, debug bool, envOverr
 		"--chdir", runnerWorkdir,
```
89luca89, line 85: here we can add `--cap-add` to address the problems
Ideally we should use the minimum used by docker, so we should add here:

```go
"--cap-add", "CAP_CHOWN",
"--cap-add", "CAP_DAC_OVERRIDE",
"--cap-add", "CAP_FSETID",
"--cap-add", "CAP_FOWNER",
"--cap-add", "CAP_MKNOD",
"--cap-add", "CAP_NET_RAW",
"--cap-add", "CAP_SETGID",
"--cap-add", "CAP_SETUID",
"--cap-add", "CAP_SETFCAP",
"--cap-add", "CAP_SETPCAP",
"--cap-add", "CAP_NET_BIND_SERVICE",
"--cap-add", "CAP_SYS_CHROOT",
"--cap-add", "CAP_KILL",
"--cap-add", "CAP_AUDIT_WRITE",
```

89luca89:
> We still may need some capabilities that are not inherited across the child user namespace. For instance, the `CAP_DAC_OVERRIDE` for the package `perl-yaml-syck` is needed in the strip final step.
> 
> Nevertheless, I think we should understand why we need them and strive instead to run as rootless.

Addressed this in one of the requested changes :+1:

pkg/container/bubblewrap_runner.go
89luca89: Reference https://github.com/moby/moby/blob/master/oci/caps/defaults.go#L6-L19

89luca89 (changes requested):

maxgio92:
Now we miss capability for melange stage 2, to `mount` ( `CAP_SYS_ADMIN` ? 🫠 )

maxgio92:
Falling back to all. The result:

```shell
$ grep Cap /proc/self/status
CapInh: 0000000000000000
CapPrm: 0000000000000000
CapEff: 0000000000000000
CapBnd: 000001ffffffffff
CapAmb: 0000000000000000
$ melange build test.yaml |& grep Cap
2024/10/19 18:43:04 WARN + grep Cap /proc/self/status
2024/10/19 18:43:04 INFO CapInh:    000001ffffffffff
2024/10/19 18:43:04 INFO CapPrm:    000001ffffffffff
2024/10/19 18:43:04 INFO CapEff:    000001ffffffffff
2024/10/19 18:43:04 INFO CapBnd:    000001ffffffffff
2024/10/19 18:43:04 INFO CapAmb:    000001ffffffffff
```

pkg/container/bubblewrap_runner.go
```diff
@@ -85,9 +85,16 @@ func (bw *bubblewrap) cmd(ctx context.Context, cfg *Config, debug bool, envOverr
...
 		"--clearenv")
 
+	// Set caps as for the current thread bounding capability set.
+	baseargs = append(baseargs, "--cap-add", "all")
+
```
xnox, line 90: I will always advocate for least privileges.

If most things build without these privileges, can we remove them from docker and qemu runners too?

maxgio92: I think it's doable. We'd need to adjust some pipeline, especially the melange stage 2 that tries to run mount. Cases where dac_override is needed I think we'd just need to set permissions properly on files, to avoid it.

xnox: nice, but likely lots of small issues to open and tackle one at a time.

pkg/container/bubblewrap_runner.go
```diff
@@ -85,9 +85,16 @@ func (bw *bubblewrap) cmd(ctx context.Context, cfg *Config, debug bool, envOverr
...
+	baseargs = append(baseargs, "--unshare-user")
 	if cfg.RunAs != "" {
-		baseargs = append(baseargs, "--unshare-user")
 		baseargs = append(baseargs, "--uid", cfg.RunAs)
+		baseargs = append(baseargs, "--gid", cfg.RunAs)
+	} else {
+		baseargs = append(baseargs, "--uid", "0")
+		baseargs = append(baseargs, "--gid", "0")
 	}
```
xnox, line 98: I am pondering about this. Many invocations of bubblewrap typically bind-mount read-only etc/passwd and etc/group such that things like this just work correctly.

Or better in the user-namespace create dedicated builder account to run.

Would that also fix many things?

maxgio92: From what I see it just creates the account for build with uid/gid 1000 in passwd/group, with the apko config - that's why I was expecting to always be 1000 in bwrap when not specifyung any runAs.

Do you have some reference to those bind mounts so I can give a check?

These options (--uid/--gid=0) just map the current user id of the user namespace of the process that runs melange to 0 inside de bwrap user namespace (rootless) - uid_maps/gid_maps in procfs.

---

Whether we want to continue with rootless and applying least privilege approach I think it's fundamental, even though we should avoid sudo that it's causing inconsistent results IMO. That's the main point of this PR

pkg/container/bubblewrap_runner.go
89luca89: @maxgio92 @xnox 
the problem here is that *before* the caps where *not* dropped (hence the boostrap was working)

Now that we fixed this, we should add the caps needed
What I would think here is that we use `--unshare-user` (this will drop caps) *only* is the original user is *not* root, this way we bypass the `sudo melange` not working

And we keep the drop + basic caps from here:  https://github.com/moby/moby/blob/master/oci/caps/defaults.go#L6-L19

maxgio92: @89luca89 I wouldn't add by default any capability. We can end up without any issue.

In order to remove the inconsistency between runs without sudo, I think we should also, when not specified by RunAs, keep a standard uid by default instead of mapping the original user id, whether is 1000 (build) or 0 with no caps.

89luca89: There are some minimal caps that we need, mainly for stuff like chmod/own and basic shell, remember that we have literal shell stuff to run

I'd say that minimal subset should work, it's a sensible default
If we want more security we should use other stuff not bwrap in the first place

89luca89: After some investigation, I see that if not specificed (cfg.RunAs) we default to NO CAPS, with the regular user.

So we need to fix the user mapping only if cfg.RunAs AND if we're not using `sudo melange` for some reasons.

So let's skip the CAPs, and just work around this problem:

```go
	// if we need to run as an user, we run as that user
	if cfg.RunAs != "" {
		baseargs = append(baseargs, "--unshare-user")
		baseargs = append(baseargs, "--uid", cfg.RunAs)
		baseargs = append(baseargs, "--gid", cfg.RunAs)
	// else if we're not using melange as root, we force the use of the
	// apko build user, this avoids problems on machines where default
	// regular user is NOT 1000
	} else if os.Getuid() > 0 {
		baseargs = append(baseargs, "--unshare-user")
		baseargs = append(baseargs, "--uid", "1000")
		baseargs = append(baseargs, "--gid", "1000")
	}
```

This should fix your inconsistent behaviour on non-1000 default users, but also don't change the default behaviour for present pipelines

maxgio92: Agree, so we split the sudo vs unprivileged running user inconsistency scenario from the inconsistency due to the current uid vs standard uid (i.e. 1000).

maxgio92:
PTAL @xnox @89luca89

89luca89 (dismissed):

pkg/container/bubblewrap_runner.go
```diff
@@ -34,7 +34,10 @@ import (
 
 var _ Debugger = (*bubblewrap)(nil)
 
-const BubblewrapName = "bubblewrap"
+const (
+	BubblewrapName = "bubblewrap"
+	BuildUserID    = "1000"
```
imjasonh, line 39: I wouldn't export this, since it's not used anywhere outside this package, and `container.BuildUserID` might be misleading to external readers of this package's API.

maxgio92: Good catch! Thanks @imjasonh. Addressed, PTAL

89luca89 (approved):

89luca89 (approved):

PR #1573, opened 2024-10-18
test: install exactly the package version to test

imjasonh:
### Problem

When testing package `foo` with `melange test foo.yaml`, it can be quite easy to accidentally test an unexpected package version.

If `foo-1.2.3-r4.apk` exists in the repo, but you're working on `-r5`, but haven't built it locally yet, `melange test foo.yaml` will request `foo` as a test dependency, which will be satisfied by the old `-r4`. Tests may pass meaninglessly, or fail mysteriously.

### Solution

When testing `foo.yaml`, request exactly the package-version represented by the config. If `foo.yaml` says `epoch: 5` that means we should test `-r5`. If it hasn't been built yet, the test will fail, but at least you won't unexpectedly be testing an old version.

This was achievable with sufficiently rigorous understanding of the system, and one could request the test-time dependency `${{package.name}}-${{package.full-version}}`, but folks shouldn't have to remember that.

### DIFF
--- a/pkg/build/compile.go
+++ b/pkg/build/compile.go
@@ -73,8 +73,8 @@ func (t *Test) Compile(ctx context.Context) error {
 
 		te := &cfg.Subpackages[i].Test.Environment.Contents
 
-		// Append the subpackage that we're testing to be installed.
-		te.Packages = append(te.Packages, sp.Name)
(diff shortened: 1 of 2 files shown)


imjasonh:
https://github.com/chainguard-dev/melange/pull/1518 beat me to it!!

PR #1586, opened 2024-10-22
only warn if git info is not passed or detected

imjasonh:
If git state can't be detected, the name and namespace will be `unknown/unknown`, and the commit will be empty.


imjasonh:
Superceded #1587

Issue #1590, opened 2024-10-23
[cargo]: cargo/build pipeline improvements


kranurag7:
This issue tracks improving our `cargo/build` pipeline to be more flexible and allow more wide usage of this pipeline across our package repositories.

We've certain limitations when we use cargo/build pipeline and below are the notes about the same.

1. As of now, the pipeline fails if we define a different modroot other than `.`

```yaml
  - name: build linkerd-network-validator
    uses: cargo/build
    with:
      modroot: ./validator
      output: linkerd-network-validator
      install-dir: lib/linkerd
```
The reason for failure here is that `cargo` stores the binary root of the tree under `target` directory. 
```bash
      cd "${{inputs.modroot}}"
```
In here, we change the directory to a different one and binary is stored at root.

After build, When we are invoking the `install` command in the pipeline it's not able to find the directory. 
```bash
        install -Dm755 "${OUTPUT_PATH}/${{inputs.output}}" "${INSTALL_PATH}/${{inputs.output}}"
```
The solution here is to use `CARGO_TARGET_DIR` environment variable but maybe there's also a simpler way to do this.

2. custom binary name 
unlike go `-o` flag, cargo doesn't have a similar flag but there's an experimental feature in cargo to specify custom binary name in there. 
https://doc.rust-lang.org/cargo/reference/unstable.html#different-binary-name

```bash
cargo new foo
cd foo
cargo run
cargo build
```
In above case, the binary name is always `foo`

If we want to change the binary name then we'll have to specify it something like this.
```toml
[[bin]]
name = "sammy"
path = "src/main.rs"
```
and then when we invoke `cargo build` or `cargo run` the binary name is `sammy`

ckoehler:
Are these fixes related to #1582 ?

kranurag7:
@ckoehler I didn't had a chance to look at the issue in detail, I quickly skimmed over it. 

but this fixes are in general for every package that relies on cargo/build. 

In wolfi-dev/os repository, you can do the following to test it out. 

```bash
make dev-container
make package/geckodriver
```

If you add your config in there and build it then you should be able to proceed ahead. In general, I always use the dev-container environment for building purpose. 
If you still want to build without docker then maybe this command is more accurate than a simple `melange build`

```bash
[...]

PR #1592, opened 2024-10-23
pass --target-dir with cargo/build pipeline

kranurag7:
this patch passes `--target-dir` flag with cargo at build time. 

This is required because `cargo/build` pipeline uses modroot and when we `cd` into a specific directory and invoke cargo build, the resulting binary still ends up in root of the repository at `GIT_ROOT_DIR/target/release/foo` 

After which when we use the following install command it fails to find the binary. 

```bash
OUTPUT_PATH="target/release"
install -Dm755 "${OUTPUT_PATH}/${{inputs.output}}" "${INSTALL_PATH}/${{inputs.output}}"
```

by passing `--target-dir` flag to target, we are creating the `target/release/foo` binary in the current working directory hence the install command will succeed. 

This change is looking good considering backwards compatibility with the use of this pipeline in existing package. 

### ⚠️ Alternative consideration
We change the existing `modroot` name to `manifest-path` and invoke all the commands from the root of the repository. 

we can keep using the word `modroot` as well but IMHO `manifest-path` closely aligns with cargo terminology. 
```bash
Manifest Options:
      --manifest-path <PATH>  Path to Cargo.toml
```
We then need to point this to `Cargo.toml` in different directory and the result will always be stored at root of the repository inside `target/release` or `target/debug` directory depending on the profile we are building for. 

To me, this looks like a more better solution than the proposed one but this will break one existing package because of name change https://github.com/wolfi-dev/os/blob/main/geckodriver.yaml 

_I proceeded with first option because the alternative consideration was breaking one package._ 

related: https://github.com/chainguard-dev/melange/issues/1590

### DIFF
--- a/pkg/build/pipelines/cargo/build.yaml
+++ b/pkg/build/pipelines/cargo/build.yaml
@@ -46,7 +46,7 @@ pipeline:
       cd "${{inputs.modroot}}"
 
       # Build and install package(s)
-      cargo auditable build ${{inputs.opts}}
+      cargo auditable build --target-dir target ${{inputs.opts}}
       if [[ ! -z "${{inputs.output}}" ]]; then
         install -Dm755 "${OUTPUT_PATH}/${{inputs.output}}" "${INSTALL_PATH}/${{inputs.output}}"
       else


pkg/build/pipelines/cargo/build.yaml
```diff
@@ -46,7 +46,7 @@ pipeline:
       cd "${{inputs.modroot}}"
 
       # Build and install package(s)
-      cargo auditable build "${{inputs.opts}}"
+      cargo auditable build --target-dir target "${{inputs.opts}}"
```
jdolitsky, line 49: does this need to remove the quotes too such as in https://github.com/chainguard-dev/melange/pull/1593/files ?

jdolitsky (approved):

Issue #1582, opened 2024-10-21
cargo pipeline error


ckoehler:
melange is giving an error even tho the build is successful. 

# Log

```
❯ melange build --arch aarch64 test.yaml
2024/10/21 15:46:45 INFO melange is building:
2024/10/21 15:46:45 INFO   configuration file: test.yaml
2024/10/21 15:46:45 INFO   workspace dir: /var/folders/pl/pj8rn0910g75pms3yx__nkt00000gn/T/melange-workspace-1968109458
2024/10/21 15:46:45 INFO evaluating pipelines for package requirements
2024/10/21 15:46:45 INFO --cache-dir ./melange-cache/ not a dir; skipping
2024/10/21 15:46:45 INFO populating workspace /var/folders/pl/pj8rn0910g75pms3yx__nkt00000gn/T/melange-workspace-1968109458 from .
2024/10/21 15:46:45 INFO building workspace in '/var/folders/pl/pj8rn0910g75pms3yx__nkt00000gn/T/melange-guest-4117346961' with apko
2024/10/21 15:46:46 INFO setting apk repositories: [https://packages.wolfi.dev/os]
2024/10/21 15:46:46 INFO image configuration:
2024/10/21 15:46:46 INFO   contents:
2024/10/21 15:46:46 INFO     build repositories: []
2024/10/21 15:46:46 INFO     runtime repositories: [https://packages.wolfi.dev/os]
2024/10/21 15:46:46 INFO     keyring:      [https://packages.wolfi.dev/os/wolfi-signing.rsa.pub]
2024/10/21 15:46:46 INFO     packages:     [build-base busybox cargo-auditable rust]
2024/10/21 15:46:46 INFO   accounts:
2024/10/21 15:46:46 INFO     runas:
2024/10/21 15:46:46 INFO     users:
2024/10/21 15:46:46 INFO       - uid=1000(build) gid=1000
2024/10/21 15:46:46 INFO     groups:
2024/10/21 15:46:46 INFO       - gid=1000(build) members=[build]
2024/10/21 15:46:46 INFO auth configured for: []
2024/10/21 15:46:47 INFO installing ca-certificates-bundle (20241010-r0)
2024/10/21 15:46:47 INFO installing wolfi-baselayout (20230201-r15)
2024/10/21 15:46:47 INFO installing glibc (2.40-r3)
2024/10/21 15:46:47 INFO installing libgcc (14.2.0-r3)
[...]

PR #1596, opened 2024-10-23
set output to required for cargo/build pipeline

kranurag7:
In some projects, we get a lot of outputs in `./target/release` directory and if we don't set output then it tries to install a lot directories which is not possible via `install` executable. 

- below is an example listing in `/target/release` directory. 
```bash
~ # ls target/release/
build        deps         examples     incremental  probe        probe.d      pub          pub.d
```

- logs from melange pipeline
```bash
2024/10/23 21:14:59 WARN install: omitting directory 'target/release/build' uses=cargo/build
2024/10/23 21:14:59 WARN install: omitting directory 'target/release/deps' uses=cargo/build
2024/10/23 21:14:59 WARN install: omitting directory 'target/release/examples' uses=cargo/build
2024/10/23 21:14:59 WARN install: omitting directory 'target/release/incremental' uses=cargo/build
```

Fixes https://github.com/chainguard-dev/melange/issues/1582

>[!NOTE]
> We still need to think about how can we support installing more than one binaries from `./target/release` directory but I'll open up another patch for that.

### DIFF
--- a/pkg/build/pipelines/cargo/build.yaml
+++ b/pkg/build/pipelines/cargo/build.yaml
@@ -9,6 +9,7 @@ needs:
 
 inputs:
   output:
+    required: true
     description: |
       Filename to use when writing the binary. The final install location inside
       the apk will be in prefix / install-dir / output


ckoehler:
`output` could be a list of files or globs, to install multiple things, including libraries.

ckoehler (approved):

PR #1599, opened 2024-10-24
chore: unexport build pkg methods

luhring:
These methods are used only in the build package.

Also delete unused functions.


luhring:
Intended as a small, incremental set of improvements to the code base. Feel free to raise any concerns, such as other tools importing this code and requiring access to these methods.

PR #1613, opened 2024-11-01
Add man-db for doc packages

murraybd:
Instead of having users install a -doc package only to discover that `man` is not installed let's add `man-db` as a runtime dependency of any -doc package.


dustinkirkland:
I'll leave it for @jonjohnsonjr and/or @xnox to review the SCA code.  But spiritually, I'm very aligned, and this makes perfectly reasonable sense to me.

xnox:
This assumes that humans are supposed to use manpages by hand.

In general, most places where we build manpages today - is because the build breaks, unless one is building/installing manpages.

We typically don't patch any functionality that is different to upstream builds; and even when we do we often do not backport the matching manpage updates.

I wonder if there is any worth in building our own manpages at all.

murraybd:
If we are packaging the latest versions of upstream software, I think it is also valuable if we package the latest version of the project's documentation. If one searches the internet for the add user man page there are a variety of results and its not clear to me which search result is associated with the upstream project. Additionally, its not always obvious which version of the software the manual applies to which could lead to user confusion.

dustinkirkland:
I think it is still worth building manpages.  But I think we should aggressively (and automatically) move documentation (including manpages and info pages and other forms of HTML docs) always and comprehensively into -docs packages, which are easy and obvious to find, if one needs them.  And those packages, when installed, yes, should drag in man-db.

jonjohnsonjr (approved):

jonjohnsonjr (approved):

PR #1627, opened 2024-11-08
fix(qemu): use proper machine type and cpu model based on host arch

89luca89:
Fix situation when crosscompiling would not work because of missing cpumodel and machine types

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -377,7 +377,6 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 		// and guest
 		if cfg.Arch.ToAPK() != apko_types.ParseArchitecture(runtime.GOARCH).ToAPK() {
 			// let's use a recent default for those
-			baseargs = append(baseargs, "-cpu", "max")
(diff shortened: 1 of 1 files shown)


pkg/container/qemu_runner.go
```diff
@@ -370,6 +373,19 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 		baseargs = append(baseargs, "-cpu", cfg.CPUModel)
 	} else {
 		baseargs = append(baseargs, "-cpu", "host")
+		// we cant use `-cpu host` when architecture does not match between host
+		// and guest
+		if cfg.Arch.ToAPK() != apko_types.ParseArchitecture(runtime.GOARCH).ToAPK() {
+			// let's use a recent default for those
+			baseargs = append(baseargs, "-cpu", "max")
```
mattmoor, line 380: I think this will end up being confusing, since the paths below max it 100% unused

mattmoor:
```suggestion
```

mattmoor (approved):

mattmoor (approved):

Issue #1508, opened 2024-09-19
Problems running on Ubuntu with apparmor / restricted unprivileged userns


smoser:
Running melange for the first time on a Ubuntu system will often fail.
There are a couple hangups.

1. need to install bubblewrap (`apt-get install bubblewrap`).  This actually gives a pretty good error:

   ```
   2024/09/19 10:33:22 WARN SOURCE_DATE_EPOCH is specified but empty, setting it to 1969-12-31 19:00:00 -0500 EST
   2024/09/19 10:33:22 WARN cannot use bubblewrap for containers: bwrap not found on $PATH arch=x86_64
   2024/09/19 10:33:22 ERRO unable to run containers using bubblewrap, specify --runner and one of [bubblewrap docker qemu]
   ```

2. Current ubuntu systems require application specific configs in order to use unprivileged user namespaces

    The failure path in wolfi-dev/os looks like this:
    ```
    $ make package/mypkg
    @SOURCE_DATE_EPOCH= /home/smoser/go/bin/melange build mypkg.yaml \
       --repository-append /home/smoser/src/wolfi/os/packages --keyring-append local-melange.rsa.pub\
       --signing-key local-melange.rsa --arch x86_64 --env-file build-x86_64.env \
      --namespace wolfi --generate-index false  --pipeline-dir ./pipelines/ \
      -k https://packages.wolfi.dev/os/wolfi-signing.rsa.pub \
      -r https://packages.wolfi.dev/os
    ...
    2024/09/19 10:33:58 INFO built image layer tarball as /tmp/apko-temp-1864610170/apko-x86_64.tar.gz
    2024/09/19 10:33:58 INFO using /tmp/apko-temp-1864610170/apko-x86_64.tar.gz for image layer
    2024/09/19 10:33:58 INFO ImgRef = /tmp/melange-guest-2642779980
    2024/09/19 10:33:58 WARN bwrap: setting up uid map: Permission denied
    2024/09/19 10:33:58 ERRO ERROR: failed to build package. the build environment has been preserved:
    2024/09/19 10:33:58 INFO   workspace dir: /tmp/melange-workspace-4104388352
    2024/09/19 10:33:58 INFO   guest dir: /tmp/melange-guest-1722197694
    2024/09/19 10:33:58 ERRO failed to build package: unable to start pod: exit status 1
    ```

The unprivileged user namespace path is a bit of a rathole. Here is some reading if you're interested:
[...]

PR #1628, opened 2024-11-08
Give a pointer to solution if bubblewrap runner cannot use user names…

smoser:
…pace.

Users of melange on ubuntu will have issues on their first attempt to use melange as a result of 24.04 changes to availability of user namespaces.

Before this change is applied, an attempt to 'make package/foo' in wolfi-dev/os would result in error messages like this:

    bwrap: setting up uid map: Permission denied
    ERROR: failed to build package. the build environment has been preserved:
      workspace dir: /tmp/melange-workspace-4104388352
      guest dir: /tmp/melange-guest-1722197694
    failed to build package: unable to start pod: exit status 1

It doesn't give the user much info on how to fix.
This change turns that failure to look like:

    ERRO failed to build package: unable to start pod: Unable to execute 'bwrap --unshare-user true'.
    Command failed with:
      bwrap: setting up uid map: Permission denied

    See https://github.com/chainguard-dev/melange/issues/1508 for fix

See https://github.com/chainguard-dev/melange/issues/1508


pkg/container/bubblewrap_runner.go
```diff
@@ -167,6 +189,10 @@ func (bw *bubblewrap) StartPod(ctx context.Context, cfg *Config) error {
 	ctx, span := otel.Tracer("melange").Start(ctx, "bubblewrap.StartPod")
 	defer span.End()
 
+	if err := bw.assertUnshareUser(ctx); err != nil {
```
jonjohnsonjr, line 192: Would it make sense to run this in `TestUsability` instead of StartPod?

smoser: yeah, that makes more sense.  I didn't know about testusability.
changed to that.

jonjohnsonjr (approved):

Issue #1254, opened 2024-06-05
docker runner fails if package has a subpackage


smoser:
This is from wolfi-dev/os at 88a385e2a4c462cbf1a527df15ae99fe70ae9d48
2 things will make this work:
1. not using --runner=docker
2. removing the subpackages in the pkg-test.yaml

```
$ cat pkg-test.yaml
package:
  name: pkg-test
  version: 1
  epoch: 1
  description: pkg test
  copyright:
    - license: MIT

environment:
  contents:
    packages:
      - build-base
      - busybox

pipeline:
  - runs: |
      mkdir -p ${{targets.contextdir}}/usr/bin/
      ln -s true ${{targets.contextdir}}/usr/bin/pkg-test

subpackages:
  - name: "${{package.name}}-sub1"
    description: "subpackage"
    pipeline:
      - runs: |
          mkdir -p ${{targets.contextdir}}/usr/bin/
          ln -s true ${{targets.contextdir}}/usr/bin/pkg-test-sub1

update:
  enabled: false

$ make MELANGE_EXTRA_OPTS="--runner=docker" package/pkg-test
SOURCE_DATE_EPOCH= /usr/bin/melange build ./pkg-test.yaml
   --repository-append /home/user/src/wolfi-os/packages
   --keyring-append local-melange.rsa.pub --signing-key local-melange.rsa
   --arch x86_64 --env-file build-x86_64.env
   --namespace wolfi --generate-index false  --pipeline-dir ./pipelines/
   --runner=docker -k https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
   -r https://packages.wolfi.dev/os --source-dir ./pkg-test 
   --log-policy builtin:stderr,packages/x86_64/buildlogs/pkg-test-1-r1.log
2024/06/05 02:25:46 WARN SOURCE_DATE_EPOCH is specified but empty, setting it to 1970-01-01 00:00:00 +0000 UTC
2024/06/05 02:25:46 INFO melange is building:
2024/06/05 02:25:46 INFO   configuration file: ./pkg-test.yaml
2024/06/05 02:25:46 INFO   workspace dir: /home/user/tmp/melange-workspace-1156798952
2024/06/05 02:25:46 INFO evaluating pipelines for package requirements
2024/06/05 02:25:46 INFO populating workspace /home/user/tmp/melange-workspace-1156798952 from ./pkg-test
[...]

PR #1646, opened 2024-11-14
fix: allow running docker runner as non-root user

wojciechka:
### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -164,6 +164,10 @@ func New(ctx context.Context, opts ...Option) (*Build, error) {
 		if err != nil {
 			return nil, fmt.Errorf("unable to create workspace dir: %w", err)
 		}
+		err = os.Chmod(tmpdir, 0o755)
+		if err != nil {
+			return nil, fmt.Errorf("unable to change permissions to workspace directory: %w", err)
+		}
 		b.WorkspaceDir = tmpdir
 	}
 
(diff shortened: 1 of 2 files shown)


maxgio92:
Related issue #1647

smoser:
Do you have a recreate on this ?
https://github.com/chainguard-dev/melange/pull/1649 landed and "fixed" a trivial case, that is described with recreate in https://github.com/chainguard-dev/melange/issues/1254 .

Also, can you clarify "allow running docker runner as non-root user"? On the workstation setups, I run as user 1000 and can build lots of packages with the docker runner.  the docker runner there does run privileged.

philroche:
I was having issues with building [emissary](https://github.com/wolfi-dev/os/blob/main/emissary.yaml) with docker runner due to error 

```
failed to build package: mkdir /tmp/melange-workspace-230936390/melange-out: permission denied
```

Using this proposed branch resolves this issue.

juburr:
I've noticed the opposite: in a normal (or dare I say lazy?) docker installation where docker is running as root, melange fails. The build process seems to write several directories into /tmp called melange-workspace-XXXXXXXX, all with 700 permissions, and some of them end up getting owned by root:root after docker runs. And then when melange runs lint towards the end of the build process it's unable to access those directories. The simple fix is to run melange with sudo. I don't love it though.

Issue #1647, opened 2024-11-14
Failing on workspace cleanup when running as root


maxgio92:
### Issue

When running as non-root, the workspace directory being owned by root can't be cleaned up when terminating the package and subpackage pipelines.

### How to reproduce

It can be reproduced running [this pipeline](https://github.com/wolfi-dev/os/blob/a62f32fe33d3a469a4041bc009b454e3ea35cfd1/openbao.yaml) with the Docker runner.

Log:

```
...
2024/11/14 18:41:24 INFO running step "go/build"
2024/11/14 18:42:08 INFO running pipeline for subpackage openbao-compat
2024/11/14 18:42:13 INFO pod 0147b793a3a4d7da5718efe02d283f08dd0b2fa2e0ea6d6ec42f79d2bf1cb639 terminated
2024/11/14 18:42:13 INFO deleting guest dir /tmp/melange-guest-3710148864
2024/11/14 18:42:13 INFO deleting workspace dir /tmp/melange-workspace-863606971
2024/11/14 18:42:13 INFO deleting image apko.local/cache:5125b9ef6e4177faa3839e5202911825e07e9584d9283856a1bc11c77c23563a
2024/11/14 18:42:13 INFO untagged apko.local/cache:5125b9ef6e4177faa3839e5202911825e07e9584d9283856a1bc11c77c23563a
2024/11/14 18:42:13 INFO deleted sha256:aacf77f979edf8c434a5b25d431f368f62599141dcba99a4d63bd9c612a76a20
2024/11/14 18:42:13 INFO deleted sha256:2c3054dd01157c4d8b7e76f0eb41a278b1d1eeff9a7e4bc13d21aa117ff361d8
2024/11/14 18:42:13 ERRO failed to build package: mkdir /tmp/melange-workspace-863606971/melange-out: permission denied
```

The workspace directory is created as root because of Docker during the bind mount of it as volume.
The pipeline is run as root, but the destination directory mkdir it seems is not.

### Additional info

It would be nice to reach a state where all the pipeline is run as an unprivileged user with proper permissions and ownership across all the filesystem as it happens for the Bubblewrap environment with the unshared user namespace.

89luca89:
To expand on this

This happens on systems like linux where melange is ran as your $USER, while dockerd is running as root, ans $USER is in docker group.

Probably something localized in the git step, that chowns as root (which is the user running inside docker)

wojciechka:
[...]

PR #1648, opened 2024-11-14
Clean up workspace dir from inside runner to avoid permission errors.

smoser:
When running privileged docker as an unprivileged user, the files that are created in the WorkspaceDir are created as root. Even if the build were run as non-root, they would not necessarily be the same ownership as the user invoking melange.

As a result, WorkspaceDir would not be able to be cleaned up and melange would just leave files there to later be cleaned up with a dangerous 'sudo rm -Rf' by the user.

The change here is to clean up the WorkspaceDir from _inside_ the container, where the uid is the same as the uid that created the files.

I believe this will waste IO and/or time on the qemu runner, where /home/build isn't actually bind'd in.  Later we could expose a CleanWorkspace from the runner that was a noop in qemu.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -58,6 +58,22 @@ import (
 
 const melangeOutputDirName = "melange-out"
 
+var shellEmptyDir = []string{
+	"sh", "-c",
+	`d="$1";
+[ $# -eq 1 ] || { echo "must provide dir. got $# args."; exit 1; }
+cd "$d" || { echo "failed cd '$d'"; exit 1; }
+set --
+for e in * .*; do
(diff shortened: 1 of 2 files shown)


89luca89:
I like this :+1:  related to #1647 we could move also the mkdir in the container this way?

This should fix the problem of subpackage building when using docker runner with an unprivileged user

> I believe this will waste IO and/or time on the qemu runner, where /home/build isn't actually bind'd in. Later we could expose a CleanWorkspace from the runner that was a noop in qemu.

Also nice yes

89luca89:
That's ok let's move the discussion for the docker+subpackage back to #1254 and #1647 and leave this for the cleanup fix :+1:

89luca89 (approved):

PR #1653, opened 2024-11-15
fix(releases): more reliable releases

luhring:
- Consolidate release workflows into one
- Continue to use a scheduled check for new changes
- Also allow manual triggering of release creation


.github/workflows/release.yaml
```diff
@@ -1,13 +1,13 @@
-name: Create Release
+name: Release
 
 on:
-  push:
-    tags:
-      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
+  schedule:
+    - cron: '0 0 * * *' # daily at 00:00
```
k4leung4, line 5: nit 1: comment which timezone this is

.github/workflows/release.yaml
```diff
@@ -22,69 +22,51 @@ jobs:
...
+          if [ -z "$(git tag --points-at HEAD)" ]; then
+            echo "Nothing points at HEAD, so we need a new tag+release."
+            echo "need_release=yes" >> $GITHUB_OUTPUT
+          else
+            echo "A tag already points to head, no need for a new tag+release."
+            echo "need_release=no" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Bump version and push tag
+        uses: mathieudutour/github-tag-action@a22cf08638b34d5badda920f9daf6e72c477b07b # v6.2
```
k4leung4, line 39: how hard would it be to do this ourselves? 
prefer not using other actions that we need to pass our GITHUB token to

cpanato: I can fix that, just need a time

k4leung4:
do we really need daily releases?
the reason i ask is that this repo is also consumed as a go dep
daily releases mean frequent dependabot PRs

what changes are we needing to release daily and would it make sense to refactor the repo into a library and cli such that downstream can depend on something a bit more stable?

.github/workflows/release.yaml
luhring: Cool — yeah could we come back to this? This isn't a net change from this PR, this PR just moves it to this file

luhring:
> do we really need daily releases? the reason i ask is that this repo is also consumed as a go dep daily releases mean frequent dependabot PRs
> 
> what changes are we needing to release daily and would it make sense to refactor the repo into a library and cli such that downstream can depend on something a bit more stable?

The short answer is that we do tend to make updates to melange fairly often (daily is not unusual) that we want to bring into other tools and services.

I'm not married to daily — but note that this isn't new as of this PR, this PR is just trying to get releases working like we meant for them to before.

.github/workflows/release.yaml
k4leung4: definitely not blocking

k4leung4:
> > do we really need daily releases? the reason i ask is that this repo is also consumed as a go dep daily releases mean frequent dependabot PRs
> > what changes are we needing to release daily and would it make sense to refactor the repo into a library and cli such that downstream can depend on something a bit more stable?
> 
> The short answer is that we do tend to make updates to melange fairly often (daily is not unusual) that we want to bring into other tools and services.
> 
> I'm not married to daily — but note that this isn't new as of this PR, this PR is just trying to get releases working like we meant for them to before.

agree this isnt new, but its a pain in mono trying to stay up to date with this dep, especially when there are breaking api changes. i just want to raise attention to the churn daily releases creates downstream

cpanato:
once a week may be fine to release an on-demand

imjasonh (approved):

PR #1657, opened 2024-11-18
release: check for both tag and release

### LINKED ISSUE #1655: Release workflow should check if there's a current release, not just a current tag
We recently made some improvements to our release workflows in #1653. But we should keep improving:

Right now, we decide "do we need to make a new release" based on the answer to "is there already a tag pointing to the tip of `main`?". This creates a case (that is happening as I write this) where there is a tag, but no release. Currently this is happening because of a vestigial failed release from before the cleanup, but it's not impossible for this to happen in the new world if something unexpected happens during a release step in the automation.

It makes more sense to base "do we need a new release" based on the answer instead to "is there already a **release** pointing to the tip of `main`?".

luhring:
This is a follow-on to #1653.

Closes #1655 (see this issue for more info!)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -22,29 +22,39 @@ jobs:
 
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
-      - name: Check if any changes since last tag
+      - name: Check if any changes since last release
         id: check
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           git fetch --tags
-          if [ -z "$(git tag --points-at HEAD)" ]; then
-            echo "Nothing points at HEAD, so we need a new tag+release."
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -76,3 +86,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           HOMEBREW_TAP_GITHUB_TOKEN: ${{ steps.octo-sts.outputs.token }}
+          TAG: ${{ steps.check.outputs.existing_tag || steps.create_tag.outputs.new_tag }}
```
jdolitsky, line 89: will this result in `TAG: ""` if existing_tag  is empty string?

luhring: I think so, but that can't happen given how this pipeline is set up currently

jdolitsky (approved):

PR #1659, opened 2024-11-19
fix(pipeline): do preliminary checks for checksum invalid chars

89luca89:
Fix https://github.com/chainguard-dev/internal-dev/issues/1571

We don't want to pass possible code execution to `expected-sha{256,512}`, the comparison is done in shell
this opens the possibility to inject values and/or commands.

SHA checksums only contains [0-9a-zA-Z] chars, so ensure we only have those. This will prevent any injection at that stage.

### DIFF
--- a/pkg/build/pipeline.go
+++ b/pkg/build/pipeline.go
@@ -145,10 +145,16 @@ func validateWith(data map[string]string, inputs map[string]config.Input) (map[s
 			if !matchValidShaChars(data[k]) {
 				return data, fmt.Errorf("checksum input %q for pipeline contains invalid characters", k)
 			}
+			if len(data[k]) != 64 {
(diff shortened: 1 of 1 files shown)


pkg/build/pipeline.go
```diff
@@ -140,6 +141,16 @@ func validateWith(data map[string]string, inputs map[string]config.Input) (map[s
 		if data[k] == "" {
 			data[k] = v.Default
 		}
+		if k == "expected-sha256" && data[k] != "" {
+			if !matchValidShaChars(data[k]) {
```
xnox, line 145: can you please assert length too? As these hashes have an expected length.

89luca89: Right! Thanks :+1:

89luca89: Done :+1:

imjasonh (approved):

PR #1666, opened 2024-11-22
feat: add project.srcdir variable

### LINKED ISSUE #1665: Add variable to provide project source directory
This variable would be helpful when trying to create paths relative to base / source directory when using a pipeline with `working-directory` - for example:

```yaml
subpackages:
  - name: subpackage
    pipeline:
      # create files inside ${{targets.subpkgdir}} from other sources
      - uses: patch
        working-directory: "${{targets.subpkgdir}}"
        with:
          patches: "update-files-in-subpkgdir.patch"
```

This fails as the `patch.yaml` logic is reading files from current directory.

Something like this would work assuming a variable would be created:

```yaml
subpackages:
  - name: subpackage
    pipeline:
      # create files inside ${{targets.subpkgdir}} from other sources
      - uses: patch
        working-directory: "${{targets.subpkgdir}}"
        with:
          patches: "${{package.srcdir}}/update-files-in-subpkgdir.patch"
```

wojciechka:

Resolves #1665 

### Functional Changes

Notes: This is a new variable that does not exist in any packages, but did not run a test yet.

### SCA Changes

Notes:

### Linter

Notes:


jonjohnsonjr:
Should update README with this variable.

PR #1675, opened 2024-12-02
Improve comment stripping errors

jonjohnsonjr:
Before, these looked like:

```
compiling main pipelines: compiling Pipeline[0]: stripping runs comments: 14:13: not a valid test operator: -m
```

Now, these look like:

```
compiling "hwloc" test pipelines: compiling Pipeline[0]: stripping runs comments: 14:13: not a valid test operator: -m:
> if [[ uname -m == 'x86_64']]; then
              ^
```

First big change is that we include the package name in the error. Without this, it was really challenging to find which package actually had a syntax error.

Second big change is that we use the nice structured error we get back from the shell parsing library to include the line that failed with a cursor pointing at the column that failed.

### DIFF
--- a/docs/md/melange_compile.md
+++ b/docs/md/melange_compile.md
@@ -43,10 +43,13 @@ melange compile [flags]
       --env-file string             file to use for preloaded environment variables
       --fail-on-lint-warning        turns linter warnings into failures
       --generate-index              whether to generate APKINDEX.tar.gz (default true)
+      --git-commit string           commit hash of the git repository containing the build config file (defaults to detecting HEAD)
+      --git-repo-url string         URL of the git repository containing the build config file (defaults to detecting from configured git remotes)
       --guest-dir string            directory used for the build environment guest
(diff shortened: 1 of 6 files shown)


k4leung4 (approved):

imjasonh (approved):

smoser:
best. change. ever.
```
2024/12/13 08:58:23 INFO evaluating pipelines for package requirements
2024/12/13 08:58:23 INFO ERROR: failed to test package. the test environment has been preserved:
2024/12/13 08:58:23 INFO   workspace dir: /home/user/tmp/melange-workspace-2064787666
2024/12/13 08:58:23 ERRO failed to test package: compiling python-as-wrapper.yaml tests: compiling subpackage "python-as-env" tests: compiling Pipeline[0]: stripping runs comments: 17:1: "for foo [in words]" must be followed by "do":
> for val in bogus-prog-name PYTHON_AS;
  ^
```

PR #1677, opened 2024-12-04
feat(volume mounts): add support for mounting volumes into the melange build context

dakaneye:
This change allows the specification of volume mounts in the format src:dest,src2:dest2 into the context of the melange build container.

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

This has general utility for Melange but the first use case will be for mounting google cloud credentials into the context of melange so that a pipeline could retrieve an artifact from GAR for rebuilt artifacts.

Fixes https://github.com/chainguard-dev/internal-dev/issues/7385

### DIFF
--- a/pkg/cli/build.go
+++ b/pkg/cli/build.go
@@ -81,6 +81,7 @@ func buildCmd() *cobra.Command {
 	var configFileGitCommit string
 	var configFileGitRepoURL string
 	var configFileLicense string
+	var volume []string
 
 	var traceFile string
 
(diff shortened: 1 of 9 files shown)


pkg/cli/build.go
```diff
@@ -263,6 +272,7 @@ func buildCmd() *cobra.Command {
 	cmd.Flags().StringVar(&configFileGitCommit, "git-commit", "", "commit hash of the git repository containing the build config file (defaults to detecting HEAD)")
 	cmd.Flags().StringVar(&configFileGitRepoURL, "git-repo-url", "", "URL of the git repository containing the build config file (defaults to detecting from configured git remotes)")
 	cmd.Flags().StringVar(&configFileLicense, "license", "NOASSERTION", "license to use for the build config file itself")
+	cmd.Flags().StringSliceVar(&volumeMounts, "volumes", []string{}, "bind mount a volume(s) into the container (e.g., /host:/container)")
```
jonjohnsonjr, line 275: I'd this `volume` and do `StringSliceVarP` so we can use `-v` (to mirror docker).

PR #1678, opened 2024-12-04
fix(qemu): fix MacOS crosscompile path

89luca89:
Properly support x86 emulation on MacOS


justinvreeland:
Tested on my Mac works now!

imjasonh (approved):

imjasonh:
We should probably invest in a smoke test of this on a Mac runner

PR #1680, opened 2024-12-07
convert/gem/ruby: Don't overwrite existing melange files

justinvreeland:


justinvreeland:
> Is this fixing a bug? If so, is this something we could add a test for?

I consider it a bug. If you run `melange convert gem X` in Wolfi it should not overwrite the existing files you already have.   This exists in [python already](https://github.com/chainguard-dev/melange/blob/e788406063b83b86bef24ce16665e82495fc77ff/pkg/convert/python/python.go#L255) though I didn't notice a matching test.

I suppose the test would just be to find a gem that has a dependency, writing an empty file to the current directory for that dependency and then running convert to make sure the empty file is not overwritten. Would this go in e2e_tests?

luhring:
>I suppose the test would just be to find a gem that has a dependency, writing an empty file to the current directory for that dependency and then running convert to make sure the empty file is not overwritten.

Yeah sounds good!

>Would this go in e2e_tests?

I think we could scope the test down to to just calling `findDependencies` (or its caller) using the Go testing framework. The "e2e" tests using bash to do larger scale operations, and I don't think we need all that here.

PR #1693, opened 2024-12-11
pipelines/split/dev - grab files in /usr/share/cmake

smoser:
The split/dev pipeline does not currently "grab" files in /usr/share/cmake.  Rather, it only considers those down /usr/lib/cmake.

I discovered this when xsimd version 13.1.0 moved files from /usr/lib/cmake to /usr/share/cmake.
The PR [1] to do that was helpful and described why.

Wolfi cmake (3.29.0-r3) does find the files in /usr/share/cmake.

The end result of this is that packages that used split/dev and shipped files in /usr/share/cmake will now have those files as part of the -dev package.

A grep of file lists for 'usr/share/cmake' shows:

    bash-completion-2.15.0-r0
    cli11-2.4.2-r2
    cmake-3.29.0-r3
    cmake-bootstrap-3.29.0-r2
    gtk-doc-1.34.0-r1
    libgeotiff-1.7.3-r2
    nlohmann-json-3.11.3-r2
    py3.10-pybind11-2.13.6-r2
    py3.11-pybind11-2.13.6-r2
    py3.12-pybind11-2.13.6-r2
    py3.13-pybind11-2.13.6-r2
    spirv-headers-1.3.261.1-r3
    tl-expected-1.1.0-r1
    vulkan-headers-1.4.303-r1
    wasi-sdk-24-r4
    xsimd-13.1.0-r0

[1] https://github.com/xtensor-stack/xsimd/commit/a507f83ead608fffc558b615b61b20f44f2f4672

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:


justinvreeland (approved):
Looks good to me but I don't think my review has any power here.

PR #1701, opened 2024-12-13
fix(qemu): improve logging, make connection  check less spammy

89luca89:
This improves a bit the logging, making it less noisy and adhere to other runners' output.
Also, improve a bit the connection check, which gets very spammy on less performant machines (or cross compiles) where IO timeout was set way too low.

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -449,6 +449,10 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 	retries := 6000
 	try := 0
 	for try <= retries {
+		if err := ctx.Err(); err != nil {
+				return fmt.Errorf("checking SSH server: %w", err)		
+		}
+		
 		try++
 		time.Sleep(time.Millisecond * 500)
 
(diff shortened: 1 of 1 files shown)


pkg/container/qemu_runner.go
```diff
@@ -450,14 +449,17 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 	retries := 6000
 	try := 0
 	for try <= retries {
+		try++
```
jonjohnsonjr, line 452:
```suggestion
		if err := ctx.Err(); err != nil {
				return fmt.Errorf("checking SSH server: %w", err)		
		}
		
		try++
```

To ensure ctrl+C works.

89luca89: Thanks :heart:

smoser (commented):
minor comment inline.

pkg/container/qemu_runner.go
```diff
@@ -620,32 +626,31 @@ func generateDiskFile(ctx context.Context, diskSize string) (string, error) {
...
 	if err != nil {
-		return err
+		return fmt.Errorf("conn read: %w", err)
 	}
 
 	// Check if the banner starts with "SSH-"
 	banner := string(buffer[:n])
-	if len(banner) >= 4 && banner[:4] == "SSH-" {
-		return err
+	if len(banner) >= 4 && strings.Contains(banner, "SSH-2.0-OpenSSH") {
```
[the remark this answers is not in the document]
89luca89: Yea right, it was done before because of the slice handling, but now it's done with contains :+1:

smoser (approved):

PR #1702, opened 2024-12-13
feat: add support to add and drop linux capabilities

### LINKED ISSUE #1703: Feature request: add support for adding linux capabilities
### Title

Make melange provide way to declare the Linux capabilities to be added to the build process.

### Short Description of Issue/Improvement Need

We have different packages in Wolfi that need to be run with additional capabilities on top of the ones set by the runners. For instance, there are cases where a `CAP_DAC_OVERRIDE` is needed in the build process, but without being explicated. So when trying running that unprivileged it would fail.

Having a possibility to specify the capabilities added in the Melange pipeline manifest, would make that explicit and help on both:
* security: allowing to apply least privilege
* efficiency and reproducibility: we wouldn't need to run pipelines with different user and capabilty sets to reproduce the same result, and in the end, also saving time during debugging.

Even with QEMU runner, this would help on reproducibility side.

### Impact of issue

There's no big impact now. We know the situation in advance and we circumvent it by using `sudo`. 

### Steps to Reproduce

I remember a package needed to override dac, unfortunately I don't remember which. `postfix` should be a valid example:
1. Set `environment.accounts.run-as=0` in the `postfix` pipeline.
2. It will fail even when running as root without capabilities. Running the same with `sudo melange` it will build with success instead.

### Suggested Fix

Introduce an API in the Melange pipeline manifest to set capabilities to be added to the build and test processes. Map them with the related runner.

maxgio92:
This PR adds the configuration to add and drop Linux capabilies to the container for bubblewrap and docker runners, enabling a declarative way to do least privilege. The capabilities added or dropped are relative to the minimum set of capabilities Docker sets. They are applied to all both build and test pipelines.

Furthermore, the minimum set of process capabilities in Bubblewrap container is now the same of Docker default capabilities, to improve result consistency.

Fixes #1703

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 	github.com/klauspost/compress v1.17.11
 	github.com/klauspost/pgzip v1.2.6
 	github.com/kubescape/go-git-url v0.0.30
+	github.com/moby/moby v27.4.0+incompatible
 	github.com/opencontainers/image-spec v1.1.0
 	github.com/package-url/packageurl-go v0.1.3
 	github.com/pkg/errors v0.9.1
@@ -93,6 +94,8 @@ require (
 	github.com/cloudflare/circl v1.5.0 // indirect
 	github.com/cncf/xds/go v0.0.0-20240905190251-b4127c9b8d78 // indirect
 	github.com/common-nighthawk/go-figure v0.0.0-20210622060536-734e95fb86be // indirect
+	github.com/containerd/containerd v1.7.24 // indirect
+	github.com/containerd/log v0.1.0 // indirect
 	github.com/containerd/stargz-snapshotter/estargz v0.16.2 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.5 // indirect
 	github.com/cyberphone/json-canonicalization v0.0.0-20231217050601-ba74d44ecf5f // indirect
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -595,6 +595,7 @@ type Input struct {
 	Required bool `json:"required,omitempty"`
 }
 
+// Capabilities is the configuration for Linux capabilities for the runner.
 type Capabilities struct {
 	// Linux process capabilities to add to the pipeline container.
 	Add []string `json:"add,omitempty" yaml:"add,omitempty"`
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -595,12 +595,22 @@ type Input struct {
 	Required bool `json:"required,omitempty"`
 }
 
-// The root melange configuration
+type Capabilities struct {
+	// Linux process capabilities to add to the pipeline container.
+	Add []string `json:"add,omitempty" yaml:"add,omitempty"`
+	// Linux process capabilities to drop from the pipeline container.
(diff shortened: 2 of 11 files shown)


go.mod
```diff
@@ -25,6 +25,7 @@ require (
 	github.com/klauspost/compress v1.17.11
 	github.com/klauspost/pgzip v1.2.6
 	github.com/kubescape/go-git-url v0.0.30
+	github.com/moby/moby v27.4.0+incompatible
```
maxgio92, line 28: This is the cost of aligning to Docker default capabilities.

pkg/container/config.go
```diff
@@ -46,6 +46,7 @@ type Config struct {
 	PackageName           string
 	Mounts                []BindMount
 	Capabilities          Capabilities
+	CapAdd                []string // List of kernel capabilities to add to the container.
```
maxgio92, line 49: Capability drop instead I think should be responsibility of the single runner (e.g. always dropping cap_sys_admin).

jonjohnsonjr: Can we fold this into the Capabilities field above?

maxgio92: Sure @jonjohnsonjr. Addressed with https://github.com/chainguard-dev/melange/pull/1702/commits/f2a5a53914168d5a0faae85173894d33113258a8

pkg/config/config.go
```diff
@@ -594,12 +594,16 @@ type Input struct {
...
-// The root melange configuration
+// Configuration is the root melange configuration.
 type Configuration struct {
 	// Package metadata
 	Package Package `json:"package" yaml:"package"`
 	// The specification for the packages build environment
 	Environment apko_types.ImageConfiguration `json:"environment" yaml:"environment"`
+	// TODO: move them a runtime specialized section.
+	// Linux process capabilities to add to the pipeline container.
+	CapAdd []string `json:"cap-add,omitempty" yaml:"cap-add,omitempty"`
```
jonjohnsonjr, line 605: Can you introduce a new Capabilities object to track what's in Build? We probably want the ability to "drop" capabilities too and it would be nice if they lived in the same struct and mirrored Build for clarity. I'd also make this just "add" so it'd look something like:

```yaml
capabilities:
  add: ["SETUID"]
```

And in the future:

```yaml
capabilities:
  add: ["SETUID"]
  drop: ["NET_ADMIN"]
```

Or whatever.

maxgio92: I like it. I just addressed it, alongside the capability drop option.
Do we want to add a specialized test for this?

jonjohnsonjr: This is great, yes, can you add this as a test and in the examples/ ?

maxgio92: Thanks @jonjohnsonjr Just added e2e tests for both build and test.
Plus a pipeline example with `iproute`.

89luca89 (approved):

philroche (approved):
Thank you. This will unblock many confusing local package build failures we see in sustaining where we end up doing a docker/bubblewrap/sudo dance to try get a successful build.

PR #1716, opened 2024-12-18
go/build pipeline: Move `go mod tidy` after `cd` to `modroot` dir AND Validate existence of `go.mod` in `modroot`

debasishbsws:
- fix: Move `go mod tidy` after `cd` to `modroot` dir

  - Previously, in the `go/build` pipeline, if `modroot` was set and `tidy=true`, the `go mod tidy` command ran in the home directory instead of the `modroot` directory.
- feat: Validate existence of `go.mod` in `modroot`

  - Added a check to ensure `go.mod` exists in the `modroot` directory, ensuring `modroot` is set correctly. This will help the reliability of `go/bump` automation as it also set its modroot by looking into the `go/build` pipeline modroot.

> Notes: this will result in build failure of PR if there is any package where the modroot is set incorrectly.

### DIFF
--- a/pkg/build/pipelines/go/build.yaml
+++ b/pkg/build/pipelines/go/build.yaml
@@ -100,7 +100,7 @@ pipeline:
   - runs: |
       cd "${{inputs.modroot}}"
 
-      // check if modroot is set correctly by checking go.mod file exist
+      # check if modroot is set correctly by checking go.mod file exist
       if [ ! -e go.mod ]; then
         echo "go.mod not found in ${{inputs.modroot}}"
         exit 1
--- a/pkg/build/pipelines/go/build.yaml
+++ b/pkg/build/pipelines/go/build.yaml
@@ -98,16 +98,22 @@ inputs:
 
 pipeline:
   - runs: |
+      cd "${{inputs.modroot}}"
+
+      // check if modroot is set correctly by checking go.mod file exist
+      if [ ! -e go.mod ]; then
+        echo "go.mod not found in ${{inputs.modroot}}"
+        exit 1
+      fi
+
       "${{inputs.tidy}}" && go mod tidy
-      
+
       LDFLAGS="${{inputs.strip}} ${{inputs.ldflags}}"
 
       BASE_PATH="${{inputs.prefix}}/${{inputs.install-dir}}/${{inputs.output}}"
 
       # Take advantage of melange's buid cache for downloaded modules
       export GOMODCACHE=/var/cache/melange/gomodcache
-
-      cd "${{inputs.modroot}}"
 
       # Install any specified dependencies
       if [ ! "${{inputs.deps}}" == "" ]; then


rawlingsj:
This will need an epoch bump

hectorj2f (changes requested):
I cannot see the need for these changes, if go mod is not found `go mod tidy` or `go/bump` will fail anyway.

hectorj2f:
The default error is also pretty descriptive today telling us that there is not a go.mod file on the specified location.

debasishbsws:
@hectorj2f yes true that the `go mod tidy` or the `go/bump` will fail in the go.mod is not present but The issue is that the `go/build` will not fail if we set the `modroot` incorrectly. and `go/bump` used the `modroot` from `go/build` so if that is set incorrectly we will have a incorrect automated go/bump with wrong modroot.

See the fix [here](https://github.com/chainguard-dev/enterprise-packages/pull/11139/files) where setting the modroot incorrectly doesn't cause any build failure. Slack [conv](https://chainguard-dev.slack.com/archives/C0799FD4Q1Y/p1734357268510169)

luhring:
I believe this change **is needed**! It enforces correct setup of the `go/build` pipeline and helps us avoid toil from broken automated PRs that depend on `go/build`'s `modroot` being set correctly.

But I'm concern about how breaking of a change this could be at the outset. @debasishbsws have you looked at which of our packages this might break the build for? This could be a disruptive change that brings our build system to a halt, if we're not careful.

hectorj2f:
> @hectorj2f You're correct that go mod tidy or go/bump will fail if the go.mod file is missing. However, the issue is that the go/build pipeline does not fail if the modroot is set incorrectly, as running go mod tidy is not enabled by default.

👍🏻

debasishbsws:
> But I'm concern about how breaking of a change this could be at the outset. @debasishbsws have you looked at which of our packages this might break the build for? This could be a disruptive change that brings our build system to a halt, if we're not careful.

@luhring I was thinking of identifying packages this might break but that need looking into the source code or try to build all the go package which use go/build pipeline. But I don't think this could be very disruptive. As we don't build every package every day. We will encounter this build failure with clear message in the package updates or CVE rem PR. And at that time we have to look and try to fix the package.

luhring:
> > But I'm concern about how breaking of a change this could be at the outset. @debasishbsws have you looked at which of our packages this might break the build for? This could be a disruptive change that brings our build system to a halt, if we're not careful.
> 
> @luhring I was thinking of identifying packages this might break but that need looking into the source code or try to build all the go package which use go/build pipeline. But I don't think this could be very disruptive. As we don't build every package every day. We will encounter this build failure with clear message in the package updates or CVE rem PR. And at that time we have to look and try to fix the package.

Right, I think once this causes a build to fail, we can put the time into to fix it once and for all. And then things will be awesome for that package.

But for me at least, I don't have a good sense of the total time that will require, and how soon we'll need to invest it. For example, if it's just an occasional extra 5 minutes, a few times a week for the next few months, then I agree that's not too bad. But if it starts block all our CVE fixes and requires dozens of hours a day from multiple staff members for the foreseeable future, that would be terrible, of course.

So my thinking is that we don't know where on that spectrum this would fall. Does that make sense? Do you have a better sense than I do about how this will play out for us if we merge this today?

debasishbsws:
@luhring can you approve this one? as we discus in the standup and we agree.

luhring (approved):
Yep! Approved. I'd just say to make sure you and whoever else are "owning" any fallout from this, keeping track of how this gets rolled out to prod, and managing whatever number of fixes end up being needed for this.

But I'm very excited to see us tightening this up! 🚀

debasishbsws:
I will put an announcement in the channel to determined when we have it applicable. and how to identify and fix this failure.

PR #1737, opened 2025-01-09
build: Write a GCC specs file into workspace

xnox:
Write out a GCC specs file into the workspace, that toolchain can
import to adjust build settings on per-package basis. For now, this
only adds Package Metadata ELF note settings.

In the future, this can be extended with other settings, for example
controlling generation of debug symbols, march/mtune/optimisations,
etc.

Potentially workspace is not a good location for this file. Maybe
worth considering to move this from /home/build/.melange.gcc.spec to
/etc/melange/gcc.spec.

By itself this PR is innert, until our default spec files start including the newly generated snippets when present. See:
- https://github.com/wolfi-dev/os/pull/39152

For the proposed usage of this change, which initially makes it opt-in for 3 packages to shake out any e2e integration.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -694,7 +694,9 @@ func (b *Build) populateWorkspace(ctx context.Context, src fs.FS) error {
 	if err := template.Must(specTemplate.Parse(gccLinkTemplate)).Execute(specFile, b); err != nil {
 		return err
 	}
-
+	if err := specFile.Close(); err != nil {
+		return err
+	}
 	return fs.WalkDir(src, ".", func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
 			return err
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -691,8 +691,7 @@ func (b *Build) populateWorkspace(ctx context.Context, src fs.FS) error {
 		return err
 	}
 	specTemplate := template.New("gccSpecFile")
-	err = template.Must(specTemplate.Parse(gccLinkTemplate)).Execute(specFile, b)
-	if err != nil {
+	if err := template.Must(specTemplate.Parse(gccLinkTemplate)).Execute(specFile, b); err != nil {
 		return err
 	}
 
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -30,6 +30,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"text/template"
 	"time"
 
 	"chainguard.dev/apko/pkg/apk/apk"
@@ -74,6 +75,10 @@ echo "cleaning Workspace by removing $# file/directories in $d"
 rm -Rf "$@"`,
 	"shellEmptyDir",
 }
+
+var gccLinkTemplate = `*link:
++ --package-metadata={"type":"apk","os":"{{.Namespace}}","name":"{{.Configuration.Package.Name}}","version":"{{.Configuration.Package.FullVersion}}","architecture":"{{.Arch.ToAPK}}"}
+`
 
 var ErrSkipThisArch = errors.New("error: skip this arch")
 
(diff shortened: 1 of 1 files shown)


stevebeattie:
> Potentially workspace is not a good location for this file. Maybe worth considering to move this from /home/build/.melange.gcc.spec to /etc/melange/gcc.spec.

Perhaps I'm overthinking things, but I do think in general a sensible direction for melange to go is to keep build configuration settings separate and ideally not tamperable by the build process of the packaged software itself, especially if we're providing a structured mechanism for said configuration (i.e. environment changes that are made outside of a run: invocation).

In this case, of course, the build can override or unset gcc/linker flags as part of how it invokes the compiler, so if it wanted to eliminate or modify the emitted ELF package note, it could still do so. Presumably we'd want a CI check to ensure that the note is generated as expected.

xnox:
> > Potentially workspace is not a good location for this file. Maybe worth considering to move this from /home/build/.melange.gcc.spec to /etc/melange/gcc.spec.
> 
> Perhaps I'm overthinking things, but I do think in general a sensible direction for melange to go is to keep build configuration settings separate and ideally not tamperable by the build process of the packaged software itself, especially if we're providing a structured mechanism for said configuration (i.e. environment changes that are made outside of a run: invocation).
> 
> In this case, of course, the build can override or unset gcc/linker flags as part of how it invokes the compiler, so if it wanted to eliminate or modify the emitted ELF package note, it could still do so. Presumably we'd want a CI check to ensure that the note is generated as expected.

Correct, all of this is bypassable if so desired.

W.r.t. checks I do need to file a feature request for melange lint checks, to verify at build time, that binaries are auditable (i.e. go binaries have buildinfo, dependency metatada, and sybmols; rust binaries have cargo-auditable info; elf binaries have package-metadata) such that we can start to be alerted about binaries getting produced in the .apk which lack these things. And possibly a mirrored ci-* check that can be rerun on existing .apks to verify these things. Note sure if this will end up in mono, melange, or wolfictl.

pkg/build/build.go
```diff
@@ -678,6 +683,19 @@ func (b *Build) populateWorkspace(ctx context.Context, src fs.FS) error {
 		return err
 	}
 
+	// Write out build settings into workspacedir
+	// For now, just the gcc spec file and just link settings.
+	// In the future can control debug symbol generation, march/mtune, etc.
+	specFile, err := os.Create(filepath.Join(b.WorkspaceDir, ".melange.gcc.spec"))
```
xnox, line 689: @jonjohnsonjr i don't like that i have to create this file inside the workspace. I really wish to have it in "/etc/melange/gcc.spec" in the container image that melange builds with apko. And apko can create files, dirs and symlinks.... but it doesn't have ability to write out a text file which would be really nice.

I really wish https://github.com/chainguard-dev/apko/blob/main/docs/apko_file.md#paths would have a key called "contents" with a multi-line string.

smoser (dismissed):
I'm fine with this as a step towards getting the file written into some immutable path (like `/melange/` or something).  

If that is doable soon, then lets do it, if not then lets land this and open an issue to do it right.

pkg/build/build.go
xnox: forking to https://github.com/chainguard-dev/melange/issues/1750 in case this is deemed not easy/quick to implement. As it really is a wishlist.

pkg/build/build.go
```diff
@@ -678,6 +683,19 @@ func (b *Build) populateWorkspace(ctx context.Context, src fs.FS) error {
...
+	if err != nil {
+		return err
+	}
+	specTemplate := template.New("gccSpecFile")
+	err = template.Must(specTemplate.Parse(gccLinkTemplate)).Execute(specFile, b)
+	if err != nil {
```
jonjohnsonjr, line 695:
```suggestion
	if err := template.Must(specTemplate.Parse(gccLinkTemplate)).Execute(specFile, b); err != nil {
```

pkg/build/build.go
```diff
@@ -678,6 +683,19 @@ func (b *Build) populateWorkspace(ctx context.Context, src fs.FS) error {
...
+		return err
+	}
+
```
jonjohnsonjr, line 698:
```suggestion
	if err := specFile.Close(); err != nil {
		return err
	}
```

pkg/build/build.go
jonjohnsonjr: We could hoist this out into an `init()` func but this is such a small template to parse it doesn't really matter.

xnox: this template is parsed and executed exactly once; irrespective of where/when it happens =)

jonjohnsonjr (approved):

PR #1738, opened 2025-01-09
Suggest adding the pkgconf test pipeline

murraybd:
This adds a warning when building a package to let developers know that the package would benefit from a pkgconf test pipeline, however it isn't smart enough to know if the test already exists.  Here's an example run of it though:

```
2025/01/09 13:55:14 INFO scanning for pkg-config data...
2025/01/09 13:55:14 WARN yajl-dev provides .pc files and could use a pkgconf test pipeline
2025/01/09 13:55:14 INFO   found pkg-config yajl for usr/share/pkgconfig/yajl.pc
2025/01/09 13:55:14 INFO scanning for python modules...
```


murraybd:
This is still in progress.

PR #1741, opened 2025-01-13
stop generation homebrew update as melange in in homebrew core

cpanato:
melange homebrew is in homebrew core now, so we dont need to generate the update script to our hosted tap

https://github.com/Homebrew/homebrew-core/blob/master/Formula/m/melange.rb

this cleans up our side a bit.
will remove the sts policy as soon this PR is approved and merged

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -72,18 +72,9 @@ jobs:
           version: latest
           install-only: true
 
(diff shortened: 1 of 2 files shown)


cpanato:
also related to https://github.com/chainguard-dev/homebrew-tap/pull/75

k4leung4 (approved):

PR #1753, opened 2025-01-16
remove 'melange convert python'

imjasonh:
This wasn't producing python packages in keeping with best practices, and keeping it up to date was cumbersome. It's unclear that anybody uses this day-to-day, so let's remove it.


EyeCantCU (dismissed):

xnox (dismissed):
this is in contrast to this pr:
- https://github.com/chainguard-dev/melange/pull/1464

which updates it to current best practices, which has been failing to land due to e2e tests.....

@justinvreeland what do you prefer here?

justinvreeland:
I don't have a problem removing it. I don't see a need to fix the failing tests again for something we're not likely to keep using.

EyeCantCU (approved):

PR #1758, opened 2025-01-22
disallow uses: with a nested pipeline

imjasonh:


imjasonh:
There will almost definitely be some configs in Wolfi that are invalid under this new rule.

kranurag7 (approved):

PR #1762, opened 2025-01-27
Add ${{subpkg.name}} string substitution

dannf:
I've noticed a growing pattern of using `$(basename ${{targets.contextdir}})` in our YAML to calculate the subpackage name. This only works when the string is processed by a shell, and is really just a hack anyway. Let's introduce a formal substitution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ Melange provides the following default substitutions which can be referenced in 
 | `${{package.full-version}}` | `${{package.version}}-r${{package.epoch}}`                               |
 | `${{package.description}}`  | Package description                                                      |
 | `${{package.srcdir}}`       | Package source directory (`--source-dir`)                                |
+| `${{subpkg.name}}`          | Subpackage name                                                          |
 | `${{targets.outdir}}`       | Directory where targets will be stored                                   |
 | `${{targets.contextdir}}`   | Directory where targets will be stored for main packages and subpackages |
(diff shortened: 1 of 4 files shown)


murraybd:
> I've noticed a growing pattern of using `$(basename ${{targets.contextdir}})` in our YAML to calculate the subpackage name. This only works when the string is processed by a shell, and is really just a hack anyway. Let's introduce a formal substitution.

Does one person's work constitute a pattern? ;-)

murraybd:
A cursory glance seems to indicate that `pkg/config/config_test.go` contains tests that might be relevant to this change.

dannf:
> > I've noticed a growing pattern of using `$(basename ${{targets.contextdir}})` in our YAML to calculate the subpackage name. This only works when the string is processed by a shell, and is really just a hack anyway. Let's introduce a formal substitution.
> 
> Does one person's work constitute a pattern? ;-)

When that person's work becomes contagious, yes:
https://github.com/wolfi-dev/os/blob/59a1f2a4d228bae07a6394ca54bd8890dfb96da5/pipelines/test/docs.yaml#L14

dannf:
> A cursory glance seems to indicate that `pkg/config/config_test.go` contains tests that might be relevant to this change.

Good call, I'll take a look.

> Also README.md could use updating.

I added it to the table of substitutions already, is there some other section that requires updating?

murraybd (approved):

PR #1763, opened 2025-01-27
Add a linter which checks for pkgconfig files

murraybd:
This checks for the package providing a `pkgconfig` directory and raises a warning about adding the pkgconf test pipeline to the package.


pkg/linter/linter.go
```diff
@@ -176,6 +176,11 @@ var linterMap = map[string]linter{
 		Explain:         "Remove all test directories from the package",
 		defaultBehavior: Warn,
 	},
+	"pkgconf": {
+		LinterFunc:      allPaths(pkgconfTestLinter),
+		Explain:         "This package provides files in a pkgconfig directory, please add the pkgconf test pipeline",
+		defaultBehavior: Warn,
```
imjasonh, line 182: Will this always warn, even if I already have a pkgconf test?

I think this might indicate the package linter doesn't have enough visibility into the package config, to be able to avoid this warning when it's already taken care of.

murraybd: Yes, it will always warn and that's something I'd like to sit down in person and sort out. It requires getting the whole config / yaml file visible to the linter.

imjasonh (dismissed):

imjasonh (approved):

PR #1778, opened 2025-02-07
Consolidate build user/group strings

egibs:
This PR controls the build user/group string from only two places rather than hard-coding the value in several places.

### DIFF
--- a/pkg/build/build_test.go
+++ b/pkg/build/build_test.go
@@ -37,7 +37,6 @@ var requireErrInvalidConfiguration require.ErrorAssertionFunc = func(t require.T
 // TestConfiguration_Load is the main set of tests for loading a configuration
 // file. When in doubt, add your test here.
 func TestConfiguration_Load(t *testing.T) {
-	gid1000 := uint32(1000)
 	tests := []struct {
 		name                string
 		skipConfigCleanStep bool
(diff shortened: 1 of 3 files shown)


pkg/build/build_test.go
```diff
@@ -209,8 +208,8 @@ func TestConfiguration_Load(t *testing.T) {
 						"GOPATH":     "/var/cache/melange/go",
 					},
 					Accounts: apko_types.ImageAccounts{
-						Users:  []apko_types.User{{UserName: "build", UID: 1000, GID: apko_types.GID(&gid1000)}},
-						Groups: []apko_types.Group{{GroupName: "build", GID: 1000, Members: []string{"build"}}},
+						Users:  []apko_types.User{{UserName: buildUser, UID: 1000, GID: apko_types.GID(&gid1000)}},
+						Groups: []apko_types.Group{{GroupName: buildUser, GID: 1000, Members: []string{buildUser}}},
```
stevebeattie, line 212: Again, limited understanding of golang declarations, but doesn't `buildUser` need to be defined first?

(But then, how does go test pass?)

pkg/build/test_test.go
```diff
@@ -39,14 +39,14 @@ const (
 	homeBuild        = "/home/build"
 )
 
-var (
-	gid1000 = uint32(1000)
-)
+const buildUser = "build"
```
stevebeattie, line 42: My understanding of golang declarations is shaky at best, but this could be in the rest of the const declarations above, no?

egibs: Oh, yep, I missed those.

pkg/build/build_test.go
egibs: `buildUser` is accessible from anywhere in the `build` package 👍🏻

stevebeattie: Nevermind, got where it's coming from.

stevebeattie (approved):

PR #1785, opened 2025-02-13
Add support for 'finalize' pipelines.

smoser:
* add greeter-build e2e-test that does build and test.
    
    This also tests the finalize code.

  * Add support for 'finalize' pipelines.
    
    The idea here is that 'finalize' has access to the fully
    populated 'melange-out' directory after all subpackages
    have finished.
    
    Without this only the last subpackage has access to
    all of the melange-out/ data.  So you could utilize
    the last subpackage to do "finalize" things, but that
    seems very yucky.
    
    The initial use for this is to melange-helpers usrmerge
    to move files around in all the packages/subpackages
    at the end of the build.
    https://github.com/wolfi-dev/os/pull/42552

### DIFF
--- a/docs/BUILD-FILE.md
+++ b/docs/BUILD-FILE.md
@@ -548,3 +548,6 @@ TODO(vaikas): melange config points to apko here:
 # pipeline
 Pipeline defines the ordered steps to build the package.
 
+# finalize
+Finalize defines a series of pipelines that are executed
+after the build and subpackages are processed.
--- a/e2e-tests/greeter-build-test.yaml
+++ b/e2e-tests/greeter-build-test.yaml
@@ -60,12 +60,31 @@ subpackages:
             ${{vars.tester-blob}}
             testrun ${{range.value}}
 
+finalize:
+  - name: "write /etc/finalized"
+    runs: |
+      ( cd "${{targets.outdir}}" && find . ! -type d )
+      etcd="${{targets.outdir}}/${{package.name}}/etc"
+      mkdir -p "$etcd"
+      echo "finalized-${{package.name}}" > "$etcd/finalized"
+
 test:
   pipeline:
     - name: test-greeter
       runs: |
         ${{vars.tester-blob}}
         testrun greeter howdy
+    - name: test-finalized
+      runs: |
+        set +f
+        f="/etc/finalized"
+        [ -e "$f" ] || { echo "FAIL: $f is not present"; exit 1; }
+        [ -f "$f" ] || { echo "FAIL: $f was not a file"; exit 1; }
+        exp="finalized-${{package.name}}"
+        read content < "$f" || { echo "FAIL: failed to read from $f"; exit 1; }
+        [ "$content" = "$exp" ] ||
+          { echo "$f contained '$content'. expected '$exp'"; exit 1; }
+        echo "PASS: $f existed and had expected content '$exp'"
 
 vars:
   tester-blob: |
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -701,6 +701,14 @@ func (b *Build) BuildPackage(ctx context.Context) error {
 		linterQueue = append(linterQueue, lintTarget)
 	}
 
+	// Run finalize pipelines
+	if !b.isBuildLess() && len(b.Configuration.Finalize) > 0 {
+		log.Infof("running finalize pipelines for %s", pkg.Name)
+		if err := pr.runPipelines(ctx, b.Configuration.Finalize); err != nil {
+			return fmt.Errorf("unable to run finalize pipelines for %s: %v", pkg.Name, err)
+		}
+	}
+
 	// Store metadata for use after the workspace is loaded into memory
 	xattrs, modes, owners, err := storeMetadata(b.WorkspaceDir)
 	if err != nil {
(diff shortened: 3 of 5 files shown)


e2e-tests/greeter-build-test.yaml
```diff
@@ -0,0 +1,121 @@
...
+        - runs: |
+            ${{vars.tester-blob}}
+            testrun aloha
+
+finalize:
+  - name: post pipeline
+    runs: |
+      find melange-out/ ! -type d
+      mkdir -p melange-out/${{package.name}}/etc
+      echo "finalized-${{package.name}}" > melange-out/${{package.name}}/etc/finalized
```
dannf, line 75: should we use `${{targets.destdir}}` and `${{targets.outdir}}` here instead of hard-coding `melange-out` ?

e2e-tests/greeter-build-test.yaml
```diff
@@ -0,0 +1,121 @@
...
+      touch melange-out/${{package.name}}/etc
```
dannf, line 76: what's this for?

dannf (dismissed):
Very cool.

e2e-tests/greeter-build-test.yaml
smoser: i didn't know of 'targets.outdir'. yes, that fits.  i think i'd pick still using 'package.name' as the assumption/contract should be that `${{targets.outdir}}/<subpackage-or-package>/` and `destdir` doesn't make that clear.

e2e-tests/greeter-build-test.yaml
```diff
@@ -0,0 +1,121 @@
...
+          mkdir -p "${{targets.contextdir}}/usr/bin"
+          ln -s greeter "${{targets.contextdir}}/usr/bin/hello"
+    test:
+      pipeline:
+        - runs: |
+            ${{vars.tester-blob}}
+            testrun hello
+
+  - name: ${{package.name}}-aloha
+    description: hawiaan greeting
```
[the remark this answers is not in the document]
smoser: there is no way i would have spelled that right... but sure, we'll say it was a typo.

e2e-tests/greeter-build-test.yaml
smoser: dropped.

pkg/build/build.go
```diff
@@ -946,6 +946,14 @@ func (b *Build) BuildPackage(ctx context.Context) error {
 		linterQueue = append(linterQueue, lintTarget)
 	}
 
+	// Run finalize pipelines
+	if !b.isBuildLess() && len(b.Configuration.Finalize) > 0 {
+		log.Infof("running pipelines for %s", pkg.Name)
+		if err := pr.runPipelines(ctx, b.Configuration.Finalize); err != nil {
+			return fmt.Errorf("unable to run post pipelines for %s: %v", pkg.Name, err)
```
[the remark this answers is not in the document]
smoser: done.

e2e-tests/greeter-build-test.yaml
smoser: done.

dannf (dismissed):
@smoser thanks, this looks like a great addition to `melange`.

dannf:
Oh, could you also document this in `BUILD-FILE.md` ? Once merged, I'd suggest linking to that doc in #eng-psa

smoser:
> Oh, could you also document this in `BUILD-FILE.md` ? Once merged, I'd suggest linking to that doc in #eng-psa

:+1:  will-do

docs/BUILD-FILE.md
```diff
@@ -36,6 +36,14 @@ The following are the high level sections for the build file, with detailed desc
 ### options
 
    Deviations to the build
+### finalize
+
+   List of pipelines to execute after main `pipeline` and `subpackages`
+   are processed.
+
+   The pipelines can used substitutions available in the package build.
+   Pipelines should use `${{vars.melange-out}}` to determine the location
```
dannf, line 45: What is `${{vars.melange-out}}` ?

smoser: i have no idea.  fixed.

dannf (dismissed):
izskatās labi

smoser:
I talked to @jonjohnsonjr and we decided to keep this as a draft for now.
The primary reason for doing so was a lack of something to use it for.
That is, I wrote it to be a place where we could easily add a "usrmerge cleaner" , but we decided to go a different route with usrmerge.

If we _do_ have a need, then we can access again.

@dannf do you have something that you'd use this for ?

[unrelated, i think i will add the e2e test for greeter and the build/test of it as those are useful]

smoser:
e2e test changes are up https://github.com/chainguard-dev/melange/pull/1801

dannf:
> I talked to @jonjohnsonjr and we decided to keep this as a draft for now. The primary reason for doing so was a lack of something to use it for. That is, I wrote it to be a place where we could easily add a "usrmerge cleaner" , but we decided to go a different route with usrmerge.
> 
> If we _do_ have a need, then we can access again.
> 
> @dannf do you have something that you'd use this for ?

Not at the moment. +1 on delaying until we do have such a need.

Issue #1731, opened 2025-01-06
qemu build runner does not preserve setuid/setgid/extended attributes in built packages


stevebeattie:
when building with the qemu runner, packages that set and should contain setuid binaries do not do so. For example, building the sudo package like so:

```
$ make package/sudo MELANGE_EXTRA_OPTS="--runner qemu"
yamlfile is sudo.yaml
Building package sudo with version sudo-1.9.16-r1 from file sudo.yaml
make yamlfile=sudo.yaml pkgname=sudo packages/x86_64/sudo-1.9.16-r1.apk
make[1]: Entering directory '$HOME/git/wolfi-dev/os'
@SOURCE_DATE_EPOCH=1732043006 $HOME/go/bin/melange build sudo.yaml --runner=qemu --repository-append $HOME/git/wolfi-dev/os/packages --keyring-append local-melange.rsa.pub --signing-key local-melange.rsa --arch x86_64 --env-file build-x86_64.env --namespace wolfi --license 'Apache-2.0' --git-repo-url 'https://github.com/wolfi-dev/os' --generate-index false  --pipeline-dir ./pipelines/ --runner qemu -k https://packages.wolfi.dev/os/wolfi-signing.rsa.pub -r https://packages.wolfi.dev/os --repository-append $HOME/git/wolfi-dev/os/packages --keyring-append local-melange.rsa.pub --signing-key local-melange.rsa --arch x86_64 --env-file build-x86_64.env --namespace wolfi --license 'Apache-2.0' --git-repo-url 'https://github.com/wolfi-dev/os' --generate-index false  --pipeline-dir ./pipelines/ --runner qemu -k https://packages.wolfi.dev/os/wolfi-signing.rsa.pub -r https://packages.wolfi.dev/os --source-dir ./sudo/
[...]
[...]

PR #1804, opened 2025-02-21
DRAFT: bug(qemu runner): use gnutar for xattr+acl support

stevebeattie:
Unlike the other runners which use a shared directory between the runner and the host OS, the qemu runner tars up the built artifacts in the vm and then untars them (through go's tar implementation) in the host environment, using busybox's tar.

There were two problems with this:

- --preserve-permissions (aka -p) was not being passed, so non-default permissions like setgid/setuid were not being captured in the tarball
- busybox tar reads PAX format tarfiles but does not generate them, so has no mechanism for including them in generated tarballs.

To fix this, convert to using gnutar by always including it in the environment when using the qemu runner, and secondly, pass the -p, --xattrs, and --acls arguments to tar so they are captured in the generated tarball.

This only partially addresses the issue, for two reasons:

1. melange is not setting xattrs when extracting from the tarball
2. in order for setuid root permissions and security xattrs (for fscaps) to be created in the host environment's file system, root (or more correctly, CAP_DAC_OVERRIDE, CAP_FOWNER, and CAP_SYSADMIN for setuid, acls, and fscap xatrrs respectively).

Fixes-partially: https://github.com/chainguard-dev/melange/issues/1731

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -310,6 +310,7 @@ func (b *Build) buildGuest(ctx context.Context, imgConfig apko_types.ImageConfig
 	if b.Runner.Name() == container.QemuName {
 		b.ExtraPackages = append(b.ExtraPackages, []string{
 			"melange-microvm-init",
+			"gnutar", // needed for xattrs/acls when extracting built files in vm
 		}...)
 	}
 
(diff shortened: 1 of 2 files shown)


stevebeattie:
Also, this really needs some end2end test cases added.

PR #1813, opened 2025-02-26
rebuild: populate Config directly without file

imjasonh:
Prior to this change `build.New`'s options only accepted a config file path, which it would parse and use. The filename of the config is passed to the SBOM.

This complicated `rebuild` which doesn't have a real file path to rebuild, but it wanted to populate the same file path to be able to rebuild reproducibly.

With this change, `build.WithConfiguration(cfg, filename)` populates the pre-parsed config struct, and populates the filename to report in the SBOM, which is a little disingenuous, but 🤷.

I suspect there are lots of places where we call `build.New` with a temp file just to be able to pass it to `build.WithConfig(tmpfile)`, which we no longer need to do. If we find them, we can clean them up and maybe hopefully deprecate `WithConfig`.

I'm not married to the name `build.WithConfiguration`, if you have a better suggestion I'm open to it.

edit: while working on this PR, it seems the bootstrap test is broken now, potentially by a change in Go 1.24 that embeds the build commit and git tree state:

```
--rw-r--r-- root/root      2100 2022-11-29 01:05:10 var/lib/db/sbom/melange-0.0.1-r0.spdx.json
+-rw-r--r-- root/root      2133 2022-11-29 01:05:10 var/lib/db/sbom/melange-0.0.1-r0.spdx.json
```

```
-      "Tool: melange (0b48159)",
+      "Tool: melange (v0.0.0-20250226183030-0b48159d0fea+dirty)",
```

I've dropped the bootstrap test in this PR, and added coverage for melange itself in the rebuild matrix, which proves its reproducible given a fresh environment, instead of relying on local git state (which breaks reproducibility now)

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,4 +1,4 @@
-name: e2e melange bootstrap + build
+name: e2e tests
 
 on:
   push:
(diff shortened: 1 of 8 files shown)


jonjohnsonjr:
Can you clean this up while you're at it https://github.com/chainguard-dev/melange/blob/a46420071cfd3beb3a72327bd4254535bc8c43b2/pkg/linter/linter.go#L723-L739

luhring:
Can we add some kind of testing around the rebuild functionality, too? It's starting to get a little complex for my brain. I also know we were talking about not letting features break without CI noticing

imjasonh:
> Can we add some kind of testing around the rebuild functionality, too? It's starting to get a little complex for my brain. I also know we were talking about not letting features break without CI noticing

I've been focusing on e2e tests for that, since there's so many things involved in building and rebuilding. But, we have a rebuild test in e2e.yaml:

https://github.com/chainguard-dev/melange/actions/runs/13549641273/job/37869881109?pr=1813

I've just made this a required check for future PRs as well.

jonjohnsonjr (approved):

PR #1814, opened 2025-02-26
rebuild: e2e test reproducibly rebuilding melange itself

imjasonh:
We had a test for this in `bootstrap` but it wasn't as thorough as the test we have for rebuilding, so let's just reuse that.


imjasonh:
This needs some of the stuff in https://github.com/chainguard-dev/melange/pull/1813 and might just get folded into that.

PR #1821, opened 2025-02-27
feat(build/pipelines): add go/covdata pipeline

maxgio92:
It leverages a go binary built with coverage instrumentation using the -cover build flag alongside its coverage data genreated to produce a statements by functions based report via the go tool covdata.


joshrwolf (approved):

joshrwolf:
I guess it kinda sucks that:

```
        environment:
          GOCOVERDIR: /home/build

```

can't be defined via the pipeline, perhaps we can add to the docs that this is required to be set _before_ the tests (as you've done in the sample via the environment contents), and that it must match what is set into the pipeline, if it differs from `/home/build`

^ doesn't need to happen this PR though!

maxgio92:
Yeah @joshrwolf. At least when you run the binary if it's not set the warning is printed every time you executed as it's instrumented and it expects it to write cov data to somewhere.

PR #1827, opened 2025-02-28
feat(pkg/build): enable pipelines to inject environment

maxgio92:
This commit enable pipelines to inject environment variables the same way they're able to inject package dependencies into the build or test configuration. This will allow to remove responsibility to satisfy environment dependencies from the consumer. When setting needs.environment all the other pipeline runs in the same build or config will inherit that additional environment.

Cases like go/covdata pipeline will benefit from this being able to inject the GOCOVERDIR that the go runtime will use to generate coverage data and the go tool covdata to generate reports from.

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:


pkg/config/config.go
```diff
@@ -463,6 +463,8 @@ func (p Package) FullCopyright() string {
 type Needs struct {
 	// A list of packages needed by this pipeline
 	Packages []string
+	// A list of environment variables to inject to the build or test config.
+	Environment map[string]string
```
maxgio92, line 467: I'm not really convinced by this API as `needs.environment` can be confusing compared to `environment`.

maxgio92:
An example of use case would be the following:

```diff
diff --git a/pkg/build/pipelines/go/covdata.yaml b/pkg/build/pipelines/go/covdata.yaml
index b90b539..737856e 100644
--- a/pkg/build/pipelines/go/covdata.yaml
+++ b/pkg/build/pipelines/go/covdata.yaml
@@ -5,6 +5,8 @@ needs:
   packages:
     - ${{inputs.package}}
     - busybox
+  environment:
+    GOCOVERDIR: /home/build

 inputs:
   package:
```

```yml
# crane.yaml
...
subpackages:
  - name: ${{package.name}}-cover
    pipeline:
      - uses: go/build
        with:
          packages: ./cmd/crane
          extra-args: -cover
          ldflags: -buildid= -X github.com/google/go-containerregistry/cmd/crane/cmd.Version=${{package.version}} -X github.com/google/go-containerregistry/pkg/v1/remote/transport.Version=${{package.version}}
          output: crane
    test:
      environment:
        contents:
          packages:
            - jq
        # The following can be removed thanks to go/covdata needs.environment['GOCOVERDIR']=/home/build
        #environment:
        #  GOCOVERDIR: /home/build
      pipeline:
        - name: Verify Crane installation
          runs: |
            crane version || exit 1
            crane --help
        - name: Fetch and verify manifest
          runs: |
            crane manifest chainguard/static | jq '.schemaVersion' | grep '2' || exit 1
        - name: List tags for a public image
          runs: |
            crane ls chainguard/static | grep -E 'latest|v[0-9]+.[0-9]+.[0-9]+' || exit 1
        - name: Validate image existence
          runs: |
            crane digest chainguard/static:latest && echo "Image exists" || exit 1
        - name: Pull and save an image locally
          runs: |
            crane pull chainguard/static:latest static_latest.tar || exit 1
            [ -f static_latest.tar ] || exit 1
        - uses: go/covdata
   ```

**Review by imjasonh — COMMENTED** [review]
Can you add an example to `examples/` to demonstrate this working? That will act both as a test, and as runnable documentation of its behavior.

**Comment by maxgio92** [comment]
Yes, sure @imjasonh! I'll do it.

File: docs/BUILD-PROCESS.md [hunk 2]
```diff
@@ -24,6 +24,17 @@ needs:
...
+Each pipeline can also declare additional environment variables that are required for the build, such that all the pipelines will inherit it.
+For example with the following:
+
+```yaml
+needs:
+  environment:
+    FOO: bar
+```
+
+not only the specific pipeline but also all the pipelines of the build environment  will inherit the additional environment variable.
```
hectorj2f, line 36: nit: extra whitespace

maxgio92: Good catch! Thanks @hectorj2f

pkg/build/compile_test.go
```diff
@@ -114,10 +122,60 @@ func TestCompileTest(t *testing.T) {
...
 	if got, want := test.Configuration.Subpackages[0].Test.Environment.Contents.Packages, []string{"subpackage-base", "subpackage", "subpackage-need"}; !slices.Equal(got, want) {
 		t.Errorf("subpackage test packages: want %v, got %v", want, got)
 	}
+	if got, want := test.Configuration.Subpackages[0].Test.Environment.Environment, map[string]string{"FOO": "bar", "BAZ": "qux"}; !mapsEqual(got, want) {
+		t.Errorf("subpackage environment: want %v, got %v", want, got)
+	}
+}
+
+func TestCompileBuild(t *testing.T) {
+	build := &Build{
```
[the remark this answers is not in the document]
maxgio92: @hectorj2f  the pipeline-level `environment` wins over `needs.environment`, so for instance:
```yml
  pipeline:
    - name: Expect an environment transparently.
      environment:
        FOO: "baz"
      runs: |
        set -u
        echo "FOO: ${FOO}"
    - name: Inject an environment to the parent pipeline.
      needs:
        environment:
          FOO: "bar"
```

will end up with `$FOO` evaluated to _baz_.

docs/BUILD-PROCESS.md
maxgio92: Fixed

maxgio92:
> Can you add an example to examples/ to demonstrate this working? That will act both as a test, and as runnable documentation of its behavior.

@imjasonh an example which can be used at test as well has been added to _examples/needs-environment.yaml_.

pkg/build/compile_test.go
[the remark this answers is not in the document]
maxgio92: @hectorj2f I've added a specific case in the e2e test _examples/needs-environment.yaml_.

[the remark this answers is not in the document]
hectorj2f: Great!

hectorj2f (commented):
lgtm

examples/needs-environment.yaml
```diff
@@ -0,0 +1,87 @@
...
+        echo "BAZ: ${BAZ}"
+        if [[ "${BAZ}" != "baz" ]]; then
+          echo "Expected BAZ=baz, got BAZ=${FOO}"
+          exit 1
+        fi
+    - name: Inject an environment transparently to the parent pipeline.
+      needs:
+        environment:
+          FOO: "bar"
+          BAZ: "baz"
```
imjasonh, line 87: This seems weird, if I'm understanding correctly. A later step in the pipeline can set env vars for a previous step? What should happen if two steps overwrite each other, who wins?

I'd have expected this to need to come before the steps that use it, and then the rules of precedence are easier to understand (latest wins)

(I'm really glad this example exists, to make it clear how this is working!)

kranurag7:
👋🏻 Hello, if we merge this then it'll enable me/us to write pipelines which will allow us to define environment in the pipeline itself. I wanted to this patch inside test/kwok/cluster pipeline.

maxgio92:
I guess @kranurag7 is to export KUBERNETES_SERVICE_HOST + KUBERNETES_SERVICE_PORT environment variable to available to subsequent step's shell session

PR #1834, opened 2025-03-04
Ldd check linter

murraybd:
This adds a check to see if a package provides a `.so` file and if it does warns about adding the ldd-check test pipeline to the package. Additionally, it checks to see if the either `test/ldd-check` or `test/tw/ldd-check` are already present in the .yaml file.

Sample output:

`2025/03/04 09:51:28 WARN linter "lddcheck" failed on package "libdbi-dev": shared object found: usr/lib/libdbi.so; suggest: This package provides shared object files, please add the ldd-check test pipeline`

### DIFF
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -183,6 +183,11 @@ var linterMap = map[string]linter{
 		Explain:         "This package provides files in a pkgconfig directory, please add the pkgconf test pipeline",
 		defaultBehavior: Warn,
 	},
+	"lddcheck": {
+		LinterFunc:      allPaths(lddcheckTestLinter),
+		Explain:         "This package provides shared object files, please add the ldd-check test pipeline",
+		defaultBehavior: Warn,
+	},
 	"usrmerge": {
 		LinterFunc:      allPaths(usrmergeLinter),
 		Explain:         "Move binary to /usr/{bin,lib/sbin}",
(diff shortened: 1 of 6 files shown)


pkg/linter/linter.go
```diff
@@ -593,6 +598,52 @@ func pkgconfTestLinter(_ context.Context, cfg *config.Configuration, pkgname, pa
 	return fmt.Errorf("pkgconfig directory found")
 }
 
+var isSharedObjectFileRegex = regexp.MustCompile(`\.so$`)
```
jonjohnsonjr, line 601: Is this sufficient to catch everything we want?

I'm genuinely asking here because I'm not familiar with conventions, but I've often seen places where we have version numbers come after `.so` e.g. [abseil](https://apk.dag.dev/https/packages.wolfi.dev/os/x86_64/abseil-cpp-spinlock-wait-20240722.0-r0.apk@sha1:78fca72af0d8b54a16ff2380b2adcf08575c103a/?all=true).

I also wonder if we'd have any "false positives" where a file happens to end in `.so` but isn't a shared object... maybe we can cross that bridge when we get to it.

sergiodj: We need a more involved regexp.  Something like:

```
\.so(?:(?:\.[0-9]+)*)*
```

Also, the files ending in `.so` are almost always symlinks to the real shared libraries.  We're interested in running `ldd` against the real objects, IMO (even though `ldd` will dereference the symlinks).

pkg/linter/linter.go
```diff
@@ -593,6 +598,52 @@ func pkgconfTestLinter(_ context.Context, cfg *config.Configuration, pkgname, pa
...
+				if test.Uses == "test/ldd-check" {
+					return nil
+				} else if test.Uses == "test/tw/ldd-check" {
+					return nil
+				}
+			}
+		}
+	} else {
+		for _, p := range cfg.Subpackages {
+			if p.Name != pkgname {
```
sergiodj, line 624: We really should rename `pkgname` to `testname` (here and elsewhere, but no need to do it in this PR) in order to avoid that problem we were investigating :-/.

pkg/linter/linter.go
```diff
@@ -593,6 +598,52 @@ func pkgconfTestLinter(_ context.Context, cfg *config.Configuration, pkgname, pa
...
+			if p.Test == nil {
+				break
+			}
+
+			for _, test := range p.Test.Pipeline {
+				if test.Uses == "test/ldd-check" {
+					return nil
+				} else if test.Uses == "test/tw/ldd-check" {
+					return nil
+				}
```
sergiodj, line 637:
```suggestion
				if test.Uses == "test/ldd-check" || test.Uses == "test/tw/ldd-check" {
					return nil
				}
```

pkg/linter/linter.go
murraybd: We don't need to catch all the symlinks and shared libraries, one is sufficient to identify that the package or subpackage needs testing subsequently I think the expanded regex is good enough.

sergiodj (approved):

PR #1835, opened 2025-03-05
set GOFIPS140=latest  (build setting) sets GODEBUG=fips140 to on by d…

ajayk:
https://tip.golang.org/doc/go1.24#fips140

GOFIPS140=latest (build setting) sets GODEBUG=fips140 to on by default. Replaces -tags=requirefips.


xnox (commented):
I don't think this will work correctly. Because sometimes people specify go-msft-1.24 or go-fips~1.24 and so on.

PR #1850, opened 2025-03-10
pkg/sca/sca.go: Implement `getLdSoConfDLibPaths`

sergiodj:
This PR implements the machinery necessary for SCA to consider files under `/etc/ld.so.conf.d/`.

The idea is that, if there is a configuration for `ld.so.conf.d` being installed by the package on hand, then the libraries shipped by the package should be listed as `provides` if they're installed inside the paths listed in the `ld.so.conf.d` configuration file.

This commit also implements a simple test to make sure that the new function is properly parsing the configuration file inside `/etc/ld.so.conf.d/`.

Notes:

### SCA Changes

Notes:

As of this writing no packages are actually using `ld.so.conf.d` on Wolfi OS.  I built `/nvidia-cudnn-9` (from extra-packages) and it works as expected:

```
2025/03/10 17:31:12 INFO scanning for ld.so.conf.d files...
2025/03/10 17:31:12 INFO   found ld.so.conf.d file etc/ld.so.conf.d/nvidia-cudnn-9.conf
2025/03/10 17:31:12 INFO     found extra lib path /usr/local/cudnn-9/lib64
```

Relates: https://github.com/chainguard-dev/internal-dev/issues/7609


pkg/sca/sca.go
```diff
@@ -86,6 +87,73 @@ func isInDir(path string, dirs []string) bool {
...
+		if !strings.HasSuffix(path, ".conf") {
+			return nil
+		}
+
+		fi, err := d.Info()
+		if err != nil {
+			return err
+		}
+
+		if !fi.Mode().IsRegular() {
```
dannf, line 126: should we support symlinks to?

pkg/sca/sca.go
```diff
@@ -766,6 +834,15 @@ func generateShbangDeps(ctx context.Context, hdl SCAHandle, generated *config.De
 // Analyze runs the SCA analyzers on a given SCA handle, modifying the generated dependencies
 // set as needed.
 func Analyze(ctx context.Context, hdl SCAHandle, generated *config.Dependencies) error {
+	var oldLibDirs []string
+
+	if extraLibPaths, err := getLdSoConfDLibPaths(ctx, hdl); err != nil {
+		return err
+	} else if extraLibPaths != nil {
```
dannf, line 841: could this `else if` just be a separate `if`? sorry if that's a dumb question, I don't read `go` that well yet.

dannf (approved):
very cool! a couple comments inline, but I'm not sure either is a real issue.

pkg/sca/sca.go
sergiodj: Sure thing.  It doesn't matter really, but I understand why you think it's better.

sergiodj:
Hi @jonjohnsonjr!

I'd be delighted if you could offer your review here as well.

Thanks!

dannf (commented):
Looks like there's some test failures that need addressing, but this LGTM!

sergiodj:
> Looks like there's some test failures that need addressing, but this LGTM!

Thanks, @dannf!  I pushed a small change to fix the failure.

pkg/sca/sca.go
```diff
@@ -86,6 +87,84 @@ func isInDir(path string, dirs []string) bool {
...
+			fd, err := fsys.Open(path)
+			if err != nil {
+				return err
+			}
+			defer fd.Close()
+
+			scanner := bufio.NewScanner(fd)
+			for scanner.Scan() {
+				line := scanner.Text()
+				if line[0] != '/' {
```
jonjohnsonjr, line 145: How does this handle empty lines? I think we might panic here if `line == ""`.

pkg/sca/sca.go
```diff
@@ -804,5 +894,9 @@ func Analyze(ctx context.Context, hdl SCAHandle, generated *config.Dependencies)
 		generated.Provides = nil
 	}
 
+	if oldLibDirs != nil {
+		libDirs = oldLibDirs
```
jonjohnsonjr, line 898: This frightens me a bit -- instead of mutating the global here do you mind plumbing this around everywhere?

pkg/sca/sca.go
sergiodj: Good catch.  I've fixed it.

pkg/sca/sca.go
sergiodj: For sure.  But let me take the opportunity and ask a broader question.

I thought about moving `libDirs` inside the `SCABuildInterface`, because this really should be a per-package variable.  However, `SCABuildInterface` only represents one single entity in the package (either the sole package, or one of its subpackages if they exist).  I couldn't find a way to say "hey, this applies to the package *and* its subpackages".

The reason I wanted to do it is because, as you can see now, `getLdSoConfDLibPaths` is being invoked for the package and its subpackages, and it iterates over all files from them, every time.  This is suboptimal.

Something else I thought about is that we could have a global flag to signal whether `getLdSoConfDLibPaths` was already invoked.  But then I have another question: is it guaranteed that we will be dealing with only one package (and its subpackages)?

Sorry for the barrage of questions and thoughts; this code is new to me and I'm trying to make sense of how to best implement what I need here.  Thanks!

jonjohnsonjr: I think this is probably fine for now to ignore it, but either put a big scary TODO on that or plumb it around all the places it needs to go (ew). Ideally we'd have some kind of struct for this.

pkg/sca/sca.go
```diff
@@ -86,6 +87,84 @@ func isInDir(path string, dirs []string) bool {
...
+
+	log.Info("scanning for ld.so.conf.d files...")
+
+	for _, pkgName := range targetPackageNames {
+		fsys, err := hdl.FilesystemForRelative(pkgName)
+		if err != nil {
+			return nil, err
+		}
+
+		if err := fs.WalkDir(fsys, ".", func(path string, d fs.DirEntry, err error) error {
```
jonjohnsonjr, line 110: I think we can just `etc/ld.so.conf.d` instead of `.` and skip the isInDir.

sergiodj: Thanks.  This has now been implemented.

pkg/sca/sca.go
sergiodj: ACK.  I decided to plumb the variable around.

jonjohnsonjr (approved):

PR #1863, opened 2025-03-20
linter:usrmerge: Improve error message

justinvreeland:


pkg/linter/linter.go
```diff
@@ -794,7 +794,7 @@ func usrmergeLinter(ctx context.Context, _ *config.Configuration, _ string, fsys
 		// without special casing it with the package name.
 		if path == "sbin" || path == "bin" {
 			if d.IsDir() || d.Type().IsRegular() {
-				return fmt.Errorf("package contains non-symlink file at /sbin or /bin in violation of usrmerge")
+				return fmt.Errorf("package contains non-symlink file: (%s) at /sbin or /bin in violation of usrmerge", path)
```
smoser, line 797: this is strictly better, but it will only print out 'sbin' or 'bin'.
i'd like to see the files that are under there if there are any.

smoser: I think you can't even get to the error below which could be the one that actually shows the errors

justinvreeland: There already is for the paths just not for directories:

https://github.com/chainguard-dev/melange/blob/31783c28ba912a523aed645c5c7128ea40fe5950/pkg/linter/linter.go#L804

smoser (approved):
I think the code on lines 803-810 is effectively unreachable .  unless WalkDir searches depth-first.  

But this is better than it was.

I would prefer a list of the things that are bad.  But Justin says this is better.

justinvreeland:
Most of our other linters exit on the first error they find and I don't see a benefit to listing the files beyond the first one you encounter since the remedy is the same. I'll check if that's unreachable and probably remove that section if it is.

justinvreeland:
need to remove the schema updates and such that snuck in. idk why these aren't update in the pr that should've changed them.

smoser:
i think this looks fine. the only question, from reading this I _think_ that the commit message is out dated.

you would not see directories listed, right? only non-directories.  Ie, in the following no usr/sbin or bin or usr/sbin in the beolow.


```
Package contains paths in violation of usrmerge:
bin
bin/test
sbin
sbin/test
usr/sbin
usr/sbin/test
```

justinvreeland:
> i think this looks fine. the only question, from reading this I think that the commit message is out dated.

Ahh yeah I need to update that or fix it in the merge command.


> in the following no usr/sbin or bin or usr/sbin in the beolow.

Every other line is a directory in that snippet?

justinvreeland:
Oh of course I updated the commit message not the github summary dislike

smoser (approved):

PR #1868, opened 2025-03-25
git-checkout - set depth=-1 when checking out a branch with commit.

smoser:
Given a git-checkout stanza like this:

    - uses: git-checkout with: repository: https://github.com/owasp-modsecurity/ModSecurity-nginx branch: master expected-commit: 0b4f0cf38502f34a30c8543039f345cfc075670d

We were guaranteed failure when the 'master' branch received a new commit.  That is because 'depth' was not specified so it would default to 1 and the checkout would not be able to checkout the provided commit.

The fix for the user would be to add `depth: -1` so that the full history would be obtained and the expected commit could be checked out.

Rather than force the user to do the right thing, just do it for them.

An example failure is mentioned
https://github.com/wolfi-dev/os/issues/47661


smoser:
Another option here is to just fail in the case that I'm fixing.  That will cause build failures, but only on things that pretty much _do_ need to be fixed.

smoser:
@justinvreeland , Thanks for the review.

> That is a really annoying scenario. I'd like the response to be to fail and fix the file to work with tag

the primary reason for 'branch' is that there _is_ no upstream tag, as in upstream doesn't do releases at all or they havent in a long time and you need something newer.

> buts that's clearly not always an option.
> I guess the downside here is that we won't get notified about updates through failures.

Yeah, i called that out... but the thing is movement on the upstream might not be relevant or desired.
See https://github.com/wolfi-dev/os/pull/48210/ and [ingress-nginx-controller-1.12](https://github.com/wolfi-dev/os/blob/951cf111ce82f8c34d188a2c3469059d0a39c14b/ingress-nginx-controller-1.12.yaml) which motivated this PR. The relevant checkout is the second one, something that is needed as a dependency of the primary package.  Its not clear at all when you should bump the version (hash) of the secondary one.

Again, thanks for the thoughtful review.

PR #1867, opened 2025-03-24
Store and apply xattrs correctly for all builds

egibs:
#1836 allowed xattrs to work for QEMU but broke xattrs for Bubblewrap and Docker. This PR stores any xattrs in the workspace directory prior to the in-memory filesystem being created and then applies them to the new filesystem afterward.

I tested this with all three runners and it seemed to work, though QEMU errored during extraction which is expected:
```
usr/sbin/fping
tar: setxattrat: Cannot set 'security.capability' extended attribute for file 'usr/sbin/fping': Operation not supported
```

This PR does not fix `getcap` tests in pipelines, but I'm not sure if that ever worked regularly or not. For example the fping `getcap` test shows nothing on `v0.22.2` prior to the QEMU changes:
```
2025/03/24 22:38:25 INFO running step "Test file capabilities"
2025/03/24 22:38:25 WARN + '[' -d /home/build ]
2025/03/24 22:38:25 WARN + cd /home/build
2025/03/24 22:38:25 WARN + getcap /usr/sbin/fping
2025/03/24 22:38:25 WARN + exit 0
```

That said, it does seem to work in QEMU with the current changes (which is the exact situation builds were in prior to this PR):
```
2025/03/24 17:43:17 INFO running step "Test file capabilities"
2025/03/24 17:43:17 WARN + '[' -d /home/build ]
2025/03/24 17:43:17 WARN + cd /home/build
2025/03/24 17:43:17 WARN + getcap /usr/sbin/fping
2025/03/24 17:43:17 WARN + exit 0
2025/03/24 17:43:17 INFO /usr/sbin/fping cap_net_raw=ep
```

PR #1871, opened 2025-03-26
Implement versioned `provides`/`depends` for shared libraries

sergiodj:
This PR implements versioned `provides`/`depends` for shared libraries.

I've had to plumb some things around to make everything work, so I'd love to have another pair of eyes go through the changes to double check I didn't mess anything up.

The implementation for versioned `provides` is somewhat simple: we already have the package version being analyzed during SCA, and we just need to use it when generating the `provides = so:...` line.

Versioned `depends` is more involved.  The general idea (kudos to @jonjohnsonjr for the invaluable input!) is the following:

- First, we have to acquire a `PkgResolver` from apko.  We do that during `buildGuest`, where we can access the current `APK` being built.  This `PkgResolver` becomes an attribute of the `Build` struct.
- After that, we have to plumb this `PkgResolver` into `SCAHandle` so that we can access it during SCA.
- During SCA, while analyzing the program(s) to determine which shared libraries it depends on, we need to:
  - Ask `PkgResolver` to return the list of packages that satisfy the query `so:<ShlibName>`.
  - Verify whether any of the packages returned is listed as being installed in the build environment.  The package *and* its version need to match.
  - If there's a match, verify whether the package has a versioned `provides` for the shared library.  If it doesn't, then we can't (yet) use versioned `depends`.

The idea is that the algorithm works with and without versioned `provides` for shared libraries, so that we won't need to rebuild the whole archive after deploying it.

This code requires the changes implemented in https://github.com/chainguard-dev/apko/commit/620f8080935e57aa9d7ffab4788774779caf064a.

### Functional Changes

Notes:

### SCA Changes

Notes:


dannf:
I went through each commit and they all are clearly described and make sense to me. I've been learning golang this week, and I feel like I even understood all of the code!

sergiodj:
This is now ready for review.  Thanks @jonjohnsonjr for the help with the linter failures!

It's also worth providing an update on some tests I did locally.  There was a concern that the following scenario could lead to problems with the patch:

- A package `foo` that builds a shared library `libfoo.so`.
- A subpackage `libfoo` that ships said library.
- Another subpackage, `foo-bin`, that links against said library.

I implemented a solution for this case: when we're iterating over the list of installed packages and deciding which ones satisfy the shared library dependency, we first check if the possible candidate is a package that's being currently built (they're marked with the `@CURRENT` version).  If it is, then we return the version being currently built.

Please let me know if there are other cases which you think might be problematic.

sergiodj:
@xnox you might be interested in this PR.

sergiodj:
Hm, the CI is failing when trying to install a glibc built with this new melange.

According to the logs, the failure is happening here:

```
docker run --rm -v $(pwd):/work --workdir /work cgr.dev/chainguard/wolfi-base apk add --allow-untrusted packages/x86_64/glibc-*.apk
```

I built a glibc locally with the patched melange, and when I try to install its packages everything works.  I verified that I'm using the same `wolfi-base` image hash as the one used in the test.

I'll try cleaning my environment and redoing everything, but it's getting late here and I'll probably continue investigating this issue only tomorrow.

sergiodj:
Ah, of course, the problem stems from the fact that the container is mounted without setting up `packages/x86_64/` as a repository.

```
$ docker run --pull=always --rm -it --mount type=bind,source="/home/user/os/packages",destination="/work/packages",readonly -w "/work/packages" cgr.dev/chainguard/wolfi-base:latest
# apk add --allow-untrusted x86_64/glibc-*.apk
fetch https://packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz
ERROR: unable to select packages:
  ld-linux-2.40-r40:
    breaks: glibc-2.40-r40[so:ld-linux-x86-64.so.2=2.40-r40]
    satisfies: world[ld-linux=2.40-r40] bash-5.2.37-r30[so:ld-linux-x86-64.so.2] busybox-1.37.0-r40[so:ld-linux-x86-64.so.2] glibc-iconv-2.40-r40[so:ld-linux-x86-64.so.2] glibc-2.40-r40[so:ld-linux-x86-64.so.2]
               ncurses-6.5_p20241228-r1[so:ld-linux-x86-64.so.2] nss-db-2.40-r40[so:ld-linux-x86-64.so.2] apk-tools-2.14.10-r2[so:ld-linux-x86-64.so.2] libgcc-14.2.0-r11[so:ld-linux-x86-64.so.2] glibc-tracing-2.40-r40[so:ld-linux-x86-64.so.2]
  nss-db-2.40-r40:
    breaks: glibc-dev-2.40-r40[so:libnss_db.so.2=2.40-r40]
  nss-hesiod-2.40-r40:
    breaks: glibc-dev-2.40-r40[so:libnss_hesiod.so.2=2.40-r40]
```

And the installation fails because the new glibc has a versioned dependency on `so:ld-linux-x86-64.so.2=2.40-r40`.  Since nothing in the index provides this specific library, `apk` just bails out.

We may have to special-case `ld-linux.so` and never version its `provides` / `depends`.  I have to think more about this.  Now it's *really* late.

xnox:
previously when we changed or introduced provides/depends; we have always staged it to generate the new provides first; and then later start generating new depends. As it is much easier to test and roll out.

Also e2e tests are then all green during step 1 and step 2 of rollouts.

sergiodj:
> previously when we changed or introduced provides/depends; we have always staged it to generate the new provides first; and then later start generating new depends. As it is much easier to test and roll out.
> 
> Also e2e tests are then all green during step 1 and step 2 of rollouts.

Thanks for the suggestion.

That makes sense, but I wonder what the deadline for rolling out phase 2 would be in this case.  And since the code is working pretty much for every other case, maybe it's OK to just do it in a single step?

I took the liberty to refactor the new `determineShlibVersion` function and:

- Special-case `ld-linux.so`, as mentioned.
- Improve the version matching for installed vs. candidate packages (it should be more robust now).

However, let me take a step back and think about the current issue that's preventing the tests to pass.  I don't know it it's a scenario that we want to cover, so I'd welcome more input.

As I explained in the comments above, the problem is happening because the new glibc has versioned `depends` on its own shared libraries, but the packages are not part of `APKINDEX` and therefore doing an `apk add --allow-untrusted glibc-*.apk` won't currently work.

I am not an expert on how apk resolves dependencies when installing multiple package files; Debian's apt will do the right thing as long as you provide all the necessary packages in the command line.  I think apk is less smart in that it will try to install the packages in the order they're provided via the command line.

Anyway, as you can see in the logs the test is now failing with:

```
ERROR: unable to select packages:
  nss-db-2.40-r40:
    breaks: glibc-dev-2.40-r40[so:libnss_db.so.2>=2.40-r40]
  nss-hesiod-2.40-r40:
    breaks: glibc-dev-2.40-r40[so:libnss_hesiod.so.2>=2.40-r40]
```

This is happening after treating `ld-linux.so` as a special case.  I don't like the idea of special-casing everything glibc-related, and I don't think that's the right solution here.

Unless I'm missing something here (which is quite possible, given my lack of knowledge about the whole apk ecosystem), doing the melange modification in phases (as @xnox suggested) still wouldn't be enough to address this situation.  For example, even if we wait until all packages have versioned `provides` before pulling the trigger on the versioned `depends` part, we'd still have this same problem if we rebuilt glibc with an epoch bump and then tried to install its packages without them being present in `APKINDEX`.

Thoughts?

sergiodj:
@xnox Hi.  Any thoughts on the issue outlined in my last comment?

Thanks!

justinvreeland:
Also can we rebase merge this PR instead of squash merging it. I'd like to keep the independent commits to look through more easily.

xnox:
The CI check is just plain buggy:
```
docker run --rm -v $(pwd):/work --workdir /work cgr.dev/chainguard/wolfi-base apk add --allow-untrusted packages/x86_64/glibc-*.apk
```

Note that not all subpackages start with ${{package.name}} due to historical raisins and conventions. Our other installability check (the one in ci-apk-add) is much better w.r.t. this.

sergiodj:
> ```
> //
> // - Asking PkgResolver to calculate the list of packages that provide "so:shlib".
> //
> // - Verifying if any package in that list matches what is currently
> //   installed in the build environment.  The package can offer a
> //   provided shared library whose version is equal or greater than
> //   the one the build requires.
> //
> // - Verifying if the matched package has a versioned shared library
> //   "provides:".
> //
> // If all steps succeed, the package version is returned.
> ```
> 
> If I'm reading this correctly we shouldn't encounter issues with upgrades because we should never add the depends if there's not a matching provides in the repo already. Neat.

That's correct.

> Philosophically I don't like that we're version .so's with package versions. It looks like we have some that are moving backwards in version:
> 
> ```
> 				"so:libtiff-91af027d.so.6.0.2=6.0.2",
> 				"so:libtiff-91af027d.so.6.0.2=0.13.2-r0",
> ```
> 
> We don't have a real epoch, our epoch is revision how is this going to resolve when it's released will everyone be stuck on the "old" higher version'd library?

Great question.

Let me run some tests here.

justinvreeland:
> If I'm understanding correctly (and also based on what @xnox said), the CI issues I'm seeing are actually problems with the CI itself, and not with my proposed changes.

It is an issue with the CI not the PR but when possible we'd like the tests pass as part of the PR and the code causing problems I believe is in this repo.

Ideally it would be as easy add adding: ` --repository ./packages` to the lines in `.github/workflows/wolfi-presubmit.yaml`.   It's just super annoying to make changes and test workflows imo.

If we add it in this PR it won't run for _this_ PR if I understand workflows correctly?  So I guess we can fix it in a PR afterwards or try to land a fix before, but testing workflows is also a bit of a pain particularly when it wouldn't be using the melange code it has objections too. I don't feel the need to block though if everything else is fine.

> It seems like we have covered all the bases here, then? Earlier today I was thinking that it would be nice if we were able to perform archive-wide rebuilds using a customized melange, but it doesn't seem to be possible?

Not that I know of, I'd love to have it though. : (

> Either way, are there any other concerns that could prevent this PR to be merged?

Just want to know about the .so version numbers which you've already said you're testing.

justinvreeland:
Oh it looks like it ran when they added the new docker test so we should be able to add the CI change here and have it all work? https://github.com/chainguard-dev/melange/pull/1867

pkg/sca/sca.go
```diff
@@ -307,7 +455,14 @@ func processSymlinkSo(ctx context.Context, hdl SCAHandle, path string, generated
...
-			generated.Runtime = append(generated.Runtime, fmt.Sprintf("so:%s", soname))
+			shlibVer, err := determineShlibVersion(ctx, hdl, soname)
+			if err != nil {
+				return err
+			}
+			if shlibVer != "" {
+				shlibVer = ">="+shlibVer
+			}
+			generated.Runtime = append(generated.Runtime, fmt.Sprintf("so:%s%s", soname, shlibVer))
 		}
```
xnox, line 466: I would like us to a runtime selector to use these new style depends as opt-in; deploy new melange; opt-into new style depends for a few packages; canary them; and if all good; flip the default; keep opt-out; canary; remove the opt-out.

melange deployments take a while to propagate into staging & production build environments; and once out and if none of this works for us; or there are incorrect cornercases; we need to have a way to revert out of this.

Look at the existing options "no-provides", "no-depends", "no-commnads". I think we want a similar packageoption for this work.

Call the new packageoption "versioned-so-deps", make it default to false.
Release melange; make it propagate into wolfict & elastic builders; tag a few packages with package option "versioned-so-deps: true" things that have lots of deps, for example ffmpeg & curl and all of their leaf shared library dependencies (as in codecs, ldap, kerberos, etc libraries).

If all good, we can cut a new melange release that makes it the default.

Then if anything breaks for any packages, we can tag them with `versioned-so-deps:false` to get that individual build back to previous style of deps.

Once we rebuild everything, and or the % of things is high including niche things (lots of vendored deps, non-standard library paths, dlopens, nvidia). we can retire/remove this buildoption if no users of opt-out remain at that point.

This does mean keeping the old code path, and make the two codepaths conditional on setting.

Changes to provides seem to be harmless, as the value after = is currently unused, and thus it shouldn't matter what it is set to. But any incorrect versioned depends could lead to intallation failures.

This is out of abundance of caution really, and creating an escape hatch in case we need it.

Withdrawal playbook will need updating; previously withdrawing packages mostly kept the whole archive installable, and previously we would not do a mass wolfictl bump on reverse dependencies built against older versions, but now we will have to consider that.

.github/workflows/verify.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@2226d7cb06a077cd73e56eedd38eecad18e5d837 # v3.1.0
         with:
-          version: v1.60
+          version: v1.64
           args: --timeout=5m
```
xnox, line 33: this really ought to be fixed in separate prs by the dependabot and rebased away, as this is required but unrelated to mix in.

go.mod
```diff
@@ -1,17 +1,19 @@
...
 
-go 1.23.4
+go 1.24.0
+
+toolchain go1.24.1
 
 require (
-	chainguard.dev/apko v0.25.4
+	chainguard.dev/apko v0.25.6-0.20250328195433-4c14aaa28144
 	cloud.google.com/go/storage v1.51.0
```
xnox, line 9: we ought to cut new apko release; and bump melange to it; regardless of this pr.

such that ideally this pr doesn't touch upgrades of apko.

xnox (changes requested):

pkg/sca/sca.go
xnox: thinking more about this.

maybe opt-in canary is not needed; given we can test this extensively and we wouldn't land this if it doesn't work at all.

thus maybe opt-out canary would be enough.

justinvreeland: I am generally in favor of providing canaries for major melange changes particularly for sca changes that have impact like this. I also think opt out aligns better with our expectation this will roll out fine and we need a quicker revert than rolling back all melange everywhere.

sergiodj: Thanks for the details comments, folks.

I'm happy to implement an opt-out mechanism a la `no-depends`, and I agree it's a good idea.

@justinvreeland's question about .so version numbers has led me into a rabbit hole here, and I'm still not convinced that it works 100%.  That, and the fact that I'm also working on other stuff made me delay my reply to this PR, but I'm on it and will provide more details later.

xnox: @justinvreeland @sergiodj nothing can prepare us for libpcre soname versionioning -> https://ubuntu-archive-team.ubuntu.com/transitions/pcre3-to-pcre2.html#good,bad,partial,unknown,!notintesting which was called pcre had a few abi breaks and was up to soname 3, then got forked/restarted/reinvented as pcre2 and that become mature and stable and then everyone had to do a soname transition from 3 to 2. And it is confusing to this date, as certain pieces of software still only support and use the old 3, instead of the new 2. If and when this will happen to us, it still will be mostly ok, as we will only have one version in the build environment; and we would add extra additional runtime deps (or constraints) by explicit package name, rather than by SCA generate `so:` alone. And that's ok. as SCA is supposed to get us 90% there. I do not want to replicate the debian solution of doing a build once; on all arches; then gather symbol lists from build logs; by hand compile them into symbols lists; and then encode by-hand, on per-arch basis the min/max dependency versions and sonames each package contains. It's nice, they are able to generate very tight and correct versioned depends (i.e. python3 (>=3.8), python3 (<<3.12) => because they know they have bytecode and compiled extensions for those versions of pythons, and no-change rebuild can change those ranges), but we really don't need such precision and strict correctness. For us the goal here is to provide a simple strict floor, as by default we want to resolve everything to latest versions of the packages. Note that even with debian's very manual and precise constraints, when they choose to rebuild libraries and change ABI or drop symbols (intentionally, i.e. the 64-bit-time_t transition or a new no-X setting in openssl), they still had to throw all of this away, change package name, keep same library soname, and pretend life is wonderful (it is not). So on grand scheme of things, this approach is very safe and automated. It will create a good enough floor; it will not create a cap; and any further tweaks will be needed to be done by hand (which is status quo).

sergiodj: @xnox ACK, I understand that we're not aiming for perfect here.  Thanks for the further explanation.

Regarding @justinvreeland's question, I noticed that we will need a bit more tweaking than initially thought.  Currently, if a package generates e.g. `provides: so:libfoo.so.6=6`, and supposing that the package version if `4.0-r1`, then the new `provides` will be `provides: so:libfoo.so.6=4.0-r1`.

`apko` (and `apk`) will obviously sort the current `provides` as being more recent than the new one, which means that builds will be stuck with using the current `provides`.

My idea to address this issue is to patch `apko`'s resolver and make it sort everything that is *not* a package version (i.e., doesn't contain the `-r` substring) as *older* than everything that *is* a package version.  This would make sure that our new `provides` trick always gets sorted before the old `provides` with the soname version.

Today's been a bit hectic here but I'm working on an `apko` patch and will post it tomorrow.

xnox: i am a bit surprised it would pick things based on virtual provides; rather than the provider's package version.

sergiodj: Alright, https://github.com/chainguard-dev/apko/pull/1619 has been merged (thanks @jonjohnsonjr!).  I have also implemented the opt-out mechanism and force-pushed the branch.

I'm resolving this conversation.

pkg/sca/sca_test.go
```diff
@@ -152,12 +160,18 @@ func TestExecableSharedObjects(t *testing.T) {
 
 	want := config.Dependencies{
 		Runtime: []string{
+			"so-ver:ld-linux-aarch64.so.1",
+			"so-ver:libc.so.6",
+			"so-ver:libcap.so.2",
+			"so-ver:libpsx.so.2",
```
xnox, line 166: to fix abi coupling, wasn't we supposed to generate versioned depends here? As in depends on "so-ver:libc.so.6~2.41" ? (or >=2.41, or >=2.41-r0, or somesuch)

sergiodj: Yeah, this is still a WIP and I fixed it locally.  Let me force-push what I have.

sergiodj:
@xnox @justinvreeland The `so-ver:` idea has been implemented.  Tests are now passing (I had to do a minor adjustment to `removeSelfProvidedDeps` in order to make sure that `so-ver:` entries for vendored libraries are properly removed).

I'm thinking about reserving a beefy machine and performing a world rebuild of wolfi-os with this new melange, but I'd love to hear your feedback on the current implementation.  It would be great if I could drive this one to completion in the next week.

Thanks!

xnox:
> @xnox @justinvreeland The `so-ver:` idea has been implemented.  Tests are now passing (I had to do a minor adjustment to `removeSelfProvidedDeps` in order to make sure that `so-ver:` entries for vendored libraries are properly removed).
> 
> I'm thinking about reserving a beefy machine and performing a world rebuild of wolfi-os with this new melange, but I'd love to hear your feedback on the current implementation.  It would be great if I could drive this one to completion in the next week.
> 
> Thanks!

Landing new provides alone is safe.

Alternative is to add a feature flag via environment variable which activates new dependencies, land this, once new melange is in elastic builds we can change environment variables for staging cluster and request all staging builds to use this. That way you can do world rebuild with this enabled, without yet affected production builds. And if happy can flip production builds to it, revert if needed, and then remove this feature flag.

sergiodj:
@xnox I took some time to perform a partial world rebuild locally, and finally had the time now to address the corner cases I encountered.  I was able to build around 500 packages, and only found the two errors which have since been fixed.

I've also implemented a feature flag for the versioned shlib depends part of the PR.  If the environment variable `MELANGE_VERSIONED_SHLIB_DEPENDS` is set to a non-empty value, the feature will be enabled.

Let me know what you think.  Thanks!

pkg/sca/e2e_test.go
```diff
@@ -94,6 +94,25 @@ func TestAnalyze(t *testing.T) {
...
+                                "so-ver:libopenblas64_p-r0-0cf96a72.3.23.dev.so=0.13.2-r0",
+                                "so-ver:libopenjp2-eca49203.so.2.5.0=0.13.2-r0",
+                                "so-ver:libpng16-78d422d5.so.16.40.0=0.13.2-r0",
+                                "so-ver:libquadmath-96973f99.so.0.0.0=0.13.2-r0",
+                                "so-ver:libsharpyuv-20f78091.so.0.0.1=0.13.2-r0",
+                                "so-ver:libtiff-91af027d.so.6.0.2=0.13.2-r0",
+                                "so-ver:libwebp-850e2bec.so.7.1.8=0.13.2-r0",
+                                "so-ver:libwebpdemux-df9b36c7.so.2.0.14=0.13.2-r0",
+                                "so-ver:libwebpmux-9fe05867.so.3.0.13=0.13.2-r0",
+                                "so-ver:libxcb-f0538cc0.so.1.1.0=0.13.2-r0",
```
justinvreeland: Nit: some weird spacing here and for libnss_myhostname.

justinvreeland (approved):
Looks good. The environment variable should let us add the provides once melange and then test depends pieces locally once enough have rebuilt.

pkg/sca/e2e_test.go
sergiodj: Ops, fixed.

sergiodj:
> Looks good. The environment variable should let us add the provides once melange and then test depends pieces locally once enough have rebuilt.

Thanks, @justinvreeland.

I force-pushed a fix to the whitespace nit you pointed out.  I believe you will need to reapprove the PR.

Cheers.

justinvreeland (approved):

sergiodj:
> I wonder how this work interrcts the @javacruft work on renaming soname packages (as in positive synergy, negative divergence, or neutral and independent of each other)

I briefly mentioned that this change would likely align and benefit the soname package renaming.  Ultimately, I believe they're somewhat orthogonal but end up complementing each other.

Do you have something specific in mind that concerns you?

javacruft (approved):

PR #1875, opened 2025-03-27
npm/install: allow using pnpm instead

imjasonh:
If users want to use `npm/install` with `pnpm`, they can do so with `npm-package: pnpm` -- the existing behavior to use `npm` remains the default.

This change also adds `npm-install.yaml` and the new `pnpm-install.yaml` examples to our e2e rebuild test, which exposed a reproducibility-breaking change in `pnpm` when using `overrides` (disabled so it doesn't block the PR, just FYI)

To help debug rebuild tests, I refactored e2e.yaml a bit so it uploads the original and rebuilt APKs for local inspection.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -71,13 +71,15 @@ jobs:
       fail-fast: false
       matrix:
         cfg:
-          - examples/cargo-build.yaml
-          - examples/gnu-hello.yaml
-          - examples/go-build.yaml
-          - examples/minimal.yaml
-          - examples/npm-install.yaml
-          - examples/pnpm-install.yaml
-          - melange.yaml
+          # build and rebuild examples
+          - cargo-build.yaml
+          - gnu-hello.yaml
+          - go-build.yaml
+          - minimal.yaml
+          - npm-install.yaml
+          - pnpm-install.yaml
+
+          - melange.yaml # special; builds melange itself
 
     container:
       image: alpine:latest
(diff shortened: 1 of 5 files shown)


indexzero:
@imjasonh ✅🐢🚀✨ this is a great next step forward:

1. Doesn't require any changes to how `nodejs` is built 
2. Preserves existing calls / semantics around `needs` for `npm`
3. Safely allows folks to opt-in to `pnpm`

imjasonh:
It looks like pnpm causes some files to be non-reproducible:

```
diff \
  <(tar -Oxf original.apk usr/lib/node_modules/.modules.yaml) \
  <(tar -Oxf rebuilt.apk usr/lib/node_modules/.modules.yaml)
69c69
< prunedAt: Thu, 27 Mar 2025 15:45:16 GMT
---
> prunedAt: Thu, 27 Mar 2025 15:45:22 GMT
```

```
diff \
  <(tar -Oxf original.apk usr/lib/node_modules/.pnpm-workspace-state.json)\
  <(tar -Oxf rebuilt.apk usr/lib/node_modules/.pnpm-workspace-state.json)
2c2
<   "lastValidatedTimestamp": 1743090316258,
---
>   "lastValidatedTimestamp": 1743090322464,
```

Is there a way to tell `pnpm` not to set these, or to set them to `SOURCE_DATE_EPOCH` so that this doesn't produce irreproducible builds?

(Not a requirement for this PR, but it would be a soft blocker for rolling this out to more packages, since we don't want to regress reproducibility)

jonjohnsonjr (approved):

PR #1884, opened 2025-03-30
rebuild: don't require `pipelineDir` when rebuilding

imjasonh:
When a package is built with a pipeline defined in a `--pipeline-dir` (typically `./pipelines`), the original pipeline will be present in the compiled output. This makes compile handle this case so that we don't have to reuse the same pipelines dirs when rebuilding such packages.


imjasonh:
Ideally we shouldn't need these. Closing for now.

jonjohnsonjr (approved):

PR #1891, opened 2025-04-03
feat: Add support for secrets in the build environment

EyeCantCU:
This allows secrets to be passed to the build environment. Unlike env vars, secrets will not end up in the parsed melange config shipped with APKs

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -141,6 +141,7 @@ type Build struct {
 	DefaultTimeout        time.Duration
 	Auth                  map[string]options.Auth
 	IgnoreSignatures      bool
+	Secrets               []string
 
 	EnabledBuildOptions []string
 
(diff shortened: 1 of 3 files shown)


pkg/build/build.go
```diff
@@ -1227,6 +1228,20 @@ func (b *Build) buildWorkspaceConfig(ctx context.Context) *container.Config {
...
+		// Split secret into key value pair
+		parsedSecret := strings.Split(secret, "=")
+
+		// Append to environment if successfully split
+		if len(parsedSecret) == 2 {
+			k := parsedSecret[0]
+			v := parsedSecret[1]
+			cfg.Environment[k] = v
+		} else {
+			log.Errorf("could not parse secret: %s", parsedSecret)
```
luhring, line 1241: Since this variable likely contains a sensitive value (even if malformed), we probably don't want to log it

luhring (commented):
I know this is still in draft, so I apologize for the premature review! 🙇  Just wanted to suggest that we get a security review for this when it's ready for primetime, just to avoid any inadvertent vulnerabilities related to leaking secrets

pkg/build/build.go
EyeCantCU: That's a great point! I'll fail with the expected format instead

imjasonh:
I think we'll find a way to do without this. If we can't, we can reopen.

Thanks for working on this @EyeCantCU !

PR #1893, opened 2025-04-04
fix(pipelines/git-checkout): preserve ownership of existing files

maxgio92:
This commit adds explicit `--no-same-owner` to tar extraction of
the git repository to the destination directory, which in most
of the cases is the Melange workspace.
--no-same-owner option makes:
>Extract files as yourself (default for ordinary users).

which means that when the Melange runner is root, ownership
of existing files is overwritten when the option is not set.

When the git-checkout pipelines extracts the content of the
git repositories locally from the temporary directory to the
destination directory [1], which is usually the Melange pipeline
workspace (/home/build inside the runner container), does not
preserve ownership of the directories that already existed in
the destination directory.

When the runner user does not match the user who runs Melange,
this ends up in permission errors for all the operation on
the workspace that happen after the pipelines have run, which
means the user that runs Melange tries to write to the
workspace that no longer owns.

This is the case of Melange run by non-root with the Docker
runner, and operations like `xattrs` restore [2].

This commit makes the git repository content extraction to run
with `tar --extract --no-same-owner`, preserving the ownership of
the workspace and all the directories inside the workspace that
already existed, like '.' in the workspace, as tar runs with `-C
$destination`.

1. https://github.com/chainguard-dev/melange/blob/a5d28473b9091cf7a45a0864bb149ae9fddfff99/pkg/build/pipelines/git-checkout.yaml#L190
2.
https://github.com/chainguard-dev/melange/blob/a5d28473b9091cf7a45a0864bb149ae9fddfff99/pkg/build/build.go#L956

### Functional Changes

Notes:

### DIFF
--- a/e2e-tests/run-tests
+++ b/e2e-tests/run-tests
@@ -14,6 +14,7 @@ MELANGE=${MELANGE:-melange}
 if [ "$MELANGE" = "melange" ]; then
     MELANGE=$(which melange)
 fi
+ARGS=${ARGS:-""}
 
 echo "Testing with melange from $MELANGE"
 $MELANGE version --json
@@ -32,7 +33,7 @@ fi
 
 fails=""
 for yaml in "$@"; do
-  args=""
+  args="${ARGS}"
   ops=""
   base=${yaml%.yaml}
   case "$base" in
--- a/pkg/build/pipelines/git-checkout.yaml
+++ b/pkg/build/pipelines/git-checkout.yaml
@@ -187,7 +187,7 @@ pipeline:
 
           vr cd "$workdir"
           msg "tar -c . | tar -C \"$dest_fullpath\" -x"
-          ( tar -c . ; echo $? > "$rcfile") | tar -C "$dest_fullpath" -x
+          ( tar -c . ; echo $? > "$rcfile") | tar -C "$dest_fullpath" -x --no-same-owner
           read rc < "$rcfile" || fail "failed to read rc file"
           [ $rc -eq 0 ] || fail "tar creation in $workdir failed"
 
--- a/e2e-tests/git-checkout-build.yaml
+++ b/e2e-tests/git-checkout-build.yaml
@@ -262,7 +262,7 @@ pipeline:
       if [[ $runner_user == $melange_user ]]; then
         # Ownership of existing files match the ownership of files created within the runner
         # (i.e. by the git-checkout pipeline tar --extract command).
-        expected=48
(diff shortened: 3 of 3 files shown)


maxgio92:
I've just added a test specific for reproducibility of this change and verifying that exactly only the expected existing files have different ownership, between runner and Melange users.

I've tested it for the cases:
* Melange user root, Docker runner
* Melange user non-root, Docker runner
* Melange user root, Bubblewrap runner
* Melange user non-root, Bubblewrap runner

e2e-tests/run-tests
```diff
@@ -14,6 +14,7 @@ MELANGE=${MELANGE:-melange}
 if [ "$MELANGE" = "melange" ]; then
     MELANGE=$(which melange)
 fi
+ARGS=${ARGS:-""}
```
maxgio92, line 17: This allows to specify Melange extra options, like the runner, which is important for this change.

pkg/build/pipelines/git-checkout.yaml
```diff
@@ -187,7 +187,7 @@ pipeline:
 
           vr cd "$workdir"
           msg "tar -c . | tar -C \"$dest_fullpath\" -x"
-          ( tar -c . ; echo $? > "$rcfile") | tar -C "$dest_fullpath" -x
+          ( tar -c . ; echo $? > "$rcfile") | tar -C "$dest_fullpath" -x --no-same-owner
```
maxgio92, line 190: Because we're using Busybox, `--no-overwrite-dir` is not supported.

egibs: We had to install `gnutar` for the recent xattr work, maybe it makes sense to install it regardless of runner?

maxgio92: I'm a bit skeptical about introducing a further change to this pipeline as it's used across all Wolfi packages, but I don't have a strong opinion on that and I'm not against it.

89luca89: I'm ok for the switch to gnutar, it's posix compliant, so it won't break current compatibility, but opens up to more modern features

89luca89:
LGTM I'd consider the tar swtich like @egibs suggests, but that can be addressed in another PR so we don't amass too many changes in one PR :+1:

89luca89 (approved):

jonjohnsonjr (approved):

PR #1900, opened 2025-04-08
github: fixup presubmit workflow

xnox:
Bump epochs prior to building new packages, such that just rebuilt
packages are tested later on.

During apk installability tests, pass through the full registry of
packages such that they can resolve any new dependants, when needed.

This should fix CI when making invasive SCA changes.

### DIFF
--- a/.github/workflows/wolfi-presubmit.yaml
+++ b/.github/workflows/wolfi-presubmit.yaml
@@ -93,6 +93,12 @@ jobs:
           sudo chmod a+x /usr/bin/melange
           melange version
 
+      # this need to point to main to always get the latest action
+      - uses: wolfi-dev/actions/install-wolfictl@main # main
+
+      - run: |
+          wolfictl bump ${{ matrix.package }}
+
       - if: matrix.runner == 'bubblewrap'
         run: |
           sudo apt-get -y install bubblewrap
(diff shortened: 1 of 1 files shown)


justinvreeland (approved):
That was a bit more involved that I thought it'd be.

.github/workflows/wolfi-presubmit.yaml
```diff
@@ -157,15 +163,12 @@ jobs:
...
-      - uses: wolfi-dev/actions/install-wolfictl@main # main
-
       - name: Test installable and Scan for CVEs
         run: |
           if [[ "${{ matrix.package }}" == "fping" ]]; then
-            docker run --rm -v $(pwd):/work --workdir /work cgr.dev/chainguard/wolfi-base /bin/sh -c "apk add --allow-untrusted packages/x86_64/${{ matrix.package }}-*.apk; apk add libcap-utils; getcap /usr/sbin/fping"
+            docker run --rm -v $(pwd):/work --workdir /work cgr.dev/chainguard/wolfi-base /bin/sh -c "sed 's|=.*||' -i /etc/apk/world; apk add --allow-untrusted -X ./packages/ packages/x86_64/${{ matrix.package }}-*.apk; apk add libcap-utils; getcap /usr/sbin/fping"
           else
-            docker run --rm -v $(pwd):/work --workdir /work cgr.dev/chainguard/wolfi-base apk add --allow-untrusted packages/x86_64/${{ matrix.package }}-*.apk
+            docker run --rm -v $(pwd):/work --workdir /work cgr.dev/chainguard/wolfi-base /bin/sh -c "sed 's|=.*||' -i /etc/apk/world; apk add --allow-untrusted -X ./packages/ packages/x86_64/${{ matrix.package }}-*.apk"
```
smoser, line 171: You didn't insert this bug though, it was present before.

`packages/x86_64/${{matrix.package}}-*.apk`  won't necessarily match all packages built by package. As specific and relevant example, `glibc` builds package `ld-linux`.

I don't have an obvious solution other than just using `packages/x86_64/*.apk`. Poking through history of this workflow, it looks like we've probably never tested the installation of 'ld-linux'.

PR #1902, opened 2025-04-08
WIP: declaratively add capabilities

imjasonh:


pkg/config/config.go
```diff
@@ -126,6 +126,9 @@ type Package struct {
 	// records, if known.
 	CPE CPE `json:"cpe,omitempty" yaml:"cpe,omitempty"`
 
+	// Capabilities to set after the pipeline completes.
+	Capabilities []Capability `json:"capabilities,omitempty" yaml:"capabilities,omitempty"`
```
jonjohnsonjr, line 130: I think I would call these `setcap` and `SetCap` just to distinguish from Configuration.Capabilities.

egibs: Made that change in [`9abf964` (#1902)](https://github.com/chainguard-dev/melange/pull/1902/commits/9abf964e9a4d86d220341b91a21c9523ddf07d65).

I think we'll need to support the same `attr, data` pattern so I added that too.

imjasonh:
@egibs is going to take this over, closing for now.

PR #1905, opened 2025-04-09
Add license checks to package builds and an external license-check command

sil2100:
There is still much tidying to do with this branch, but I think it's good enough for some initial review. Please keep in mind Go was not my native language!

What this branch adds:
- License linting pkg/ package using licenseclassifier and go-licenses (for type classification). This package is used in a few different places internally
- At the end of each melange build, we now run a license check run that tries to list out all the found licenses and all the differences against the yaml configuration
- New `melange license-check` command is added to do the same, either on a yaml file, an apk package or source tree
- New renovate package that modifies the copyright section of a melange file configuration to add all the licenses listed out
- ...which can be accessed via `melange license-check --fix`
- Some tests for a few of the new modules. Still missing tests for the renovate bits - coming soon!


pkg/license/license.go
```diff
@@ -0,0 +1,329 @@
...
+			Source:     licensePath,
+		})
+	}
+
+	return licenses, nil
+}
+
+// FindLicenseFiles returns a list of license files in a directory, sorted by their relevance score.
+func FindLicenseFiles(fsys fs.FS) ([]LicenseFile, error) {
+	// Copied and adjusted from Ruby to Go from the "licensee" project
```
dannf, line 125: do we need to include the MIT license text?

pkg/license/license.go
```diff
@@ -0,0 +1,329 @@
...
+	var licenseFiles []LicenseFile
+	var ignore bool
+
+	err := fs.WalkDir(fsys, ".", func(filePath string, info fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.IsDir() {
+			return nil
+		}
```
dannf, line 162: do we need to be weary of symlinks? (recursion, or pointing to files outside of this tree)

pkg/license/license.go
```diff
@@ -0,0 +1,329 @@
...
+
+		// Let's ignore all files in the melange-out/ directory, as it's not part of the source
+		if strings.Contains(filePath, "melange-out") {
```
dannf, line 164: should we use `melangeOutputDirName`?

pkg/license/license.go
```diff
@@ -0,0 +1,329 @@
...
+			return nil
+		}
+
+		// Check if the file matches any of the license-related regex patterns
+		for regex, weight := range filenameRegexes {
+			if regex.MatchString(info.Name()) {
```
dannf, line 170: We could save on indention with:

```
if !regex.MatchString(info.Name()) {
  continue
}
// rest
```

pkg/license/license.go
```diff
@@ -0,0 +1,329 @@
...
+				for _, ext := range ignoredExt {
+					if ignore = strings.HasSuffix(info.Name(), ext); ignore {
```
dannf, line 174: right now `ext` is e.g. `xml` - should we prefix it with a `.` ?

pkg/license/license.go
```diff
@@ -0,0 +1,329 @@
...
+	// Sort the license files by their 'score' (weight)
+	sort.Slice(licenseFiles, func(i, j int) bool {
+		return licenseFiles[i].Weight > licenseFiles[j].Weight
+	})
+
+	return licenseFiles, nil
+}
+
+// melangeLicenseToLicense is a helper that converts a melange license info to a License struct.
+func melangeLicenseToLicense(license string, licensePath string) License {
```
dannf: fwiw, something like `buildLicense()` would be a more explanatory name for me. And do we need the `melange` prefix in the name?

pkg/license/license.go
```diff
@@ -0,0 +1,329 @@
...
+	log := clog.FromContext(ctx)
+
+	// Find all license-text files
+	licenseFiles, err := FindLicenseFiles(fsys)
+	if err != nil {
+		return nil, fmt.Errorf("finding license files: %w", err)
+	}
+
+	if len(licenseFiles) == 0 {
+		// No license files detected, no linting performed.
```
dannf, line 232: that might be a useful debug level message

pkg/license/license.go
```diff
@@ -0,0 +1,329 @@
...
+				detectedLicenses = append(detectedLicenses, l)
+			}
+		}
+	}
+
+	log.Infof("checking gathered license information against the configuration")
+
+	// Let's first turn the melange licensing information into a coherent licensing list, similar to what Identify returns
+	// We first start off by splitting license information that has OR and AND into separate license entries
+	// Every entry can have multiple AND or ORs
```
dannf: excellent comment

pkg/license/license.go
```diff
@@ -0,0 +1,329 @@
...
+			splitLicenses := regexp.MustCompile(`\s+( AND | OR )\s+`).Split(ml.License, -1)
+			for _, sl := range splitLicenses {
+				melangeLicenses = append(melangeLicenses, melangeLicenseToLicense(sl, ml.LicensePath))
+			}
+		} else {
+			melangeLicenses = append(melangeLicenses, melangeLicenseToLicense(ml.License, ml.LicensePath))
+		}
+	}
+
+	// Now let's check if the detected licenses are in the configuration
```
dannf: this feels like a possible place to split this function in 2 for readability

pkg/license/license_test.go
```diff
@@ -0,0 +1,209 @@
...
+	"path/filepath"
+	"testing"
+
+	apkofs "chainguard.dev/apko/pkg/apk/fs"
+	"chainguard.dev/melange/pkg/config"
+)
+
+func TestFindLicenseFiles(t *testing.T) {
+	// Create a temporary directory for testing
+	tempDir, err := os.MkdirTemp("", "license_test")
```
dannf, line 30: could we use `t.TempDir()` here?

pkg/license/license.go
```diff
@@ -0,0 +1,329 @@
...
+	log.Infof("checking gathered license information against the configuration")
+
+	// Let's first turn the melange licensing information into a coherent licensing list, similar to what Identify returns
+	// We first start off by splitting license information that has OR and AND into separate license entries
+	// Every entry can have multiple AND or ORs
+	melangeLicenses := []License{}
+	for _, ml := range cfg.Package.Copyright {
+		if strings.Contains(ml.License, " OR ") || strings.Contains(ml.License, " AND ") {
+			// Split the license into separate entries using regexp
+			splitLicenses := regexp.MustCompile(`\s+( AND | OR )\s+`).Split(ml.License, -1)
```
dannf: I assume capitalization is required for AND and OR?

pkg/license/license_test.go
```diff
@@ -0,0 +1,209 @@
...
+		for _, diff := range diffs {
+			if diff.Path == expected.Path && diff.Is == expected.Is && diff.Should == expected.Should {
+				found = true
+				break
+			}
+		}
+		if !found {
+			t.Errorf("Expected license difference %+v not found", expected)
+		}
+	}
```
dannf, line 209: maybe add tests for the AND/OR constructs?

dannf (commented):
newbie go developer comments inline. this all seems clear and easy to follow - i feel good about being able to jump in and fix/add things. nice job @sil2100 !

pkg/license/license.go
jonjohnsonjr: if so, might make sense to move this function into a separate file to make the attribution clear

pkg/license/license.go
```diff
@@ -0,0 +1,386 @@
...
+			}
+		}
+
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Sort the license files by their 'score' (weight)
```
jonjohnsonjr, line 202: Can weights ever be tied? If so, should we have some tiebreaker field? Alternatively use `sort.SliceStable`?

pkg/license/license.go
```diff
@@ -0,0 +1,386 @@
...
+		if isLicenseMatchConfident(dl) {
+			s = " ignored"
+		}
+		log.Infof("  %s: %s (%f%s) (%s)", dl.Source, dl.Name, dl.Confidence, s, dl.Type)
+	}
+
+	// TODO: Handle low-confidence licenses, possibly by printing out info about those separately!
+
+	diffs := []LicenseDiff{}
+	if cfg != nil {
```
jonjohnsonjr, line 290: The rest of this function might benefit from early returns (or pulling some out into a separate function) to undo some of the deep nesting?

pkg/license/license.go
sil2100: That's a great point! I think we should simply ignore symlinks altogether as it's a valid security concern. WalkDir walks through all the paths in the whole tree anyway.

pkg/license/license.go
sil2100: Excellent idea! Even split it into 3

pkg/license/license_test.go
sil2100: Whoops, missed that one! I converted many test cases from my earlier usage of MkdirTemp, but missed this one.

pkg/license/license.go
sil2100: That is correct!

pkg/license/license_test.go
sil2100: One of these is already sort-of handled in `TestLicenseCheck`. There is one "GPL-2.0 OR GPL-3.0" input there. But I might just test the separated out function now.

pkg/cli/license_check.go
```diff
@@ -0,0 +1,92 @@
...
+	"chainguard.dev/melange/pkg/renovate"
+	"chainguard.dev/melange/pkg/renovate/copyright"
+	"chainguard.dev/melange/pkg/source"
+)
+
+func licenseCheck() *cobra.Command {
+	var workDir string
+	var fix bool
+	cmd := &cobra.Command{
+		Use:     "license-check",
```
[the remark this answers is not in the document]
sil2100: Doing that! I looked at the other commands and didn't see them specifying `file` or any other input via the `Use:` field, so I didn't know!

pkg/license/license.go
sil2100: As `melangeOutputDirName` is defined in the build package, and the build package is using the license package, we'd have to move it somewhere out to not introduce a circular dependency (and make it exported). I think this is a valid thing to do, but I might tackle it a bit later.

pkg/license/license.go
sil2100: I named it this way since that helper is only used when we convert melange configuration's licensing information to the `License` type. I would be worried that `buildLicense()` would give the impression that this is a function that builds the License type itself, while it's meant to be a quick helper that simply fills out some default values for fields we don't really have in melange. I'll think of maybe hm, just making one helper for building any `License` struct, but was worried as we'd need to make some function arguments optional.

sil2100: Actually, removed the function altogether! I think now it's actually better!

sil2100:
Thank you for the review comments, they were excellent! Hopefully addressed most of them :)

jonjohnsonjr (approved):

pkg/license/license_test.go
dannf: oh, i see that now!

PR #1910, opened 2025-04-11
pipelines: cmake/build: Use verbose mode by default

dannf:
This will make it easier to debug failures.


murraybd:
I think surfacing as much information as possible about failures is quite helpful when debugging it saves people time from having to redo the build after modifying options.

We've made similar changes to the Makefiles so that make test always runs with --debug.

EyeCantCU (approved):

PR #1941, opened 2025-04-22
Revert "fix: propagate Range field of subpackages (#1939)"

jonjohnsonjr:
This reverts commit 8f00e845fa17e610b8135480725db80c22380df7.

### DIFF
--- a/pkg/build/build_test.go
+++ b/pkg/build/build_test.go
@@ -59,26 +59,22 @@ func TestConfiguration_Load(t *testing.T) {
 					},
 				},
 				Subpackages: []config.Subpackage{{
-					Name:  "cats",
-					Range: "animals",
+					Name: "cats",
 					Pipeline: []config.Pipeline{{
 						Runs: "cats are angry",
 					}},
 				}, {
-					Name:  "dogs",
-					Range: "animals",
+					Name: "dogs",
 					Pipeline: []config.Pipeline{{
 						Runs: "dogs are loyal",
 					}},
 				}, {
-					Name:  "turtles",
-					Range: "animals",
+					Name: "turtles",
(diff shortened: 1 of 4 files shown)


hectorj2f (dismissed):
Please, add a comment to the Range field explaining why it is empty and not parsed.

hectorj2f:
@jonjohnsonjr Could you add `yaml:"-"` to reflect that we don't care about that field ?

jonjohnsonjr:
> Could you add yaml:"-" to reflect that we don't care about that field ?

I'd rather not to keep this revert clean. You are welcome to send a follow-up PR.

hectorj2f:
Could you explain why we don't need the reverted PR? That would help to give some context to everyone and understand the initial purpose of having this field.

jonjohnsonjr:
> Could you explain why we don't need the reverted PR? That would help to give some context to everyone and understand the initial purpose of having this field.

I think the burden of proof for keeping this change is on you. The purpose of this field is described in the code:

> // Optional: The iterable used to generate multiple subpackages

This is meaningful in the yaml as an _input_ to ParseConfiguration. It doesn't make sense for this to exist in the generated subpackages as an _output_ of ParseConfiguration. There's not much that you could meaningfully do with it, and I'm worried about how this will interact with e.g. `melange rebuild` because I suspect in a subsequent change you are going to realize that you need to "fix" [this TODO](https://github.com/chainguard-dev/melange/blob/6edbfd0d0c69d0604586bbdf1bf641c4451aa4a7/pkg/config/config.go#L1482) in order to meaningfully use the `Range` field, which will cause `melange rebuild` to ~forkbomb.

imjasonh (approved):

PR #1946, opened 2025-04-24
Set acpi=off for micro VMs; attempt to use micro VMs by default

egibs:
@smoser noted that setting `acpi=off` resolved the issues alluded to in #1945. 

Given that, this PR adds that option in the `-machine` args and now defaults to using a microVM as opposed to what was previously defined in the aforementioned PR. 

The package-level config was also not ideal so this behavior is now opt-out solely via `QEMU_USE_MICROVM`.

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -507,6 +507,8 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 
 	// we need to fallback to -machine virt, if not machine has been specified
 	if !useVM {
+		// aarch64 supports virt machine type, let's use that if we're on it, else
+		// if we're on x86 arch, but without microvm machine type, let's go to q35
 		log.Infof("qemu: not attempting to use a microVM")
 		if cfg.Arch.ToAPK() == "aarch64" {
 			baseargs = append(baseargs, "-machine", "virt")
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -476,16 +476,18 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 	}
 
 	baseargs := []string{}
-	microvm := false
-	// by default, cfg.MicroVM will be false
-	// override the MicroVM config value with the environment variable if present
-	// and said variable is parseable as a bool
-	if env, ok := os.LookupEnv("QEMU_USE_MICROVM"); ok {
-		if val, err := strconv.ParseBool(env); err == nil {
-			cfg.MicroVM = val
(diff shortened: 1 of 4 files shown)


pkg/container/qemu_runner.go
```diff
@@ -495,18 +497,17 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 		} {
 			if _, err := os.Stat(p); err == nil && cfg.Arch.ToAPK() != "aarch64" {
 				// only enable pcie for network, enable RTC for kernel, disable i8254PIT, i8259PIC and serial port
-				baseargs = append(baseargs, "-machine", "microvm,rtc=on,pcie=on,pit=off,pic=off,isa-serial=off")
+				baseargs = append(baseargs, "-machine", "microvm,rtc=on,pcie=on,pit=off,pic=off,isa-serial=off,acpi=off")
```
smoser, line 500: for me, `acpi=off` worked as a kernel parameter. i didn't actually test as a machine arg.

smoser: yeah, i dont think this works here.
if i add acpi=off as part of the -machine args, i get:
```
$ qemu-system-x86_64 
  -machine microvm,rtc=on,pcie=on,pit=off,pic=off,acpi=off 
  -kernel ./vmlinuz -bios ./ubuntu-bios-microvm.bin 
  -initrd ./initramfs.cpio 
  -m 4092069k -smp 8 -accel kvm -cpu host -display none 
  -no-reboot -no-user-config -nodefaults -parallel none 
  -serial mon:stdio -echr 0x05 
  -device virtio-rng-pci,rng=rng0 
  -object rng-random,filename=/dev/urandom,id=rng0 
  -append "console=ttyS0 nomodeset panic=-1 sshkey=blob" 
  -fsdev local,security_model=mapped,id=fsdev100,path=./workspace 
  -device virtio-9p-pci,id=fs100,fsdev=fsdev100,mount_tag=defaultshare 
  -object iothread,id=io1 
  -device virtio-blk-pci,drive=disk0,iothread=io1 
  -drive if=none,id=disk0,cache=none,format=raw,aio=threads,werror=report,rerror=report,file=./disk.img

qemu-system-x86_64: -device virtio-rng-pci,rng=rng0: No 'PCI' bus found for device 'virtio-rng-pci'

```

egibs: Ah, yea, I misunderstood. I'm fine with shipping #1945 while we follow-up on this.

smoser: I think the variable-only change is nice.

PR #1957, opened 2025-05-01
QEMU runner tweaks

egibs:
This PR makes a few tweaks to the QEMU runner code, namely:
- Increased memory usage instead of capping at 1/4th of the system's memory (we can go lower than 85%, too)
- Relying on a single way to specify microVM usage
- Pedantic change for units

TBD if we can land a fix for using QEMU on baremetal ARM instances.

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -1070,38 +1070,31 @@ func generateSSHKeys(ctx context.Context, cfg *Config) ([]byte, error) {
 	return ssh.MarshalAuthorizedKey(publicKey), nil
 }
 
+// Use binary values for all units
 const (
-	_   = iota
-	KiB = 1 << (10 * iota)
-	MiB
-	GiB
-	TiB
-)
-
-const (
-	B  int64 = 1
-	KB int64 = B * 1000
-	MB int64 = KB * 1000
-	GB int64 = MB * 1000
-	TB int64 = GB * 1000
+	_  = iota
+	KB = 1 << (10 * iota)
+	MB
+	GB
+	TB
 )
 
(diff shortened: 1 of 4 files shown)


pkg/container/qemu_runner.go
k4leung4, line 1: should there be unit test for this?

egibs: Yes. We have the tests [here](https://github.com/chainguard-dev/melange/blob/main/.github/workflows/wolfi-presubmit.yaml) which will use the QEMU runner but nothing akin to the Bubblewrap/Docker runner tests.

89luca89 (approved):

pkg/container/qemu_runner.go
```diff
@@ -1117,25 +1131,50 @@ func convertHumanToKB(memory string) (int64, error) {
...
+		if len(parts) != 2 {
+			return mem
+		}
+
+		memsize, err := strconv.ParseInt(parts[1], 10, 64)
+		if err != nil {
+			return mem
 		}
+		// hw.memsize returns the memory size in bytes
+		return int(memsize / 1024)
```
89luca89, line 1177: You're using 1000 based units for conversion but here we do 1024, let's change this?

89luca89 (changes requested):

pkg/container/qemu_runner.go
egibs: Funnily enough, using 1000 causes QEMU to throw exceptions. I don't know why that is.

egibs: I can revert to using 1024 everywhere.

89luca89 (approved):

PR #1959, opened 2025-05-02
fix(bump): Only update first git-checkout node with expected-commit

debasishbsws:
- Previously, all git-checkout nodes in a config were updated with the same commit SHA.
- This caused incorrect updates when multiple git-checkout steps exist, each referring to different repositories or commits.
- Now, only the first git-checkout node is updated to prevent partial or broken updates.
- Reduces incorrect automated updates and the need for manual intervention.

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

### DIFF
--- a/pkg/renovate/bump/bump.go
+++ b/pkg/renovate/bump/bump.go
@@ -140,7 +140,8 @@ func New(ctx context.Context, opts ...Option) renovate.Renovator {
 			RecurseNodes().
 			Filter(yit.WithMapValue("git-checkout"))
 
-		for gitCheckoutNode, ok := it(); ok; gitCheckoutNode, ok = it() {
+		// Only run updateGitCheckout once for the first git-checkout node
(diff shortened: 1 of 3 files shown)


pkg/renovate/bump/bump.go
```diff
@@ -140,7 +140,8 @@ func New(ctx context.Context, opts ...Option) renovate.Renovator {
 			RecurseNodes().
 			Filter(yit.WithMapValue("git-checkout"))
 
-		for gitCheckoutNode, ok := it(); ok; gitCheckoutNode, ok = it() {
+		// Only run updateGitCheckout once for the first git-checkout node
```
hectorj2f, line 143: This sounds like a limitation. I'd consider instead to request support for checking updates for any of the git-checkouts.

debasishbsws:
Closing as we already have: https://github.com/chainguard-dev/melange/pull/2008

PR #1961, opened 2025-05-04
Retrieve the full workspace.

mattmoor:
This changes the QEMU runner to retrieve the full workspace, not just `melange-out`.  This is because the license checker is failing to extract livense files from the workspace.

This also changes the `retrieveWorkspace` logic to allow overwriting files, since we otherwise see issues with files we seeded the workspace with prior to the build.

We see this using the QEMU runner on `db.yaml` and other builds with `license-path` set.


mattmoor:
cc @89luca89 FYI

89luca89 (approved):

PR #1962, opened 2025-05-05
Turn relative working-dir into absolute paths.

mattmoor:
The QEMU runner doesn't handle relative `working-directory` paths properly because unlike the bubblewrap runner it doesn't start each invocation by changing into `/home/build`.

Prior to this change `php-8.3-pecl-pdosqlsrv.yaml` failed with the QEMU runner and now it builds just fine!


mattmoor:
Looks like this fixes quite a number of the outstanding package failures.  A large chunk of `py3-*` and `ruby` failures (I sampled a few that appear to use `working-directory:` with a relative path, and they all built now).

mattmoor:
cc @stevebeattie FYI

jonjohnsonjr (approved):

PR #1973, opened 2025-05-07
fix: exclude fifos in WorkspaceTar

89luca89:
Fifos when compressed/extracted won't work anyway and they are usually problematic (see dpkg package)

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -402,7 +402,7 @@ func (bw *qemu) WorkspaceTar(ctx context.Context, cfg *Config) (io.ReadCloser, e
 		nil,
 		outFile,
 		false,
-		[]string{"sh", "-c", "cd /home/build && tar cvpzf - --xattrs --acls *"},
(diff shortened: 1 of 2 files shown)


pkg/container/qemu_runner.go
```diff
@@ -402,7 +402,7 @@ func (bw *qemu) WorkspaceTar(ctx context.Context, cfg *Config) (io.ReadCloser, e
 		nil,
 		outFile,
 		false,
-		[]string{"sh", "-c", "cd /home/build && tar cvpzf - --xattrs --acls *"},
+		[]string{"sh", "-c", "cd /home/build && tar cvpzf - --xattrs --acls --exclude='*fifo*' *"},
```
mattmoor, line 405: Should we include `dpkg` in the list of wolfi packages we build on PRs both to confirm this works, but also as a regression test?

89luca89: right, will add :+1:

89luca89: done

mattmoor (approved):

PR #1974, opened 2025-05-07
cli: Implement new `compare-versions` command

sergiodj:
Sometimes it's useful to check how two versions relate to one another. Inspired by Debian's `dpkg --compare-versions`, I decided to implement our very own version comparison CLI.

### DIFF
--- a/docs/md/melange.md
+++ b/docs/md/melange.md
@@ -22,6 +22,7 @@ toc: true
 
 * [melange build](/docs/md/melange_build.md)	 - Build a package from a YAML configuration file
 * [melange bump](/docs/md/melange_bump.md)	 - Update a Melange YAML file to reflect a new package version
+* [melange compare-versions](/docs/md/melange_compare-versions.md)	 - Compare two package versions
 * [melange compile](/docs/md/melange_compile.md)	 - Compile a YAML configuration file
 * [melange completion](/docs/md/melange_completion.md)	 - Generate completion script
 * [melange convert](/docs/md/melange_convert.md)	 - EXPERIMENTAL COMMAND - Attempts to convert packages/gems/apkbuild files into melange configuration files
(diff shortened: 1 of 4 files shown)


murraybd:
One thing that always bothered me about this argument with dpkg was having to check the exit code. Could we do something different?

sergiodj:
> One thing that always bothered me about this argument with dpkg was having to check the exit code. Could we do something different?

Certainly!

Do you have something in mind?  I guess the command could accept a `--slient` argument, but otherwise be verbose about the result.  WDYT?

murraybd:
> > One thing that always bothered me about this argument with dpkg was having to check the exit code. Could we do something different?
> 
> Certainly!
> 
> Do you have something in mind? I guess the command could accept a `--silent` argument, but otherwise be verbose about the result. WDYT?

No, I didn't have something specific in mind but your suggestion sounds good to me.

sergiodj:
> > > One thing that always bothered me about this argument with dpkg was having to check the exit code. Could we do something different?
> > 
> > 
> > Certainly!
> > Do you have something in mind? I guess the command could accept a `--silent` argument, but otherwise be verbose about the result. WDYT?
> 
> No, I didn't have something specific in mind but your suggestion sounds good to me.

Great.  I've force-pushed an implementation of my idea.  Let me know what you think.

Thanks.

xnox:
> Sometimes it's useful to check how two versions relate to one another. Inspired by Debian's `dpkg --compare-versions`, I decided to implement our very own version comparison CLI.

And the `apk version --test|--check` is not enough?

See https://man.archlinux.org/man/apk-version.8.en for documentation

sergiodj:
> > Sometimes it's useful to check how two versions relate to one another. Inspired by Debian's `dpkg --compare-versions`, I decided to implement our very own version comparison CLI.
> 
> And the `apk version --test|--check` is not enough?
> 
> See https://man.archlinux.org/man/apk-version.8.en for documentation

It certainly fits the bill in a way, but I still believe it's useful to have melange's own version checker because, in the (unlikely?) scenario where we choose to deviate from apk's versioning scheme, having our own checker can really help.

PR #1975, opened 2025-05-08
Drop the `isBuildLess` optimization.

mattmoor:
There is only a single package in Wolfi that exhibits this.

### DIFF
--- a/.github/workflows/wolfi-presubmit.yaml
+++ b/.github/workflows/wolfi-presubmit.yaml
@@ -73,6 +73,7 @@ jobs:
           - fping
           - subversion
           - sudo
+          - py3-supported-python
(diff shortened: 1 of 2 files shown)


.github/workflows/wolfi-presubmit.yaml
```diff
@@ -73,6 +73,7 @@ jobs:
           - fping
           - subversion
           - sudo
+          - py3-supported-python
```
stevebeattie, line 76: Was this intended to be part of the change?

mattmoor: Yeah, this was the only "buildless" package in Wolfi, and I added it here to make sure it works going forward

PR #1976, opened 2025-05-08
license-check: tweaks to license display and renovate

sil2100:
More tweaks to the license checking logic:

- List out all licenses that could not be detected properly and mark them as NOASSERTION
- Print out all the low-confidence licenses as a separate list and urge the developers to look at those
- Fix renovate to ignore low-confidence licenses (as we can't really vouch for those)
- Fix renovate not to clear out license information when there are no valid licenses detected

### DIFF
--- a/pkg/renovate/copyright/copyright.go
+++ b/pkg/renovate/copyright/copyright.go
@@ -59,11 +59,19 @@ func New(ctx context.Context, opts ...Option) renovate.Renovator {
 	return func(ctx context.Context, rc *renovate.RenovationContext) error {
 		log.Infof("attempting to update copyright")
 
-		// Funny thing: we first have to parse the configuration afresh for license linting.
-		// The reason is that for determining the license information, we need to fetch the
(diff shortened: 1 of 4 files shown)


pkg/renovate/copyright/copyright.go
```diff
@@ -29,20 +29,30 @@ import (
...
+// renovator.
 func WithLicenses(licenses []license.License) Option {
 	return func(cfg *CopyrightConfig) error {
 		cfg.Licenses = licenses
 		return nil
 	}
 }
 
+// WithLicenses sets the differences to consider for the
+// renovator.
```
dannf, line 48: similar to above, should this comment be about what WithDiffs() does vs what WithLicenses() does?

dannf (approved):
just a couple nits about comments inline, otherwise LGTM :)

sil2100:
Thanks for the review! ugh, I'm like very bad in copy-paste of comments it seems! :sweat_smile:

PR #1977, opened 2025-05-09
experiment: on the fly disk for melange QEMU runner

89luca89:
With this we adapt to the equivalent branch of 89luca89/os this way we use this minimal initramfs to unpack the rootfs.tar.gz mounted to /dev/vdb into /mount, and then use SSHD's ChrootDirectory to chroot into it

This solves the following problem:

- multiple compression/decompression cycle of the rootfs, hence a big performance improvement
  - we now use `layer.Uncompressed()` which is the same as bubblewrap runner now.
- removes the ~2gb limit for the initramfs as now we're using the disk
- cpio initramfs is now cache-able, so for local development things are also faster
- removes need of SSHD inside the build environment
- migrates to use SSHD's `ChrootDirectory` to enter the build environment (avoid access at build time to vmlinuz modules sshd binary etc)
- removes extra files in the build environment (eg: ssh keys, authorized keys) thus solving issues with "approved licence" file scanners

Generally speaking with this PR we improve performance but also reliability and make the build environment less "tainted"

Depends on:
- [x] https://github.com/chainguard-dev/enterprise-packages/pull/23129
- [x] https://github.com/wolfi-dev/os/pull/52911
- [x] https://github.com/wolfi-dev/os/pull/52948

### DIFF
--- a/.github/workflows/wolfi-presubmit.yaml
+++ b/.github/workflows/wolfi-presubmit.yaml
@@ -68,7 +68,7 @@ jobs:
           - guac
           - mdbook
           - s3cmd
-          - db  # Uses license-path
+          - py3-pyelftools  # Uses license-path
           - cadvisor  # uses cgroups
           - fping  # uses get/setcaps
           - perl-yaml-syck
--- a/.github/workflows/wolfi-presubmit.yaml
+++ b/.github/workflows/wolfi-presubmit.yaml
@@ -68,7 +68,10 @@ jobs:
           - guac
           - mdbook
           - s3cmd
-          - db # Uses license-path
+          - db  # Uses license-path
+          - cadvisor  # uses cgroups
+          - dotnet-9  # huge workspace size 3+gb
+          - fping  # uses get/setcaps
           - perl-yaml-syck
           - ncurses
           - fping
--- a/.github/workflows/wolfi-presubmit.yaml
+++ b/.github/workflows/wolfi-presubmit.yaml
@@ -70,7 +70,6 @@ jobs:
           - s3cmd
           - db  # Uses license-path
           - cadvisor  # uses cgroups
-          - dotnet-9  # huge workspace size 3+gb
           - fping  # uses get/setcaps
           - perl-yaml-syck
           - ncurses
@@ -156,8 +155,8 @@ jobs:
         run: |
           make SHELL="/bin/bash" MELANGE="sudo melange" test/${{ matrix.package }}
 
-      - name: Run tests to verify xattrs with QEMU runner
-        if: matrix.runner == 'qemu' && matrix.package == 'fping'
+      - name: Run tests with QEMU runner
+        if: matrix.runner == 'qemu'
         run: |
           make \
             SHELL="/bin/bash" \
--- a/.github/workflows/wolfi-presubmit.yaml
+++ b/.github/workflows/wolfi-presubmit.yaml
@@ -71,6 +71,7 @@ jobs:
           - py3-pyelftools  # Uses license-path
           - cadvisor  # uses cgroups
           - fping  # uses get/setcaps
+          - memcached  # uses a diff test user
           - perl-yaml-syck
           - ncurses
           - subversion
--- a/.github/workflows/wolfi-presubmit.yaml
+++ b/.github/workflows/wolfi-presubmit.yaml
@@ -74,7 +74,6 @@ jobs:
           - fping  # uses get/setcaps
           - perl-yaml-syck
           - ncurses
-          - fping
           - subversion
           - sudo
           - py3-supported-python
--- a/.github/workflows/wolfi-presubmit.yaml
+++ b/.github/workflows/wolfi-presubmit.yaml
@@ -157,6 +157,17 @@ jobs:
         run: |
           make SHELL="/bin/bash" MELANGE="sudo melange" test/${{ matrix.package }}
 
+      - name: Run tests to verify xattrs with QEMU runner
+        if: matrix.runner == 'qemu' && matrix.package == 'fping'
+        run: |
+          make \
+            SHELL="/bin/bash" \
+            QEMU_KERNEL_IMAGE=/tmp/kernel/boot/vmlinuz-virt \
+            QEMU_KERNEL_MODULES=/tmp/kernel/lib/modules/ \
+            MELANGE="/usr/bin/melange" \
+            MELANGE_EXTRA_OPTS="--runner qemu" \
+            test/${{ matrix.package }}
+
       - name: Check package ${{ matrix.package }} xattrs for QEMU-built package
         if: matrix.runner == 'qemu' && matrix.package == 'fping'
         run: |
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -16,7 +16,6 @@ package build
 
 import (
 	"archive/tar"
-	"compress/gzip"
 	"context"
 	"encoding/hex"
 	"encoding/json"
@@ -1136,7 +1135,14 @@ func (b *Build) retrieveWorkspace(ctx context.Context, fs apkofs.FullFS) error {
 	ctx, span := otel.Tracer("melange").Start(ctx, "retrieveWorkspace")
 	defer span.End()
 
-	r, err := b.Runner.WorkspaceTar(ctx, b.containerConfig)
+	extraFiles := []string{}
+	for _, v := range b.Configuration.Package.Copyright {
+		if v.LicensePath != "" {
+			extraFiles = append(extraFiles, v.LicensePath)
+		}
+	}
+
+	r, err := b.Runner.WorkspaceTar(ctx, b.containerConfig, extraFiles)
 	if err != nil {
 		return err
 	} else if r == nil {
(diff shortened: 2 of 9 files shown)


mattmoor:
We should add a handful of test cases to the wolfi leg based on what we've been seeing.

Some classes that come to mind:
 - `cadvisor` for cgroups
 - Something with a `getcap` test
 - Something with an enormous `initrd` (maybe a synthetic test here with an otherwise nop build would be best?)
 - Something with a test user (this may be separate from this work)

89luca89:
> We should add a handful of test cases to the wolfi leg based on what we've been seeing.
> 
> Some classes that come to mind:
> 
> * `cadvisor` for cgroups
> * Something with a `getcap` test
> * Something with an enormous `initrd` (maybe a synthetic test here with an otherwise nop build would be best?)
> * Something with a test user (this may be separate from this work)

Added some packages to the presubmit, should cover all the fixes we're introducing

.github/workflows/wolfi-presubmit.yaml
```diff
@@ -68,7 +68,10 @@ jobs:
           - guac
           - mdbook
           - s3cmd
-          - db # Uses license-path
+          - db  # Uses license-path
+          - cadvisor  # uses cgroups
+          - dotnet-9  # huge workspace size 3+gb
+          - fping  # uses get/setcaps
```
mattmoor, line 74: Looks like this is already below 🤔 

I think what this is missing is running tests!

89luca89: Added the test case, now we need to merge this: https://github.com/wolfi-dev/os/pull/52948
to fix the xattr losing

mattmoor:
I think that for dotnet, instead of doing the actual build we should create a contrived example with a large build environment, but a quick build.  WDYT?

.github/workflows/wolfi-presubmit.yaml
```diff
@@ -154,6 +156,17 @@ jobs:
         run: |
           make SHELL="/bin/bash" MELANGE="sudo melange" test/${{ matrix.package }}
 
+      - name: Run tests to verify xattrs with QEMU runner
+        if: matrix.runner == 'qemu' && matrix.package == 'fping'
```
mattmoor, line 160: I was thinking that we should probably always do this.  We can exclude stuff that fails (with a TODO to fix them)?

pkg/build/build.go
```diff
@@ -310,7 +309,6 @@ func (b *Build) buildGuest(ctx context.Context, imgConfig apko_types.ImageConfig
 
 	if b.Runner.Name() == container.QemuName {
 		b.ExtraPackages = append(b.ExtraPackages, []string{
-			"melange-microvm-init",
```
mattmoor, line 313: Do we need this special-case block at all anymore?

pkg/container/qemu_runner.go
```diff
@@ -418,42 +426,23 @@ func (b qemuOCILoader) LoadImage(ctx context.Context, layer v1.Layer, arch apko_
 	defer span.End()
 
 	// qemu does not have the idea of container images or layers or such, just
-	// create an initramfs from the layer
-	guestInitramfs, err := os.CreateTemp("", "melange-guest-*.initramfs.cpio")
+	// create a rootfs from the layer
+	guestRootfs, err := os.CreateTemp("", "melange-guest-*.tar.gz")
```
mattmoor, line 430: not `.gz` anymore

pkg/container/qemu_runner.go
```diff
@@ -621,19 +600,39 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 		cfg.Disk = defaultDiskSize
 	}
 
+	// always be sure to create the VM rootfs first!
```
mattmoor, line 603: Why?

pkg/container/qemu_runner.go
```diff
@@ -1188,3 +1115,187 @@ func randomPortN() (int, error) {
...
+	if err := apko_cpio.FromLayer(layer, gzw); err != nil {
+		clog.FromContext(ctx).Errorf("failed to convert cpio initramfs: %v", err)
+		return "", err
+	}
+	if err := gzw.Close(); err != nil {
+		clog.FromContext(ctx).Errorf("failed to close gzip writer: %v", err)
+		return "", err
+	}
+
+	return guestInitramfs.Name(), nil
```
mattmoor, line 1224: I would skip the gzip since this initrd is always small and one of the key motivations for zipping before was to squeeze under the 2GB limit.

pkg/container/qemu_runner.go
```diff
@@ -1188,3 +1115,187 @@ func randomPortN() (int, error) {
...
+}
+
+// in case of external modules (usually for 9p and virtio) we need a matching /lib/modules/kernel-$(uname)
```
mattmoor, line 1227: Is this usrmerge safe?

pkg/container/qemu_runner.go
```diff
@@ -1188,3 +1115,187 @@ func randomPortN() (int, error) {
...
+	}
+
+	spec := apko_types.ImageConfiguration{
+		Contents: apko_types.ImageContents{
+			RuntimeRepositories: []string{
+				"https://packages.wolfi.dev/os",
+			},
+			Keyring: []string{
+				"https://packages.wolfi.dev/os/wolfi-signing.rsa.pub",
+			},
```
mattmoor, line 1149:
```suggestion
			RuntimeRepositories: []string{
				"https://apk.cgr.dev/chainguard",
			},
```

pkg/container/qemu_runner.go
```diff
@@ -1188,3 +1115,187 @@ func randomPortN() (int, error) {
 
 	return l.Addr().(*net.TCPAddr).Port, nil
 }
+
+func generateCpio(ctx context.Context) (string, error) {
+	/*
+	 * we only build once, useful for local development, we
+	 * cahce it into kernel/ARCH/melange-guest.initramfs.cpio
```
mattmoor, line 1122:
```suggestion
	 * cache it into kernel/ARCH/melange-guest.initramfs.cpio
```

pkg/container/qemu_runner.go
```diff
@@ -1188,3 +1115,187 @@ func randomPortN() (int, error) {
...
+	 * if present, we nop and return, else we build it.
```
mattmoor, line 1123: Perhaps we should use this, so it doesn't dirty trees and potentially shares across repos?
https://pkg.go.dev/os#UserCacheDir

This should have some sort of expiration, I'd think...

pkg/build/build.go
89luca89: Yes, we need gnutar/attr to copy back the workspace, I'll try to think of a way to avoid it

pkg/container/qemu_runner.go
89luca89: Fixed, moved to buildrepo also

pkg/container/qemu_runner.go
89luca89: Fixed

pkg/container/qemu_runner.go
89luca89: Fixed

pkg/container/qemu_runner.go
89luca89: Fixed

pkg/container/qemu_runner.go
89luca89: Leftover comment from time ago, removed

.github/workflows/wolfi-presubmit.yaml
89luca89: Ok moving to run all tests with qemu

pkg/build/build.go
mattmoor: I think that's fine.  Let's just add comments saying why we need certain things?

89luca89: Done :+1:

pkg/container/qemu_runner.go
89luca89: Right, done :+1:

mattmoor (approved):
LGTM.  I think the plan is to merge my changes into here and then merging this branch into `main`.

mattmoor (approved):

pkg/container/qemu_runner.go
```diff
@@ -393,6 +385,19 @@ func (bw *qemu) WorkspaceTar(ctx context.Context, cfg *Config) (io.ReadCloser, e
 	// We could just cp -a to /mnt as it is our shared workspace directory, but
 	// this will lose some file metadata like hardlinks, owners and so on.
 	// Example of package that won't work when using "cp -a" is glibc.
+	retrieveCommand := "cd /home/build && tar cvpf - --xattrs --acls --exclude='*fifo*' melange-out"
+	// we append also all the necessary files that we might need, for example Licenses
+	// for license checks
+	for _, v := range extraFiles {
+		retrieveCommand = retrieveCommand + " " + v
+	}
```
mattmoor, line 393: I don't think this works if the license-path includes a subdirectory?

I am seeing a failure on: https://github.com/chainguard-dev/extra-packages/blob/main/camunda-zeebe-8.6.yaml

89luca89: This will fix it: https://github.com/chainguard-dev/melange/pull/1985
it's a problem with the unpack, not the packing

PR #1979, opened 2025-05-12
fetch-source: export the simple fetch source mechanism via cli

sil2100:
We have added very simple source-fetching functionality to melange for the license-checking work. Let's export this via an explicit CLI command. This works for both .yaml files and .apk files.


sil2100:
Found a small bug in this right now, fixed by getting https://github.com/chainguard-dev/melange/pull/2106 applied. Let's rebase after the other is landed.

PR #1982, opened 2025-05-12
feat: avoid adding additional packages on QEMU runners

89luca89:
When using QEMU runners we were forced to add gnutar/attr packages to the build environment in order to copy back the workspace outputs.

With the new version of the microvm init, we can instead use the alternative SSHD port, that avoid chrooting into the build environment.

This will let us use the initramfs packages instead of injecting them into the build environment.

Depends on:

- [ ] https://github.com/wolfi-dev/os/pull/53169

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -338,6 +338,13 @@ func (bw *qemu) StartPod(ctx context.Context, cfg *Config) error {
 
 	cfg.SSHAddress = "127.0.0.1:" + strconv.Itoa(port)
 
+	port, err = randomPortN()
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/container/qemu_runner.go
```diff
@@ -1151,6 +1161,8 @@ func generateCpio(ctx context.Context) (string, error) {
 			BuildRepositories: []string{
 				"https://apk.cgr.dev/chainguard",
 			},
+			Keyring: []string{
+			},
```
mattmoor, line 1165:
```suggestion
```

mattmoor (dismissed):

pkg/container/qemu_runner.go
89luca89: removed

mattmoor (approved):

PR #1983, opened 2025-05-13
sbom: Populate downloadLocation and checksums

javacruft:
For packages that related to fetch and git-checkout actions in build pipelines populate the downloadLocation and checksums fields alongside the purl's that are already present in externalRefs.

Download location is either a URL or a Vcs reference as defined in the SPDX specification.  Checksums are provided for both types.

[SPDX reference for examples](https://spdx.github.io/spdx-spec/v2.3/package-information/#773-examples)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -588,8 +588,6 @@ func (p Pipeline) SBOMPackageForUpstreamSource(licenseDeclared, supplier string,
 		tag := with["tag"]
 		expectedCommit := with["expected-commit"]
 		downloadLocation := "git+" + repo
-		checksums := make(map[string]string)
-		checksums["SHA1"] = expectedCommit
 
 		// We'll use all available data to ensure our SBOM's package ID is unique, even
 		// when the same repo is git-checked out multiple times.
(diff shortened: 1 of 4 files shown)


sil2100 (dismissed):
I know this is still a draft, but I look at this changeset here and I feel like this is the way to go. Thank you!

javacruft:
> I know this is still a draft, but I look at this changeset here and I feel like this is the way to go. Thank you!

Once we've discussed with consumers to validate the approach then I'll mark ready for review.

pkg/config/config.go
```diff
@@ -649,6 +660,14 @@ func (p Pipeline) SBOMPackageForUpstreamSource(licenseDeclared, supplier string,
 		// Encode vcs_url with git+ prefix and @commit suffix
 		vcsUrl := "git+" + repo
 
+		if len(tag) > 0 {
+			downloadLocation += "@" + tag
+		} else {
+			if len(expectedCommit) > 0 {
+				downloadLocation += "@" + expectedCommit
+			}
+		}
```
imjasonh, line 669:
```suggestion
		} else if len(expectedCommit) > 0 {
			downloadLocation += "@" + expectedCommit
		}
```

(super nit)

imjasonh (dismissed):

imjasonh (approved):

PR #1985, opened 2025-05-14
fix: ensure make parent dirs when unpacking files from workspace

89luca89:
Add some output to tar generation
Make sure that parent directory of unpacked files exists


maxgio92 (approved):
LGTM

PR #1986, opened 2025-05-14
Fix/metapackage builds

89luca89:
Revert c5071e721ed96d0e940ea75176f2e3c1c6346458 that broke metapackage builds also on other runners
Support buildless packages on qemu


maxgio92 (approved):
LGTM

PR #1988, opened 2025-05-15
Fix/random port pwd shell

89luca89:
1 - ensure debug shell goes in the right working directory
2 - ensure the two ssh ports won't collide

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -298,7 +298,8 @@ func (bw *qemu) Debug(ctx context.Context, cfg *Config, envOverride map[string]s
 
 	clog.FromContext(ctx).Debugf("running debug command: %v", args)
 
-	err = session.Shell()
+	cmd := shellquote.Join(args...)
+	err = session.Run(cmd)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/container/qemu_runner.go
```diff
@@ -298,7 +298,8 @@ func (bw *qemu) Debug(ctx context.Context, cfg *Config, envOverride map[string]s
 
 	clog.FromContext(ctx).Debugf("running debug command: %v", args)
 
-	err = session.Shell()
+	cmd := shellquote.Join(args...)
+	err = session.Run(cmd)
```
mattmoor, line 302: This doesn't look right.  Won't this run the command instead of creating a Shell?

89luca89: well the command is a shell (it's cd /home/build && /bin/sh)
it works well from my local testing, you have a shell resizable and with all the ctrl-c/d/a/z stuff working

mattmoor: Got it.  I thought this was the command being debugged.

mattmoor (approved):

PR #1989, opened 2025-05-15
license-check: fix running checks for builds using qemu

sil2100:
We have recently added license-checking to the build pipeline, running our license identification against the build workspace to come up with what licenses are present in the source.

This works fine for bubblewrap as the workspace directory operates via bind mounts, but for qemu, the workspace is on the qemu disk and we only get the artifacts via `retrieveWorkspace()` - which only syncs the `melange-out/` dir + any other files we explicitly ask for. So licensing checks can't run properly, as they need the source from the workspace for analysis.
Modify the qemu runner to fetch all the license related files and copy them over as part of `retrieveWorkspace()`.

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -392,12 +392,13 @@ func (bw *qemu) WorkspaceTar(ctx context.Context, cfg *Config, extraFiles []stri
 	// For qemu, we also want to get all the detected license files for the
 	// license checking that will be done later.
 	// First, get the list of all files from the remote workspace.
-	licenseFiles, err := getWorkspaceLicenseFiles(ctx, cfg)
+	licenseFiles, err := getWorkspaceLicenseFiles(ctx, cfg, extraFiles)
 	if err != nil {
 		clog.FromContext(ctx).Errorf("failed to extract list of files for licensing: %v", err)
 		return nil, err
 	}
-	// Now, append those files to the extraFiles list.
+	// Now, append those files to the extraFiles list (there should be no
+	// duplicates)
 	extraFiles = append(extraFiles, licenseFiles...)
 
 	outFile, err := os.Create(filepath.Join(cfg.WorkspaceDir, "melange-out.tar"))
@@ -788,7 +789,7 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 
 // getWorkspaceLicenseFiles returns a list of possible license files from the
 // workspace
-func getWorkspaceLicenseFiles(ctx context.Context, cfg *Config) ([]string, error) {
+func getWorkspaceLicenseFiles(ctx context.Context, cfg *Config, extraFiles []string) ([]string, error) {
 	// default to root user, unless a different user is specified
 	user := "root"
 	if cfg.RunAs != "" {
@@ -817,10 +818,20 @@ func getWorkspaceLicenseFiles(ctx context.Context, cfg *Config) ([]string, error
 		return nil, err
 	}
 
-	// Now, we can read the list of files from the string writer.
+	// Turn extraFiles into a map for faster lookup
+	extraFilesMap := make(map[string]struct{})
+	for _, file := range extraFiles {
+		extraFilesMap[filepath.Clean(file)] = struct{}{}
+	}
+
+	// Now, we can read the list of files from the string writer and add those
+	// license files that are not in the extraFiles list
 	licenseFiles := []string{}
 	foundFiles := strings.SplitSeq(buf.String(), "\n")
 	for f := range foundFiles {
+		if _, ok := extraFilesMap[filepath.Clean(f)]; ok {
+			continue
+		}
 		if strings.Contains(f, "melange-out") {
 			continue
 		}
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -45,6 +45,7 @@ import (
 	apko_types "chainguard.dev/apko/pkg/build/types"
 	apko_cpio "chainguard.dev/apko/pkg/cpio"
 	"chainguard.dev/melange/internal/logwriter"
+	"chainguard.dev/melange/pkg/license"
 	"github.com/chainguard-dev/clog"
 	"github.com/charmbracelet/log"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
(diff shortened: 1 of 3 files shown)


pkg/license/license.go
```diff
@@ -187,6 +171,29 @@ func FindLicenseFiles(fsys fs.FS) ([]LicenseFile, error) {
...
+		// licensee does this check as part of the regex, but in go we don't have
+		// the same regex capabilities
+		for _, ext := range ignoredExt {
+			if ignore = filepath.Ext(filename) == ext; ignore {
+				break
+			}
+		}
+		if ignore {
+			continue
+		}
```
sergiodj, line 191:
```suggestion
		// licensee does this check as part of the regex, but in go we don't have
		// the same regex capabilities
		if slice.Contains(ignoredExt, filenameExt) {
			continue
		}
```

pkg/license/license.go
```diff
@@ -187,6 +171,29 @@ func FindLicenseFiles(fsys fs.FS) ([]LicenseFile, error) {
 	return licenseFiles, nil
 }
 
+// IsLicenseFile checks if a file is a license file based on its name.
```
sergiodj, line 174: I like to leave a brief description of each return value of a function.  For example, it took me a few seconds to find what `weight` represents here.  WDYT?

sergiodj (changes requested):
Thanks, @sil2100!

This looks great.  I'm leaving a small suggestion to optimize `IsLicenseFile`, let me know what you think.

sil2100:
> I thought this was enough
> 
> https://github.com/chainguard-dev/melange/blob/83d4d3ec5dc419946ab0510833d6e006b6b29797/pkg/build/build.go#L1190
> 
> what would that be missing?
> 
> EDIT: to clarify, this snippet is adding the declared license files to the tar, which is unpacked in the /tmp/melange-workspace-xxxxx/ directory, so files should be found there (and it seems testing locally they are found by license checks)

That is not enough, no. That code copies the existing declared license files in the given melange .yaml - so this already requires someone to do the due diligence and declare all the license files manually. Where our license-linting initiative aims to scan the *whole* tree, look for files that seem to be licenses and then perform detection/classification on those. This is needed because we need to constantly know if any new license files have been added, moved or changed.

So we need to do this from the builder POV, since we need to actually scan the source tree and find all files that seem to be licenses without relying on humans preparing the lists first. Then the goal is to add the relevant info to the yaml, yes, but that's the end goal.

89luca89:
> That is not enough, no. That code copies the existing declared license files in the given melange .yaml - so this already requires someone to do the due diligence and declare all the license files manually. Where our license-linting initiative aims to scan the _whole_ tree, look for files that seem to be licenses and then perform detection/classification on those. This is needed because we need to constantly know if any new license files have been added, moved or changed.

Oh I got it, we scan also the various license in the worktree, got it :+1: 
 
> So we need to do this from the builder POV, since we need to actually scan the source tree and find all files that seem to be licenses without relying on humans preparing the lists first. Then the goal is to add the relevant info to the yaml, yes, but that's the end goal.

:+1: that's ok disregard my comment then :joy:

sil2100:
> > That is not enough, no. That code copies the existing declared license files in the given melange .yaml - so this already requires someone to do the due diligence and declare all the license files manually. Where our license-linting initiative aims to scan the _whole_ tree, look for files that seem to be licenses and then perform detection/classification on those. This is needed because we need to constantly know if any new license files have been added, moved or changed.
> 
> Oh I got it, we scan also the various license in the worktree, got it 👍
> 
> > So we need to do this from the builder POV, since we need to actually scan the source tree and find all files that seem to be licenses without relying on humans preparing the lists first. Then the goal is to add the relevant info to the yaml, yes, but that's the end goal.
> 
> 👍 that's ok disregard my comment then 😂

It is a valid thing too look into, since we might as well remove the previous code that was copying over the licensing files! Since in theory the new code should just copy over all the files as-is. Then again, I can see some cases where our automation doesn't detect that a file is a license (as it's name-based at this moment, the actual classification happens later), but say we *know* it's a license, so we'd still want to pull it into the workstation :) I'll follow up with some nice addons to the license-checker to fully handle this!

pkg/container/qemu_runner.go
```diff
@@ -774,6 +786,52 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
...
+	err := sendSSHCommand(ctx,
+		user,
+		cfg.SSHWorkspaceAddress,
+		cfg,
+		nil,
+		nil,
+		nil,
+		bufWriter,
+		false,
+		[]string{"sh", "-c", "cd /mount/home/build && find . -type f -print"},
```
89luca89, line 812: you can just do 

```go
		[]string{"sh", "-c", "find /mount/home/build -type f -print"},
```

89luca89 (changes requested):

pkg/container/qemu_runner.go
sergiodj: Without looking at the code further, it actually depends if the code is expecting a filename-only output, or if it can deal with absolute pathnames:

```
$ find /tmp/aaa/ -type f -print
/tmp/aaa/c
/tmp/aaa/b
/tmp/aaa/a
$ cd aaa/ && find . -type f -print
./c
./b
./a
```

89luca89: From what I see it later does `filepath.Base()` so it shouldn't be a problem

sergiodj (dismissed):
LGTM, thanks!

You might want to adopt @89luca89's suggestion before merging.

pkg/container/qemu_runner.go
sil2100: hm, it doesn't seem to work as expected with the proposed change! The workspace ends up with directory structures like these:

```
mount/home/build/LICENSE
mount/home/build/runtime/pack/dist/opt/editorconfig/LICENSE
...
```

So I think the ugly `cd` should stay!

89luca89:
@sil2100 py3-pyelftools is failing here:

```
ERRO failed to build package: retrieving workspace: link /tmp/melange-workspace-27751064/LICENSE /tmp/melange-workspace-27751064/LICENSE: file exists
```

I think we either need to deduplicate the `extraFiles` before we use that in `WorkspaceTar()` or we might just want to check for reguar files and skip symlinks?

89luca89 (changes requested):
@sil2100 py3-pyelftools is failing here:

```
ERRO failed to build package: retrieving workspace: link /tmp/melange-workspace-27751064/LICENSE /tmp/melange-workspace-27751064/LICENSE: file exists
```

I think we either need to deduplicate the extraFiles before we use that in WorkspaceTar() or we might just want to check for reguar files and skip symlinks?

sil2100:
You are right, there is an issue here indeed! Let me deduplicate files before continuing. Good that the existing tests caught this!

89luca89 (approved):

PR #1990, opened 2025-05-15
fix: improve QEMU resource limits handling

89luca89:
Default to use 85% or available memory and all CPU cores. If a resource snippet is declared, it will be treated as an upbound limit for resources.

Fixes: https://github.com/chainguard-dev/prodsec/issues/143

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -550,29 +550,29 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 		}
 	}
 
+	// default to use 85% of available memory, if a mem limit is set, respect it.
+	mem := int64(float64(getAvailableMemoryKB())*0.85)
 	if cfg.Memory != "" {
 		memKb, err := convertHumanToKB(cfg.Memory)
 		if err != nil {
 			return err
 		}
 
-		if memKb > int64(getAvailableMemoryKB()) {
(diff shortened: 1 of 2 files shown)


pkg/container/qemu_runner.go
```diff
@@ -550,29 +550,29 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
...
+	// default to use 85% of available memory, if a mem limit is set, respect it.
+	mem := int64(float64(getAvailableMemoryKB())*0.85)
 	if cfg.Memory != "" {
 		memKb, err := convertHumanToKB(cfg.Memory)
 		if err != nil {
 			return err
 		}
 
-		if memKb > int64(getAvailableMemoryKB()) {
-			log.Warnf("qemu: requested too much memory, requested: %d, have: %d", memKb, getAvailableMemoryKB())
```
mattmoor, line 560: Should we keep a warning when the build wants more resources than folks have?

89luca89: This shouldn't happen
we default to 85% ram, and we use cfg.Memory only if it's lower than that

mattmoor:
Looks like maybe we need to drop memcached for now.

mattmoor (approved):

89luca89:
> Looks like maybe we need to drop memcached for now.

Moved to sonarqube-10

mattmoor (approved):

PR #1992, opened 2025-05-16
sbom/direct download git locations

javacruft:
Switch hosted git repositories to provide a direct archive download location in the downloadLocation field.

This is referenced using the expected commit (immutable) or the tag if the commit is not provided.

Supported for Github and Gitlab based git sources.

Add `type-hint` parameter to git-checkout pipeline to allow undiscoverable Gitlab sources to be identified.


sil2100 (changes requested):
I know this is still a draft, but still took a look! I think this way of handling the download location is the right way to go. I have some inline thoughts though. Thank you!

pkg/config/config.go
```diff
@@ -608,6 +608,11 @@ func (p Pipeline) SBOMPackageForUpstreamSource(licenseDeclared, supplier string,
 
 		if strings.HasPrefix(repo, "https://github.com/") {
 			namespace, name, _ := strings.Cut(strings.TrimPrefix(repo, "https://github.com/"), "/")
+			name = strings.TrimSuffix(name, ".git")
+
+			// Always use the expected commit for the downloadLocation as this is immume to
+			// projects mutating a tag.
+			downloadLocation = fmt.Sprintf("https://github.com/%s/%s/archive/%s.tar.gz", namespace, name, expectedCommit)
```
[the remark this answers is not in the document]
javacruft: I'd not realized that expected-commit was optional; I'll refactor to fallback to tag if not provided

sil2100 (approved):
This looks good. I like the idea of using `url.Parse()` for it too.

pkg/config/config.go
```diff
@@ -523,6 +524,113 @@ type Pipeline struct {
...
+		repoType = purl.TypeGitlab
+		downloadLocation = fmt.Sprintf("%s://gitlab.com/%s/%s/-/archive/%s/%s.tar.gz", repoURL.Scheme, namespace, name, ref, ref)
+
+	case strings.HasPrefix(repoURL.Host, "gitlab") || hint == "gitlab":
+		repoType = purl.TypeGeneric
+		downloadLocation = fmt.Sprintf("%s://%s/%s/%s/-/archive/%s/%s.tar.gz", repoURL.Scheme, repoURL.Host, namespace, name, ref, ref)
+
+	default:
+		repoType = purl.TypeGeneric
+		namespace = ""
```
vishal-chdhry, line 567: @javacruft Is there a reason why this field is explicitly set to empty?
Its causing an issue where supplier is not correctly set: https://github.com/chainguard-dev/customer-issues/issues/2375

vishal-chdhry: we can set the namespace to the git repo if that is okay

javacruft: I think we should revert to the previous behaviour and use the supplier - let me raise a PR to fix.

javacruft: https://github.com/chainguard-dev/melange/pull/2081

PR #1993, opened 2025-05-16
sbom: pass SHA512 not SHA256 when in use.

javacruft:
Ensure the correct checksum is used when generating downloadLocation checksums.


sil2100 (approved):
Yep!

PR #1994, opened 2025-05-19
feat: use user's SSH keys to talk to QEMU runner

89luca89:
This makes debugging a lot easier, so people can easily enter a debug shell when building with `melange build --debug`

If no SSH Keys are present, they'll be generated on the spot like before.

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -22,10 +22,8 @@ import (
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rand"
-	"crypto/x509"
 	_ "embed"
 	"encoding/base64"
-	"encoding/pem"
 	"fmt"
 	"io"
 	"log/slog"
@@ -53,6 +51,7 @@ import (
 	"github.com/kballard/go-shellquote"
 	"go.opentelemetry.io/otel"
 	"golang.org/x/crypto/ssh"
+	"golang.org/x/crypto/ssh/agent"
 	"golang.org/x/crypto/ssh/knownhosts"
 	"golang.org/x/term"
 )
@@ -118,7 +117,9 @@ func (bw *qemu) Run(ctx context.Context, cfg *Config, envOverride map[string]str
 }
 
 func (bw *qemu) Debug(ctx context.Context, cfg *Config, envOverride map[string]string, args ...string) error {
-	clog.InfoContextf(ctx, "debugging command %s", strings.Join(args, " "))
+	clog.FromContext(ctx).Debugf("running debug command: %v", args)
+	clog.InfoContextf(ctx, "To enter this environment: ssh root@localhost -p %s",
+		strings.Split(cfg.SSHAddress, ":")[1])
 
 	// handle terminal size, resizing and sigwinch to keep
 	// it updated
@@ -139,12 +140,6 @@ func (bw *qemu) Debug(ctx context.Context, cfg *Config, envOverride map[string]s
 	signal.Notify(winch, syscall.SIGWINCH)
 	defer signal.Stop(winch)
 
-	signer, err := ssh.ParsePrivateKey(cfg.SSHKey)
-	if err != nil {
-		clog.FromContext(ctx).Errorf("Unable to parse private key: %v", err)
-		return err
-	}
-
 	hostKeyCallback, err := knownhosts.New(cfg.SSHHostKey)
 	if err != nil {
 		clog.FromContext(ctx).Errorf("could not create hostkeycallback function: %v", err)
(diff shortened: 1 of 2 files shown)


pkg/container/qemu_runner.go
```diff
@@ -118,7 +117,9 @@ func (bw *qemu) Run(ctx context.Context, cfg *Config, envOverride map[string]str
 }
 
 func (bw *qemu) Debug(ctx context.Context, cfg *Config, envOverride map[string]string, args ...string) error {
-	clog.InfoContextf(ctx, "debugging command %s", strings.Join(args, " "))
+	clog.FromContext(ctx).Debugf("running debug command: %v", args)
+	clog.InfoContextf(ctx, "To enter this environment: ssh root@localhost -p %s",
```
mattmoor, line 121: Should `root` become `cfg.RunAs`?

pkg/container/qemu_runner.go
```diff
@@ -498,10 +492,14 @@ func (b qemuOCILoader) RemoveImage(ctx context.Context, ref string) error {
 
 func createMicroVM(ctx context.Context, cfg *Config) error {
 	log := clog.FromContext(ctx)
-	log.Debug("qemu: ssh - create ssh key pair")
-	pubKey, err := generateSSHKeys(ctx, cfg)
-	if err != nil {
-		return err
+	log.Debug("qemu: ssh - get user ssh key pair")
+	pubKey, err := getUserSSHKey(cfg)
+	if err != nil || pubKey == nil {
```
mattmoor, line 497: IIUC this switches away from ephemeral keys by default and instead uses the user's SSH key pairs, if it can?

I would rather stick to an ephemeral key by default, but maybe we do this by default under `debug`?

pkg/container/qemu_runner.go
89luca89: Yep, fixed

pkg/container/qemu_runner.go
89luca89: Right, I've now made so that user's keys are injected just before the debug shell, this way we keep the ephemeral keypair during the whole build until the debug point happens

pkg/container/qemu_runner.go
```diff
@@ -118,8 +118,37 @@ func (bw *qemu) Run(ctx context.Context, cfg *Config, envOverride map[string]str
...
+
+	log.Debug("qemu: ssh - get user ssh key pair")
+	pubKey, err := getUserSSHKey()
+	if err != nil {
+		log.Warn("qemu: could not get user ssh key pair, using ephemeral ones")
+	}
+	if pubKey != nil {
+		command := fmt.Sprintf("echo '%s' | tee -a /root/.ssh/authorized_keys /home/*/.ssh/authorized_keys", string(pubKey))
+		err := sendSSHCommand(ctx,
+			"root",
```
mattmoor, line 136: Won't this break using our ephemeral key pair?

When debugging you can `exit 0` to continue the build, and I think this breaks that?

pkg/container/qemu_runner.go
```diff
@@ -1083,6 +1088,30 @@ func sendSSHCommand(ctx context.Context, user, address string,
 	return nil
 }
 
+func getUserSSHKey() ([]byte, error) {
+	socket := os.Getenv("SSH_AUTH_SOCK")
+	conn, err := net.Dial("unix", socket)
```
mattmoor, line 1093: What if this isn't set?

pkg/container/qemu_runner.go
89luca89: It shouldn't, `tee -a` is appending them

pkg/container/qemu_runner.go
mattmoor: I have this set to empty string...

egibs: This did cause issues in CI: https://github.com/chainguard-dev/melange/actions/runs/15114937303/job/42483172441?pr=1994#step:16:87

89luca89: it will not find an ssh agent to talk to
another approach would be to naively list `~/.ssh/*.pub` and add those

pkg/container/qemu_runner.go
89luca89: Confirm this works as intended

pkg/container/qemu_runner.go
89luca89: Fixed, now it will check the socket, if fails, fallback to check for files, else again just return nil and will skip the injection

pkg/container/qemu_runner.go
```diff
@@ -118,7 +118,38 @@ func (bw *qemu) Run(ctx context.Context, cfg *Config, envOverride map[string]str
...
+		user = cfg.RunAs
+	}
+
+	log.Debug("qemu: ssh - get user ssh key pair")
+	pubKey, err := getUserSSHKey()
+	if err != nil {
+		log.Warn("qemu: could not get user ssh key pair, using ephemeral ones")
+	}
+	if pubKey != nil {
+		command := fmt.Sprintf("echo '%s' | tee -a /root/.ssh/authorized_keys /home/*/.ssh/authorized_keys", string(pubKey))
```
mattmoor, line 135: I think this is "safer":
```suggestion
		command := fmt.Sprintf("echo %q | tee -a /root/.ssh/authorized_keys /home/*/.ssh/authorized_keys", string(pubKey))
```

mattmoor (approved):

PR #1999, opened 2025-05-21
test(e2e/pipelines): run tests on qemu runner

maxgio92:
This PR sets the QEMU runner as the default runner for e2e tests for pipelines.
It uses the Alpine kernel, the same way we do for the Wolfi presubmit CI workflow.

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

### DIFF
--- a/.github/workflows/melange-test-pipelines.yaml
+++ b/.github/workflows/melange-test-pipelines.yaml
@@ -40,7 +40,7 @@ jobs:
     name: Test packages
     needs:
       - build-melange
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-8-core
 
     permissions:
       contents: read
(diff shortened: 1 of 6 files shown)


.github/workflows/melange-test-pipelines.yaml
```diff
@@ -74,6 +74,27 @@ jobs:
           go-version-file: './go.mod'
           check-latest: true
 
+      - name: Download kernel for VMs
+        run: |
+          KERNEL_PKG="$(curl -sL https://dl-cdn.alpinelinux.org/alpine/edge/main/x86_64/APKINDEX.tar.gz | tar -Oxz APKINDEX | awk -F':' '$1 == "P" {printf "%s-", $2} $1 == "V" {printf "%s.apk\n", $2}' | grep "linux-virt" | grep -v dev)"
+          curl -LSo linux-virt.apk "https://dl-cdn.alpinelinux.org/alpine/edge/main/x86_64/$KERNEL_PKG"
```
joshrwolf, line 80: does it make sense to pin to a version here? and verify the checksum?

e2e-tests/capabilities-add-drop-build.yaml
```diff
@@ -20,12 +20,18 @@ environment:
 pipeline:
   - name: Test default effective capability
     runs: |
+      # Skip test if on QEMU runner, since the runner does not support process capabilities Melange feature.
+      { test -f /sys/class/dmi/id/sys_vendor && grep -E "^QEMU$" /sys/class/dmi/id/sys_vendor && exit 0; } || true
```
mattmoor, line 24: Hmm, what isn't supported?  cc @89luca89

maxgio92: @mattmoor because the capabilities add and drop feature has been implemented for the Docker and Bubblewrap runners only
* https://github.com/chainguard-dev/melange/blob/90c96db19b0b2a05d98fa7bc90b8775edeb0ddd1/pkg/container/bubblewrap_runner.go#L136
* https://github.com/chainguard-dev/melange/blob/90c96db19b0b2a05d98fa7bc90b8775edeb0ddd1/pkg/container/docker/docker_runner.go#L102

mattmoor: cc @antitree as well

.github/workflows/melange-test-pipelines.yaml
maxgio92: Good point. I think that would be easier to take the latest, to reduce maintenance since we're not really depending on specific kernel features for these tests. Also, latest provided by linux-virt is an LTS (ref https://wiki.alpinelinux.org/wiki/Kernels)

e2e-tests/capabilities-add-drop-build.yaml
89luca89: @antitree @mattmoor we can open a separate issue to track if we want cap-add/drop in QEMU I think and tackle it later

maxgio92:
I think we need force merge in this case, since test failures are not related to the changes, but CVEs found on fixture packages.
cc @chainguard-dev/tools-squad

.github/workflows/melange-test-pipelines.yaml
dannf: it feels like we should maybe import alpine's kernel into wolfi and build/maintain it like anything else

maxgio92:
@dannf I'm not against it. I would like to have approval from @chainguard-dev/sustaining-team for managing it.
In the meantime, I would unlock this PR if possible and manage it in a separated one to introduce the alpine kernel package and make Melange CI to use it. WDYT?

89luca89 (approved):

PR #2000, opened 2025-05-21
fix: add ignore paths to license files

89luca89:
This way we ignore huge number of files that are not actually part of the source

Fixes: https://github.com/chainguard-dev/prodsec/issues/156

### DIFF
--- a/pkg/license/license.go
+++ b/pkg/license/license.go
@@ -184,7 +184,7 @@ func IsLicenseFile(filename string) (bool, float64) {
 		"venv",
 	}
 	for _, i := range ignoredPaths {
-		if strings.Contains(filepath.Dir(filename), i) {
+		if slices.Contains(strings.Split(filename, string(filepath.Separator)), i) {
 			return false, 0.0
 		}
 	}
--- a/pkg/license/license.go
+++ b/pkg/license/license.go
@@ -143,7 +143,7 @@ func FindLicenseFiles(fsys fs.FS) ([]LicenseFile, error) {
 			return nil
 		}
 
-		is, weight := IsLicenseFile(info.Name())
+		is, weight := IsLicenseFile(filePath)
 		if is {
 			// Licenses in the top level directory have a higher weight so that they
 			// always appear first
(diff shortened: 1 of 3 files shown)


pkg/license/license.go
```diff
@@ -176,6 +176,22 @@ func FindLicenseFiles(fsys fs.FS) ([]LicenseFile, error) {
 // Returns true/fals if the file is a license file, and the weight value
 // associated with the match, as some matches are potentially more relevant.
 func IsLicenseFile(filename string) (bool, float64) {
+	// Ignore files in these paths
+	ignoredPaths := []string{
```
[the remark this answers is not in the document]
89luca89: Done :+1:

egibs (commented):
LGTM -- deferring to @sil2100 for final approval 😄

pkg/license/license.go
```diff
@@ -176,6 +176,22 @@ func FindLicenseFiles(fsys fs.FS) ([]LicenseFile, error) {
...
+		".virtualenv",
+		"env",
+		"node_modules",
+		"venv",
+	}
+	for _, i := range ignoredPaths {
+		if strings.Contains(filepath.Dir(filename), i) {
```
sil2100, line 187: One thing I think we should consider here: right now this is done with `strings.Contains()` on a string of the whole directory path chain. The problem is that this will match even if the `ignoredPaths` elements are part of a different word, like "**env**ious" etc. I think we'd need to do a == match on each path element instead. Maybe a filepath.SplitList and then trying to == match each of the split path elements?

sil2100 (changes requested):
Looking good, but I think there's one more thing we'd need to consider - see my inline comment. Thank you!

pkg/license/license.go
89luca89: or we could do a contains of `/foo/` that should isolate the directory name only :+1:

sil2100: Actually... yes! Originally I was worried that there might be cases where this would be the first or last directory in the path string... but I think it might be good? Since `Dir()` does normalize the path. Does it make absolute? Anyway, something surely that we could check!

89luca89: Done now :+1:

imjasonh (approved):

egibs (approved):

sil2100 (approved):
Crisp!

PR #2005, opened 2025-05-22
Fix detection of available memory.

dlorenc:
The old approach was getting total memory, which caused errors on my mac. This passes locally now!

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -1225,28 +1225,90 @@ func getAvailableMemoryKB() int {
 
 		for s.Scan() {
 			var n int
-			if nItems, _ := fmt.Sscanf(s.Text(), "MemTotal: %d kB", &n); nItems == 1 {
+			// Try to get MemAvailable first (available on newer kernels)
+			if nItems, _ := fmt.Sscanf(s.Text(), "MemAvailable: %d kB", &n); nItems == 1 {
 				return n
 			}
 		}
+
+		// If MemAvailable is not found, fall back to MemFree + Buffers + Cached
+		// Reset the file position
+		if _, err := f.Seek(0, 0); err != nil {
+			return 0
+		}
+		s = bufio.NewScanner(f)
+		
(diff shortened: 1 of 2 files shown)


pkg/container/qemu_runner.go
```diff
@@ -1221,28 +1221,97 @@ func getAvailableMemoryKB() int {
...
+			
+			// Get inactive pages
+			if strings.HasPrefix(line, "Pages inactive:") {
+				if _, err := fmt.Sscanf(line, "Pages inactive: %d.", &pagesInactive); err != nil {
+					continue
+				}
+				continue
+			}
+			
+			// Get speculative pages
```
89luca89, line 1295: We should avoid this, free+inactive 
Speculative is memory marked as "possible use" so under higher memory pressure situations it will encounter OOM problems

89luca89 (changes requested):

89luca89 (approved):

pkg/container/qemu_runner.go
```diff
@@ -1221,28 +1221,90 @@ func getAvailableMemoryKB() int {
 
 		for s.Scan() {
 			var n int
-			if nItems, _ := fmt.Sscanf(s.Text(), "MemTotal: %d kB", &n); nItems == 1 {
+			// Try to get MemAvailable first (available on newer kernels)
+			if nItems, _ := fmt.Sscanf(s.Text(), "MemAvailable: %d kB", &n); nItems == 1 {
```
89luca89, line 1225: This is giving some problems on Linux, can we revert this to the original and only modify the MacOS behaviour?

89luca89 (changes requested):

pkg/container/qemu_runner.go
89luca89: CI problem is unrelated to this, disregard my comment

mattmoor (approved):

PR #2006, opened 2025-05-22
fix: improve SSH command handling

89luca89:
Instead of using session.Run() we use session.Shell() and use session.Stdin to "write" the command in it.

Also, deduplicate clients, and only keep them as singletons.

Makes it more resilient to random EOFs

Avoid leaking qemu processes, kill by pid if things go wrong

Fix: https://github.com/chainguard-dev/prodsec/issues/170

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -674,6 +674,7 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 
 	// qemu-system-x86_64 or qemu-system-aarch64...
 	qemuCmd := exec.CommandContext(ctx, fmt.Sprintf("qemu-system-%s", cfg.Arch.ToAPK()), baseargs...)
+	clog.FromContext(ctx).Info("qemu: starting VM")
 	clog.FromContext(ctx).Debugf("qemu: executing - %s", strings.Join(qemuCmd.Args, " "))
 
 	outRead, outWrite := io.Pipe()
@@ -725,6 +726,7 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 
 	started := make(chan struct{})
 
+	clog.FromContext(ctx).Info("qemu: waiting for SSH")
 	go func() {
 		// one-hour timeout with a 500ms sleep
 		retries := 7200
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -65,6 +65,10 @@ const (
 	defaultDiskSize = "50Gi"
 )
 
+var sshClient *ssh.Client
+var workspaceClient *ssh.Client
+var qemuPID int
+
 type qemu struct{}
 
 // QemuRunner returns a Qemu Runner implementation.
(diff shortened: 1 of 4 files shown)


pkg/container/qemu_runner.go
```diff
@@ -1222,30 +1231,25 @@ func getAvailableMemoryKB() int {
 		for s.Scan() {
 			var n int
 			if nItems, _ := fmt.Sscanf(s.Text(), "MemTotal: %d kB", &n); nItems == 1 {
-				return n
+				return int64(n), nil
 			}
 		}
 	case "darwin":
-		cmd := exec.Command("sysctl", "hw.memsize")
+		cmd := exec.Command("sysctl", []string{"-n", "hw.memsize"}...)
```
egibs, line 1238: Good call.

After a bit of digging around, it looks like `hw.memsize_usable` is a separate field if we're interested in using this:
```
hw.memsize_usable: 37984092160
```

89luca89: my bad left changes that are superseded by https://github.com/chainguard-dev/melange/pull/2005#pullrequestreview-2862524419

removed them

pkg/container/qemu_runner.go
```diff
@@ -385,18 +380,17 @@ func (bw *qemu) TerminatePod(ctx context.Context, cfg *Config) error {
...
 		nil,
 		nil,
-		nil,
 		false,
 		[]string{"sh", "-c", "echo s > /proc/sysrq-trigger && echo o > /proc/sysrq-trigger&"},
 	)
 	if err != nil {
-		return err
+		// in case of graceful shutdown failure, axe it with pkill
+		return syscall.Kill(-qemuPID, syscall.SIGKILL)
```
mattmoor, line 393: We should check the exit code of this and return the error

pkg/container/qemu_runner.go
```diff
@@ -65,6 +65,10 @@ const (
 	defaultDiskSize = "50Gi"
 )
 
+var sshClient *ssh.Client
+var workspaceClient *ssh.Client
+var qemuPID int
```
mattmoor, line 70: I think we should make these fields on `qemu`.

My concern is that these potentially carry multiple values over the process lifetime if multiple test pods get run, and if we were smart enough to run them in parallel then they could end up racing and conflict!

pkg/container/qemu_runner.go
mattmoor: discussed offline.  Let's log it and then this is fine.

89luca89: added a warning

pkg/container/qemu_runner.go
89luca89: done

mattmoor (approved):

mattmoor:
I think that `glibc` for QEMU is getting killed due to resource consumption :-/

89luca89:
> I think that `glibc` for QEMU is getting killed due to resource consumption :-/

It might be that the heuristic to find free memory is reporting far too low memory to actually boot the test, so we need to add a lower bound for it (eg 4gb) I'd say if possible we first force-merge https://github.com/chainguard-dev/melange/pull/2005 then we can do the fix in this PR after rebase

PR #2008, opened 2025-05-27
fix: make checkout expected-commit bump less greedy

bentasker:

Prior to this commit, bumps updated `expected-commit` for any git checkout node who's `tag` attribute matched the regular expression `\${{vars\..+}}` (i.e. anything which used a variable).

This made the logic very greedy, clobbering values for unrelated checkouts (for example [here](https://github.com/wolfi-dev/os/pull/52449/commits/4bc6632241cd0487b22349344dcea4c25647ef7d))

We only want bumps to update those where `tag` matches either `package.version` or `vars.mangled-package-version`. 

Currently we use the following vars in `tag` attributes:
```
ben@cosmos:~/Repos/os$ grep -h 'tag: ${' *.yaml | grep -v 'package\.version' | cut -d: -f2 | sort | uniq
 ${{vars.docker-machine-tag}}
 ${{vars.exporter-tag}}
 ${{vars.mangled-package-version}}
```

It also adds a check to ensure that `expected-commit` does not contain a variable (i.e. `${{`) before overwriting it's value

### DIFF
--- a/pkg/renovate/bump/bump.go
+++ b/pkg/renovate/bump/bump.go
@@ -225,8 +225,7 @@ func updateGitCheckout(ctx context.Context, node *yaml.Node, expectedGitSha stri
 	if err != nil {
 		log.Infof("git-checkout node does not contain a tag, assume we need to update the expected-commit sha")
 	} else {
-		versionPattern := regexp.MustCompile(`\${{vars\..+}}`)
-		if !strings.Contains(tag.Value, "${{package.version}}") && !versionPattern.MatchString(tag.Value) {
+		if !strings.Contains(tag.Value, "${{package.version}}") && !strings.Contains(tag.Value, "${{vars.mangled-package-version}}") {
 			log.Infof("Skipping git-checkout node as it does not contain a version substitution so assuming it is not the main checkout")
 			return nil
 		}
--- a/pkg/renovate/bump/bump.go
+++ b/pkg/renovate/bump/bump.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"regexp"
 	"strconv"
 	"strings"
 
--- a/pkg/renovate/bump/bump.go
+++ b/pkg/renovate/bump/bump.go
@@ -236,7 +236,7 @@ func updateGitCheckout(ctx context.Context, node *yaml.Node, expectedGitSha stri
 	if expectedGitSha != "" {
 		// Update expected hash nodes.
 		nodeCommit, err := renovate.NodeFromMapping(withNode, "expected-commit")
-		if err == nil {
+		if err == nil && !strings.Contains(nodeCommit.Value, "${{") {
 			nodeCommit.Value = expectedGitSha
 			log.Infof("  expected-commit: %s", expectedGitSha)
 		}
--- a/docs/VAR-TRANSFORMS.md
+++ b/docs/VAR-TRANSFORMS.md
@@ -46,6 +46,21 @@ pipeline:
       uri: https://github.com/openjdk/jdk17u/archive/refs/tags/jdk-${{vars.mangled-package-version}}.tar.gz
 ```
 
+`mangled-package-version` can also be used with `git-checkout`:
+
+```yaml
+pipeline:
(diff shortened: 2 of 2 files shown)


pkg/renovate/bump/bump.go
```diff
@@ -237,7 +235,7 @@ func updateGitCheckout(ctx context.Context, node *yaml.Node, expectedGitSha stri
 	if expectedGitSha != "" {
 		// Update expected hash nodes.
 		nodeCommit, err := renovate.NodeFromMapping(withNode, "expected-commit")
-		if err == nil {
+		if err == nil && !strings.Contains(nodeCommit.Value, "${{") {
```
debasishbsws, line 238: Love this ❤️. Can we change it some how so that it also go through subpackage and test pipelines and update the commit SHA 
```sh
grep -rl 'tag:.*\${{package.version}}.*' . | xargs -I{} sh -c "grep -c 'tag:.*\${{package.version}}.*' '{}' | awk -v f='{}' '\$1>1{print f}'"
./cluster-api-1.9.yaml
./ruby3.2-concurrent-ruby.yaml
./py3-pydantic.yaml
./kserve-modelmesh-serving.yaml
./py3-cherrypy.yaml
./py3-portend.yaml
./openpmix.yaml
./ruby3.4-timers.yaml
./vendir.yaml
./ruby3.3-concurrent-ruby.yaml
./zig.yaml
./ruby3.2-net-imap.yaml
./py3-execnet.yaml
./ruby3.2-timers.yaml
./ruby3.3-net-imap.yaml
./py3-jaraco.text.yaml
./ruby3.2-prometheus-client.yaml
./sops.yaml
./ruby3.4-net-imap.yaml
./pgaudit-17.yaml
./opensearch-dashboards-3.yaml
./opensearch-3.yaml
./ruby3.3-timers.yaml
./py3-path.yaml
./py3-cheroot.yaml
./py3-jaraco.collections.yaml
./py3-pytest-xdist.yaml
./py3-autocommand.yaml
./ruby3.3-prometheus-client.yaml
./tinydir.yaml
./py3-webob.yaml
./ruby3.4-prometheus-client.yaml
./ruby3.4-concurrent-ruby.yaml
./efs-utils.yaml
```

In wolfi there are a lot of package where we do the same main git-checkout in the test pipeline to test our package. we need to update them manually every time.
Maybe not in this PR.

powersj:
@imjasonh @luhring - thoughts on this PR and who I should poke? This seems to keep coming up and more of our PRs would land if this were fixed up.

pkg/renovate/bump/bump.go
```diff
@@ -225,8 +224,7 @@ func updateGitCheckout(ctx context.Context, node *yaml.Node, expectedGitSha stri
 	if err != nil {
 		log.Infof("git-checkout node does not contain a tag, assume we need to update the expected-commit sha")
 	} else {
-		versionPattern := regexp.MustCompile(`\${{vars\..+}}`)
-		if !strings.Contains(tag.Value, "${{package.version}}") && !versionPattern.MatchString(tag.Value) {
+		if !strings.Contains(tag.Value, "${{package.version}}") && !strings.Contains(tag.Value, "${{vars.mangled-package-version}}") {
```
imjasonh, line 227: This seems to treat this variable name as special, where before it had only been a convention used in a lot of packages. Should we document this, and/or add a test case to document this for future generations?

bentasker: Seems reasonable - `mangled-package-version` is special cased here purely because it was a package with that var which caused the removed regex to be removed.

I'll go find the docs related to bump and add something there

bentasker: [Commit 97f1399](https://github.com/chainguard-dev/melange/pull/2008/commits/97f139990848486480c93f757567ce643df04009) adds docs.

It turns out there's already a test-case for bumping a `git-checkout` with `mangled-package-version` - https://github.com/chainguard-dev/melange/blob/97f139990848486480c93f757567ce643df04009/pkg/renovate/bump/bump_test.go#L110 (looks like it was added when the original regex was introduced)

PR #2009, opened 2025-05-28
Fix/respect additional mounts

89luca89:
Fix: https://github.com/chainguard-dev/prodsec/issues/216

We were not respecting additional mountpoints before
Now we will map them with 9p if present, and auto-discover them at the end of the boot if needed

Also, small fix in sendSSHCommand not returning the error value breaking the debug mode


antitree:
Chatting with team. I believe @egibs is going to test this locally before merging

egibs (approved):
Changes seem to work as advertised. I built one `ecosystems/gradle` package and two `ecosystems/maven` packges without any auth failures.

PR #2016, opened 2025-05-30
SBOM: add an OperatingSystem package to each apk SBOM

sil2100:
Work in progress. Things missing:

- Docker runner support
- Making sure tests pass
- Adding new tests

### DIFF
--- a/pkg/config/release_data.go
+++ b/pkg/config/release_data.go
@@ -1,0 +1,65 @@
+// Copyright 2025 Chainguard, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package config
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"strings"
+
+	apko_build "chainguard.dev/apko/pkg/build"
+)
+
+// This is a copy of the ReleaseData related code from apko's pkg/build/sbom.go
+
+// ParseReleaseData parses the information from /etc/os-release
+//
+// If no os-release file is found, it returns a Data struct with ID set to "unknown".
+// TODO: this should best be imported from apko, but right now this function is not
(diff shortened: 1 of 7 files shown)


pkg/config/release_data.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// This is a copy of the ReleaseData related code from apko's pkg/build/sbom.go
+
+// ParseReleaseData parses the information from /etc/os-release
+//
+// If no os-release file is found, it returns a Data struct with ID set to "unknown".
+// TODO: this should best be imported from apko, but right now this function is not
+// exported and not ready to be used outside of apko.
+func ParseReleaseData(osRelease io.Reader) (*apko_build.ReleaseData, error) {
```
sergiodj: WDYT about converting this into a helper function that lives inside apko, and is just used by melange?  I'm worried of code duplication and maintainability, mostly.

sil2100: Yes! That was the end goal - if you see the TODO: it mentions we should just import this from apko. Since we're no longer under time pressure, let me do that.

sergiodj: Thanks, @sil2100!

go.mod
```diff
@@ -3,7 +3,7 @@ module chainguard.dev/melange
 go 1.24.0
 
 require (
-	chainguard.dev/apko v0.27.9
+	chainguard.dev/apko v0.27.10-0.20250623150342-bfcb016b5959
```
sergiodj, line 6: Out of curiosity, is this correct?  I'm not sure whether we should release apko first and then depend on the released version here, or if it's OK to just depend on unreleased commits.  Genuinely curious!

sergiodj (changes requested):
This is much better now, thanks @sil2100!

Some tests are failing which is blocking the merge, but I think it's just a matter of adjusting expected outputs.

go.mod
sil2100: In general it's of course better if we tag a release, but by looking through history of the `go.mod` file I saw even recent (in March) instances of pulling in non-released apko versions, which is why I decided to just go for it!

sergiodj: That makes sense, thanks!

sergiodj:
> One of the failures demonstrated an actual situation that needed handling. Now it should look better!

Hm, some tests are still failing :-(.

sil2100:
> > One of the failures demonstrated an actual situation that needed handling. Now it should look better!
> 
> Hm, some tests are still failing :-(.

I *think* those are unrelated. If you take a look at the failures, they fail due to `No rule to make target 'package/logstash-8'.  Stop.` and similar. It probably needs a different fix, possibly separate PR?

PR #2028, opened 2025-06-05
sbom: generate downloadLocations for unknown gits

javacruft:
Use tarballs.chainguard.dev for the downloadLocation for git sources
which are not github or gitlab based; these will be populated
automatically by a post commit job that generates and uploads the
tarballs to the detailed location.


javacruft:
flakey access to an external resource for the hello-world tarball.

PR #2029, opened 2025-06-09
split/manpages: Manpage path expansion for better lookup  

AbduAzim424:
Add support for relocating man pages from two additional directories

### DIFF
--- a/pkg/build/pipelines/split/manpages.yaml
+++ b/pkg/build/pipelines/split/manpages.yaml
@@ -17,21 +17,17 @@ pipeline:
         PACKAGE_DIR="${{targets.outdir}}/${{inputs.package}}"
       fi
 
-      if [ "$PACKAGE_DIR" == "${{targets.contextdir}}" ]; then
+      if [ "$PACKAGE_DIR" = "${{targets.contextdir}}" ]; then
         echo "ERROR: Package can not split files from itself!" && exit 1
       fi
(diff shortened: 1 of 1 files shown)


pkg/build/pipelines/split/manpages.yaml
```diff
@@ -25,3 +25,13 @@ pipeline:
         mkdir -p "${{targets.contextdir}}/usr/share"
         mv "$PACKAGE_DIR/usr/share/man" "${{targets.contextdir}}/usr/share"
       fi
+
+      if [ -d "$PACKAGE_DIR/usr/local/share/man" ]; then
```
sergiodj: Can you convert these 3 blocks into a `for` loop?  I think it's easier to read and to expand later.

Also, while at it, you might want to fix the comparison operator being used in line 20 (shell's `[` uses `=` for comparison, not `==`).

sergiodj (changes requested):
Thanks, @AbduAzim424.

I'm leaving a small request to make the code easier to maintain, but otherwise it looks good.

pkg/build/pipelines/split/manpages.yaml
AbduAzim424: WIll do. Thank you

sergiodj (approved):
LGTM, thanks!

PR #2030, opened 2025-06-09
Add split/bin pipeline

dannf:

### DIFF
--- a/pkg/build/pipelines/split/bin.yaml
+++ b/pkg/build/pipelines/split/bin.yaml
@@ -1,0 +1,38 @@
+name: Split executable files
+
+needs:
+  packages:
+    - busybox
+
+inputs:
+  package:
+    description: |
+      The package to split executable files from
+    required: false
+
+pipeline:
+  - runs: |
+      pkgname="${{context.name}}"
+      pkgnobin="${pkgname%%-bin}"
+      pkgnobindir="${{targets.outdir}}/$pkgnobin"
+
+      if [ -n "${{inputs.package}}" ]; then
+        PACKAGE_DIR="${{targets.outdir}}/${{inputs.package}}"
+      elif [ "$pkgname" != "$pkgnobin" ] && [ -d "$pkgnobindir" ]; then
+        PACKAGE_DIR="$pkgnobindir"
+      else
+        PACKAGE_DIR="${{targets.destdir}}"
+      fi
+
+      if [ "$PACKAGE_DIR" == "${{targets.contextdir}}" ]; then
+        echo "ERROR: Package can not split files from itself!"
+        exit 1
+      fi
+
+      dir=usr/bin
+      if [ -e "$PACKAGE_DIR/$dir" ] || [ -L "$PACKAGE_DIR/$dir" ]; then
+        d="${{targets.contextdir}}/${i%/*}" # dirname $dir
+        mkdir -p "$d"
+        mv "$PACKAGE_DIR/$dir" "$d"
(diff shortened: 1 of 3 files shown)


pkg/build/pipelines/split/bin.yaml
```diff
@@ -0,0 +1,38 @@
...
+      fi
+
+      if [ "$PACKAGE_DIR" == "${{targets.contextdir}}" ]; then
+        echo "ERROR: Package can not split files from itself!"
+        exit 1
+      fi
+
+      dir=usr/bin
+      if [ -e "$PACKAGE_DIR/$dir" ] || [ -L "$PACKAGE_DIR/$dir" ]; then
+        d="${{targets.contextdir}}/${i%/*}" # dirname $dir
```
smoser: `$i` came from the loop that I had you drop. sorry. it is not defined here.

pkg/build/pipelines/split/bin.yaml
```diff
@@ -0,0 +1,38 @@
...
+        PACKAGE_DIR="${{targets.destdir}}"
+      fi
+
+      if [ "$PACKAGE_DIR" == "${{targets.contextdir}}" ]; then
+        echo "ERROR: Package can not split files from itself!"
+        exit 1
+      fi
+
+      dir=usr/bin
+      if [ -e "$PACKAGE_DIR/$dir" ] || [ -L "$PACKAGE_DIR/$dir" ]; then
```
smoser: `-L` here seems wierd.  it only matters if bin/ is a dangling symlink.  And if it is a symlink.. what does the 'mv' do?

maybe its best to just do something like:
```
if [ ! -d "$PACKAGE_DIR/$dir" ]; then
   echo "$PACKAGE_DIR/$dir: not a directory"
   exit 1
fi

File: pkg/build/pipelines/split/bin.yaml [hunk 3]
```diff
@@ -0,0 +1,38 @@
...
+        d="${{targets.contextdir}}/${i%/*}" # dirname $dir
+        mkdir -p "$d"
+        mv "$PACKAGE_DIR/$dir" "$d"
```
smoser: would it make more sense to move the contents of usr/bin/* rather than usr/bin itself?  That'd allow you to use this pipeline after something else had already created $d.

If you dont want to do that, then lets fail if $d exists. othereise you'll end up with the `mv` here just moving a 'bin' into usr/bin resulting in `usr/bin/bin/stuff-here`.


if you _do_ want to move the content of usr/bin rather than the directory itself (`mv $PACKAGE_DIR/$dir`)

smoser (commented):
I feel like its best to move the contents of the directory rather than the directory.  But maybe I'm not thikning of something.

Also, if the source dir doesn't exist, lets error rather than just ending up creating an empty package or some other non-expected behavior.

pkg/build/pipelines/split/bin.yaml
dannf: Apologies. -ERR_EOD_NO_TESTING

pkg/build/pipelines/split/bin.yaml
dannf: yeah, i wondered the same thing - kept it for consistency w/ `split/dev`. I'm forgetting about `split/dev` now.

pkg/build/pipelines/split/bin.yaml
dannf: +1

PR #2032, opened 2025-06-10
fix: pass cap add/drop to ssh session

89luca89:
Pass caps add/drops via env variables

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -208,8 +208,16 @@ func (bw *qemu) Debug(ctx context.Context, cfg *Config, envOverride map[string]s
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/container/qemu_runner.go
```diff
@@ -1369,6 +1390,10 @@ func generateCpio(ctx context.Context, cfg *Config) (string, error) {
 		Contents: apko_types.ImageContents{
 			BuildRepositories: []string{
 				"https://apk.cgr.dev/chainguard",
+				"/home/luca_dimaio_chainguard_dev/Projects/chainguard/wolfi-os/packages/",
```
egibs: Leftover testing artifacts?

89luca89: sadly yes removing

PR #2033, opened 2025-06-10
Add a linter for CUDA driver-specific libraries

dannf:

### DIFF
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -194,6 +194,11 @@ var linterMap = map[string]linter{
 		Explain:         "Move binary to /usr/bin",
 		defaultBehavior: Require,
 	},
+	"cudaruntimelib": {
+		LinterFunc:      allPaths(cudaDriverLibLinter),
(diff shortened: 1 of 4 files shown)


pkg/linter/linter.go
```diff
@@ -194,6 +194,11 @@ var linterMap = map[string]linter{
 		Explain:         "Move binary to /usr/bin",
 		defaultBehavior: Require,
 	},
+	"cudaruntimelib": {
+		LinterFunc:      allPaths(cudaDriverLibLinter),
+		Explain:         "CUDA driver-specific libraries should be passed into the container by the host. If this library is needed for build-time linking or ldd-check tests, please include a stub library during build or test instead.",
```
murraybd: Personally, I don't know if this is enough information for me to fix the warning. Is there further documentation or an example to which you can point?

murraybd (approved):

pkg/linter/linter.go
dannf: i've added more specific instructions

PR #2035, opened 2025-06-10
Delete named pipes instead of broadly excluding fifo files

egibs:
We're currently excluding `fifo` files by name which is preventing certain legitimate files and directories that are needed in the emitted APK.

This PR deletes any named pipe files before we try to archive the workspace instead of potentially excluding valid directories or files.


stevebeattie (approved):
Confirmed that both dpkg (the original reason for excluding fifos) and grafana-image-renderer build and test correctly, excluding files of type fifo without excluding files with fifo in their names.

PR #2036, opened 2025-06-11
Test package to test for file ownership in a package

zmarano:
From @smoser https://github.com/chainguard-dev/customer-issues/issues/2294#issuecomment-2963329058

### DIFF
--- a/e2e-tests/file-permissions-build-test.yaml
+++ b/e2e-tests/file-permissions-build-test.yaml
@@ -1,0 +1,91 @@
+package:
+  name: test-permissions
+  version: "1.0.0"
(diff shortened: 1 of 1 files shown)


smoser:
Why did you close this?

The test would work with privileged bubblewrap (and docker) before the regression.

It would not work as non-priv in either case.

Also, don't know if you inadvertantly pushed over my updated file https://github.com/chainguard-dev/melange/blob/2144c651f2cccc760ef0151506b77b2767a2db3d/e2e-tests/testperms-build-test.yaml .

(you are correct that it should be 'testownership').

Issue #2041, opened 2025-06-12
make integration is not run?


smoser:
I put up #2039 and it passed c-i.
The test I tried to add to `make integration` does not seem to have failed, 
but it is broken. I used `${{target.destdir}` -- singular rather than plural 
in the added `pkg/build/testdata/build_configs/bogus-version.yaml` file.

So i think nothing actually runs 'make integration' during the c-i.

If I fix the typo, and try to run it locally, dropping the `-race` and running like:

    go test -v -run ^TestBuild_BuildPackage$ ./... -tags=integration 2>&1 | tee integration.log

It takes a very long time to run (several minutes now).

smoser:
It seems the "really long teime" is likely related to all the things that are currently down.

I'll come back when there is a better chance of this working.

smoser:
fixed by https://github.com/chainguard-dev/melange/pull/2042

PR #2042, opened 2025-06-13
Fix integration tests and then enable them.

smoser:
* Move version check to Build.BuildPackage for easier test.
   Having the version check in the cli meant I could not actually
   test it from the integration tests.  
   Moving it slightly later makes it more testable.

* integration test - slightly better log message.

* go-tests workflow - run integration tests, a superset of unit
   integration tests (make integration) runs a superset of unit tests.
   Nothing was running the integration tests.

* Fix typo in bogus-version.yaml

Fixes: https://github.com/chainguard-dev/melange/pull/2041

### DIFF
--- a/.github/workflows/go-tests.yaml
+++ b/.github/workflows/go-tests.yaml
@@ -26,5 +26,5 @@ jobs:
         with:
           go-version-file: './go.mod'
 
-      - name: Unit Tests
+      - name: Integration and Unit Tests
         run: make integration


smoser:
I verified that nothing is running integration tests in commit 7d3abe1be8510077b968c6f92895b4d995698bc3 .  The 'Go Tests' [action](https://github.com/chainguard-dev/melange/actions/runs/15623811882) was successful as well as all the other workflows other that 'Build packages' and 'Test melange test'. 

So I'm now going to push commit a9520e5bf4f5f43df1965c028b3953aadf30229f that runs integration test and it should fail.

PR #2043, opened 2025-06-13
fix: avoid sending huge command when fetching extra files from workspace

89luca89:
Right now to fetch extra files from the workspace (mainly licenses) we simply append the list of files to the `tar` command.

This will lead to "-ash: sh: Argument list too long" for cases where the list of arguments surpasses `MAX_ARG_STRINGS`

Instead we write the list to a file, and use the `-T/--files-from` option in order to populate the tar from that list. To avoid network problems, also the streaming of the list is divided in small chunks.


sil2100 (approved):
This looks good and works as expected. Thank you Luca!

PR #2049, opened 2025-06-16
fix: avoid SSH lockups, use script for tests

89luca89:
Using script to send our test commands, will setup a new tty, in order this
will avoid SSH hanging when stdout/err pipes are open, unifying the
behaviour of backgorund process with the other solutions

This will not require changes in packages, but will result in stdout and
stderr to be merged in melange output

This change only involves test runs, so we introduce a new config entry
to be able to discern between build and test runs.

### DIFF
--- a/pkg/build/test.go
+++ b/pkg/build/test.go
@@ -141,9 +141,10 @@ func (t *Test) BuildGuest(ctx context.Context, imgConfig apko_types.ImageConfigu
 	}
 	defer os.RemoveAll(tmp)
 
+	// see qemu_runner.go: 1194
 	if t.Runner.Name() == container.QemuName {
 		t.ExtraTestPackages = append(t.ExtraTestPackages, []string{
-			"util-linux-misc",
+			"cmd:script",
 		}...)
 	}
 


pkg/build/pipeline.go
```diff
@@ -182,6 +182,7 @@ func buildEvalRunCommand(pipeline *config.Pipeline, debugOption rune, workdir st
 [ -d '%s' ] || mkdir -p '%s'
 cd '%s'
 %s
+[ -n "$(jobs -p)" ] && kill $(jobs -p)
```
smoser: negative logic is harder to read, but with shell it is the rigth thing to use.

if for some reason, 'jobs -p' fails and doesn't output anything, then you dont kill any jobs. 
i can agree that might sound unlikely.  however, doing it this way means you swallow errors kill.

heres an example:

```
$ sh -ec '
> sleep 100s &
> [ -n "$(jobs -p)" ] && kill adf $(jobs -p)
> echo still here, gonna exit 0 now.
> exit 0
> '
still here, gonna exit 0 now.
```

smoser: Better is to do something like this:

```
set -e
jobs=$(jobs -p)
if [ -n "$jobs" ]; then
  echo "killing jobs $jobs"
  kill -9 $jobs
fi
exit 0
```

this way `set -e` will catch the failures at each step.

also, re-enable `set -e` becuse the executed code block could have disabled it.

huh... i guess that code block could also have exited zero, which would mean this code would never execute. that needs fixing as well.  `trap` would be an option, but i know some of the tests and things use trap themselves. (sorry, probably most are from me).

antitree:
`nginx-prometheus-exporter-fips`

>2025/06/16 15:43:46 INFO > time=2025-06-16T19:43:45.535Z level=ERROR source=nginx.go:57 msg="error getting stats" error="failed to get http://127.0.0.1:8080/stub_status: Get \"http://127.0.0.1:8080/stub_status\": round trip failed: dial tcp 127.0.0.1:8080: connect: connection refused"

pkg/build/pipeline.go
smoser: also, the location of this code (in "generic" vs in ssh) will affect other runners.

89luca89: I'm changing to the if/else like you suggested 
this should not be a problem with other runners, as this is what happens anyway with the one-shot execution

89luca89: also moved to trap, this trap is executed both first and last so ensure this cleanup is triggered

89luca89:
> `nginx-prometheus-exporter-fips`
> 
> > 2025/06/16 15:43:46 INFO > time=2025-06-16T19:43:45.535Z level=ERROR source=nginx.go:57 msg="error getting stats" error="failed to get http://127.0.0.1:8080/stub_status: Get "[http://127.0.0.1:8080/stub_status\](http://127.0.0.1:8080/stub_status%5C)": round trip failed: dial tcp 127.0.0.1:8080: connect: connection refused"



> > `nginx-prometheus-exporter-fips`
> > > 2025/06/16 15:43:46 INFO > time=2025-06-16T19:43:45.535Z level=ERROR source=nginx.go:57 msg="error getting stats" error="failed to get http://127.0.0.1:8080/stub_status: Get "[http://127.0.0.1:8080/stub_status](http://127.0.0.1:8080/stub_status%5C)": round trip failed: dial tcp 127.0.0.1:8080: connect: connection refused"
> 
> @89luca89 -- I think we'll need to append the new logic to the final pipeline step, otherwise we'll terminate jobs that subsequent pipelines rely on.

This is happening for me with the `main` branch too, so I don't think it's related to this change

89luca89:
As suggested by @smoser now the `script` workaround only involves the test runs, leaving intact the build runs
@egibs  this should unblock the whole thing

pkg/build/test.go
```diff
@@ -141,6 +141,12 @@ func (t *Test) BuildGuest(ctx context.Context, imgConfig apko_types.ImageConfigu
 	}
 	defer os.RemoveAll(tmp)
 
+	if t.Runner.Name() == container.QemuName {
+		t.ExtraTestPackages = append(t.ExtraTestPackages, []string{
+			"util-linux-misc",
```
smoser: Am I right that this pollutes the test runner somewhat heavily (see below) and arbitrarily.  Would _every_ test invocation now have all of the listed packages present?  That makes for a lot of assumptions and also will slow things down (several  test environments only specify `busybox`).

Is it not possible to run `script` from the outside of the test environment here?  It still is very heavy for what it is, but it would at least would not pollute the test environment and would be less icky to change later.g

```
# apk add util-linux-misc
fetch https://apk.cgr.dev/chainguard/x86_64/APKINDEX.tar.gz
(1/11) Installing setarch (2.41.1-r0)
(2/11) Installing libblkid (2.41.1-r0)
(3/11) Installing libuuid (2.41.1-r0)
(4/11) Installing libfdisk (2.41.1-r0)
(5/11) Installing sqlite-libs (3.50.1-r0)
(6/11) Installing util-linux (2.41.1-r0)
(7/11) Installing libmount (2.41.1-r0)
(8/11) Installing ncurses-terminfo-base (6.5_p20250621-r0)
(9/11) Installing ncurses (6.5_p20250621-r0)
(10/11) Installing libsmartcols (2.41.1-r0)
(11/11) Installing util-linux-misc (2.41.1-r0)
Executing busybox-1.37.0-r46.trigger
OK: 42 MiB in 26 packages
```

smoser (commented):
I think the answer to my question is "not easily", but I thought I would ask.
This is really kind of like when we had ssh added to the build/test environment.  ultimately we removed it, and that was a good thing :tm:.

pkg/build/test.go
89luca89: we could alternatively use `ForceCommand` for SSH, and manually use script and chroot in a wrapper
I'm not sure if using script outside of the chroot will lead to the same results, but I will try :+1:

89luca89: Splitting script: https://github.com/wolfi-dev/os/pull/57648

89luca89: Ok now with 9ac39836bddea17c027484533b03562704506c24 we use `cmd:script` and the new script subpackage will be automatically pulled instead of utils-linux

Marking this as solved

PR #2050, opened 2025-06-17
license: allow grouping of licenses with AND or OR clauses

sil2100:
The world of licensing is complex. Some software is dual-licensed, some requires following multiple licenses. Let's add the ability to group licenses by AND and OR groupings. This is all optional.

Still a draft. Need to address the license checker and license rennovator.


sil2100:
Should we go more with something like this?

```
- group: multi-licensed
  licenses:
  - ...
- group: choose-your-license
  licenses:
  - ...
```

PR #2051, opened 2025-06-18
feat: add SLSA provenance generation for package builds

egibs:
This PR adds opt-in SLSA provenance generation for package builds and is consistent with what we've implemented elsewhere. 

By default, we're leaving this off while we land on the final structure. That said, CI will use the new flag (`--generate-provenance`) to validate changes and, of course, local builds can also use this.

The resulting provenance statement is stored in a separate `.attest.tar.gz` file and is signed in the same way as the APK. This new tarball contains the `.attestation` file for the package and the public signing key.


dannf (commented):
I haven't reviewed the code in any detail, but I know we have at least one test that makes assumptions about the contents of packages:
https://github.com/wolfi-dev/os/blob/main/pipelines/test/docs.yaml#L21

We should probably add a `pipeline-utils` package to `tw` that has a `tw-list-apk-contents` command that hides sbom/provenance (with maybe a `--show-melange-generated` to unhide them)

egibs:
> I haven't reviewed the code in any detail, but I know we have at least one test that makes assumptions about the contents of packages: [wolfi-dev/os@`main`/pipelines/test/docs.yaml#L21](https://github.com/wolfi-dev/os/blob/main/pipelines/test/docs.yaml#L21)
> 
> We should probably add a `pipeline-utils` package to `tw` that has a `tw-list-apk-contents` command that hides sbom/provenance (with maybe a `--show-melange-generated` to unhide them)

Ack. The new file lives next to `.PKGINFO` and `.melange.yaml` so it's not part of the emitted package like SBOMs are. I'll verify that locally, though.

egibs:
> I haven't reviewed the code in any detail, but I know we have at least one test that makes assumptions about the contents of packages

This is now a non-issue with the changes in [`e5c67e8` (#2051)](https://github.com/chainguard-dev/melange/pull/2051/commits/e5c67e8a27b0eac32e2d98cf470e8ba16374c410). Storing the provenance next to the package is [called out](https://slsa.dev/spec/v1.1/distributing-provenance#where-attestations-are-published) by SLSA as well.

sil2100:
So this is looking good, I'm +1 on getting this merged. However: I'm a bit worried about all the failing tests. The `pkg/container/qemu_runner.go:496:3: too many arguments in call to sendSSHCommand` error comes up a lot and I'm worried that this might be masking any potential issues.

Can you see what's that about? If it's unrelated, can you make sure that all the tests are run and there are actually no failures? Thanks!

sil2100 (approved):
Okay, I actually see this failure on my PR as well. Approving this, but please make sure the tests are passing (at least locally) before pressing merge! Thanks!

PR #2052, opened 2025-06-19
fix: avoid fetching hardlinks in licenses

89luca89:
Hardlinks will not behave well when uncompressing the fetched archive on the host, so we avoid them, they're duplicate files anyway


maxgio92 (approved):

maxgio92:
LGMT

PR #2053, opened 2025-06-23
Add usrmerge linter failure on files in lib

smoser:
After usrmerge-lib finished (2025-06-23), files are not allowed to be in /lib.


zmarano (approved):

zmarano (dismissed):

pkg/linter/linter.go
```diff
@@ -805,14 +805,17 @@ func usrmergeLinter(ctx context.Context, _ *config.Configuration, _ string, fsys
 			}
 		}
 
-		if path == "sbin" || path == "bin" || path == "usr/sbin" {
+		if path == "sbin" || path == "bin" || path == "usr/sbin" ||
```
a-crate: Add `lib` and `lib64` to the list of directories we shouldn't skip on line 802.

smoser: ack. slightly cleaned up. it passes test, but please do think through it.
the logic is non-obvious.

a-crate (approved):

PR #2057, opened 2025-06-27
licensing: default to concatenating with AND, add simple mode  for license-check --fix

sil2100:
Rationale: even with our current license checking mechanism, listing out all the licenses properly will end up concatenating them in the package with an OR statement. This is bad. Think of GPL software containing some vendored MIT code - its SBOM will end up with a license string of "GPLv3 OR MIT", which is untrue and potentially dangerous as the code cannot be redistributed with a less restrictive license. Defaulting to AND is a safer bet.

There is an ongoing design and implementation effort to add ability to express relations to licensing: https://github.com/chainguard-dev/melange/pull/2050 . But this is a big change and needs an actual design doc I'm afraid.

In the meantime, switching to AND seems a good bet, and I added a renovate `--simple` option for the `--fix` mode that doesn't list all the licenses individually (with links to files) but simply deals with listing out just the licenses in one string. It's something we'll want to modify and do properly (listing all the files), but before we do, this is the only safe way of ensuring the person working on the licensing bits has power over relationships between the licenses listed out. `--fix` just gives a guideline, nothing more.

### DIFF
--- a/pkg/cli/license_check.go
+++ b/pkg/cli/license_check.go
@@ -32,6 +32,7 @@ import (
 func licenseCheck() *cobra.Command {
 	var workDir string
 	var fix bool
+	var simple bool
 	cmd := &cobra.Command{
 		Use:     "license-check file",
 		Short:   "Gather and check licensing data",
@@ -81,6 +82,7 @@ func licenseCheck() *cobra.Command {
 					ctx,
 					copyright.WithLicenses(detectedLicenses),
 					copyright.WithDiffs(diffs),
+					copyright.WithSimple(simple),
 				)
 				err = rc.Renovate(cmd.Context(), copyrightRenovator)
 			}
@@ -91,6 +93,7 @@ func licenseCheck() *cobra.Command {
 
 	cmd.Flags().StringVar(&workDir, "workdir", "", "path to the working directory, e.g. where the source will be extracted to")
 	cmd.Flags().BoolVar(&fix, "fix", false, "fix license issues in the melange yaml file")
+	cmd.Flags().BoolVar(&simple, "simple", false, "use a simple copyright format with all detected licenses joined together")
 
 	return cmd
 }
--- a/pkg/renovate/copyright/copyright.go
+++ b/pkg/renovate/copyright/copyright.go
@@ -17,6 +17,9 @@ package copyright
 import (
 	"context"
 	"fmt"
+	"maps"
+	"slices"
+	"strings"
 
 	"github.com/chainguard-dev/clog"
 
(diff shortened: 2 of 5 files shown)


pkg/cli/license_check.go
```diff
@@ -91,6 +93,7 @@ func licenseCheck() *cobra.Command {
 
 	cmd.Flags().StringVar(&workDir, "workdir", "", "path to the working directory, e.g. where the source will be extracted to")
 	cmd.Flags().BoolVar(&fix, "fix", false, "fix license issues in the melange yaml file")
+	cmd.Flags().BoolVar(&simple, "simple", false, "use a simple copyright format with all detected licenses joined together")
```
dannf: To me it would make sense to add a `--format [simple|flat]`, where "flat" remains the default. This way we could continue to extend it. That would also somehow force us to document what flat format is, which I think might be helpful for the user.

pkg/renovate/copyright/copyright.go
```diff
@@ -104,46 +118,98 @@ func New(ctx context.Context, opts ...Option) renovate.Renovator {
...
+				return err
 			}
-
-			licenseNode := &yaml.Node{
-				Kind:    yaml.MappingNode,
-				Style:   yaml.FlowStyle,
-				Content: []*yaml.Node{},
+		} else {
+			// Use flat license listing (original behavior)
+			if err = populateCopyright(ctx, copyrightNode, ccfg.Licenses); err != nil {
```
dannf: Could we rename this to `populateFlatCopyright` or similar?

pkg/renovate/copyright/copyright.go
```diff
@@ -53,6 +57,16 @@ func WithDiffs(diffs []license.LicenseDiff) Option {
 	}
 }
 
+// WithSimple sets whether the copyright should be populated with a single
+// node containing all detected licenses joined together, or with multiple
+// nodes, one per license.
```
dannf: That's a good explanation of the differences - it would be nice to have in the help output.

dannf (dismissed):
Nice! I've a suggestion inside about the flag name, but it's non-blocking

pkg/cli/license_check.go
sil2100: I like that! Let me go with that in that case. Let's merge it afterwards.

sil2100:
@dannf pushed some changes on top. The flag help description could use some expansion, but I'd recommend doing that as part of future changes. It's all very experimental right now!

dannf (approved):
LGTM!

PR #2064, opened 2025-07-02
feat: allow nested symlinks in workspaces

a-crate:
### Functional Changes

Notes:

This change is newly-capable of building packages such as [this pr](https://github.com/chainguard-dev/enterprise-packages/pull/12886) which share workspaces and files within that workspace via symlinks.

This already worked on Mac for some reason, which is probably another bug.


e2e-tests/symlinks-in-workspace-build.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+environment:
+  contents:
+    packages:
+      - busybox
+
+pipeline:
+  - name: Test for symlink presence in workspace
+    runs: |
+      cat testdata-symlink.txt
```
justinvreeland: Do we need to compare the data to to make sure both parts were copied correctly or it points correctly or is the issue soley that it doesn't copy the simlink file and if it's a broken link it was inappropriate/broken in melange?

justinvreeland (approved):

e2e-tests/symlinks-in-workspace-build.yaml
a-crate: I think the latter. All we want to know is that melange copied the link. However, melange will only copy it if it evaluates to a regular file so it's nice to also validate that the symlink still resolves in the workspace too.

Ideally we would copy all symlinks and not care if they are broken or where the evaluate but copying symlinks to non-regular files (specifically the top-level workspace dir `.` being a symlink) causes problems here, so we have to care about it.

We might as well compare the data as long as we're doing this though.

PR #2070, opened 2025-07-07
Avoid running cachedir mount commands via script

egibs:
We currently run all QEMU commands via `script` when running tests. This does not work for the cache directory mount commands:
```
2025/07/07 16:46:49 INFO qemu: setting up melange cachedir: /tmp/melange-cache
2025/07/07 16:46:49 WARN ash: script: not found
2025/07/07 16:46:49 ERRO Failed to run command "script -f -q --log-in /dev/null --log-out /dev/null -e -c 'sh -c '\\''mkdir -p /var/cache/melange /mount/var/cache/melange /mount/upper /mount/work && mount -t 9p melange_cache /var/cache/melange && mount -t overlay overlay -o lowerdir=/var/cache/melange,upperdir=/mount/upper,workdir=/mount/work /mount/var/cache/melange'\\'": Process exited with status 127
```

This PR avoids running the initial cache directory mount commands via `script` for tests while allowing for all other commands.

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -59,6 +59,8 @@ import (
 
 var _ Debugger = (*qemu)(nil)
 
+var setupMountCommand string
+
 const QemuName = "qemu"
 
 const (
@@ -843,7 +845,7 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 
 	if cfg.CacheDir != "" {
 		clog.FromContext(ctx).Infof("qemu: setting up melange cachedir: %s", cfg.CacheDir)
-		setupMountCommand := fmt.Sprintf(
+		setupMountCommand = fmt.Sprintf(
 			"mkdir -p %s %s /mount/upper /mount/work && mount -t 9p melange_cache %s && "+
 				"mount -t overlay overlay -o lowerdir=%s,upperdir=/mount/upper,workdir=/mount/work %s",
 			DefaultCacheDir,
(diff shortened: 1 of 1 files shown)


stevebeattie:
I think this is the right approach, but it seems possible in the future that we'd have potentially other setup commands to run outside of the chroot environment under `/mount` -- I wonder if it'd be better to have either an argument to the function or variant of it that is for environment setup that causes the injection of the `script ...` bits to be avoided, rather than specifically hard-coding it for the overlay mount command.

egibs:
> I think this is the right approach, but it seems possible in the future that we'd have potentially other setup commands to run outside of the chroot environment under `/mount` -- I wonder if it'd be better to have either an argument to the function or variant of it that is for environment setup that causes the injection of the `script ...` bits to be avoided, rather than specifically hard-coding it for the overlay mount command.

[`2d4719d` (#2070)](https://github.com/chainguard-dev/melange/pull/2070/commits/2d4719d6d2b1eb3161395b7e79aee23a6e750a4a) is a better attempt that only sets `TestRun` at the pipeline level after we run `StartPod`.

e.g.,
```
cmd: sh -c 'mkdir -p /var/cache/melange /mount/var/cache/melange /mount/upper /mount/work && mount -t 9p melange_cache /var/cache/melange && mount -t overlay overlay -o lowerdir=/var/cache/melange,upperdir=/mount/upper,workdir=/mount/work /mount/var/cache/melange' test run: false
2025/07/07 19:05:58 INFO qemu: setting up local workspace
cmd: sh -c 'find /mnt/ -mindepth 1 -maxdepth 1 -exec cp -a {} /home/build/ \;' test run: false                                                                        2025/07/07 19:05:58 INFO running the main test pipeline
cmd: /bin/sh -c 'set -ex
[ -d '\''/home/build'\'' ] || mkdir -p '\''/home/build'\''
cd '\''/home/build'\''
find / -name "script"                                                                                                                                                   cd "${ZK_HOME}"
cp conf/zoo_sample.cfg conf/zoo.cfg
bin/zkServer.sh start-foreground &                                                                                                                                      ZOOKEEPER_PID=$!
bin/zkCli.sh | grep "Welcome to ZooKeeper!"
bin/zkCli.sh create /test-$(uuidgen) '\''$(uuidgen)'\''                                                                                                                 kill "${ZOOKEEPER_PID}"

exit 0' test run: true

89luca89 (approved):

PR #2071, opened 2025-07-08
Revert "feat: allow symlinks in workspaces (#2064)"

a-crate:
This reverts commit 75ee8c561e307394b4b565e6e7b23ce7cf059245.

No packages depending on this were merged.


a-crate:
Reason for revert is that this does not work well. It copies symlink target contents into the new workspace (resolving at workspace population time) rather than copying the symlink as a symlink. I did prototype just fixing that behavior and copying symlinks as symlinks correctly, but this causes some issues when moving the workspace folder around later. This is solvable somehow, but I decided to just revert the broken feature for now and implement it correctly later.

PR #2072, opened 2025-07-08
SCA: Generate `depends` for shlibs ending in `.so`

sergiodj:
It turns out that this feature uncovered another bug in melange, which
is also being addressed by this commit.

When a package X ships a vendored shared library, and a subpackage Y
links against said library, melange fails to detect that the shlib is
vendored and generates a "depends" for it on Y, which is wrong because
X will never have a "provides" for said library.  I believe we hadn't
come accross this problem before because vendored libraries tend not
to be versioned (i.e., their names end in ".so"), which melange
happily ignored until now.  The underlying cause of this problem is
the fact that melange performs SCA for each (sub)package
independently, without taking into consideration everything that's
being built by the package.

I implemented a way to detect vendored libraries on the package and
all of its subpackages during SCA, and if melange detects that a
subpackage Y depends on a vendored library provided by its parent
package X, then a "depends" will be generated (in the form
"X=<version").  It is possible to disable this feature by setting the
"no-vendored-cross-package-deps" option to true.

Fixes: https://github.com/chainguard-dev/internal-dev/issues/7609


sergiodj:
Related to #2067

sergiodj:
Converting to draft because I have to think about whether `isInDir` needs to be used in some places here.

sergiodj:
Kudos to @jonjohnsonjr for helping with the design here.

jonjohnsonjr (approved):

PR #2073, opened 2025-07-08
Use ftpmirror.gnu.org whenever applicable

sergiodj:
We should use ftpmirror.gnu.org instead of ftp.gnu.org whenever applicable.


sergiodj:
Even with ftpmirror.g.o the connection is timing out... Sigh.

xnox:
> Even with ftpmirror.g.o the connection is timing out... Sigh.

URGH

murraybd (approved):

PR #2077, opened 2025-07-10
Expand check for libraries provided by the host

murraybd:
In addition to not generating runtime dependencies or provides for libcuda.so.1 we also do not want to create them for libnvcuvid.so.1 and libnvidia-encode.so.1.

### DIFF
--- a/pkg/sca/sca.go
+++ b/pkg/sca/sca.go
@@ -99,6 +99,24 @@ func isInDir(path string, dirs []string) bool {
 	return false
 }
 
+// isHostProvidedLibrary returns true if the library is provided by the host
+// system and should not be included in dependency or provides generation.
+// These are typically NVIDIA libraries that are installed by the host driver.
+func isHostProvidedLibrary(lib string) bool {
+	hostLibs := []string{
+		"libcuda.so.1",
+		"libnvcuvid.so.1",
+		"libnvidia-encode.so.1",
+	}
+	
+	for _, hostLib := range hostLibs {
+		if lib == hostLib {
+			return true
+		}
+	}
+	return false
+}
+
 // getLdSoConfDLibPaths will iterate over the files being installed by
 // the package and all its subpackages, and for each configuration
 // file found under /etc/ld.so.conf.d/ it will parse the file and add
@@ -590,8 +608,8 @@ func generateSharedObjectNameDeps(ctx context.Context, hdl SCAHandle, generated 
 		}
 
 		for _, lib := range libs {
-			// Cuda is a dangling library, which must come from the host
-			if lib == "libcuda.so.1" {
+			// These are dangling libraries, which must come from the host
+			if isHostProvidedLibrary(lib) {
 				continue
 			}
 			if strings.Contains(lib, ".so.") {
@@ -631,9 +649,9 @@ func generateSharedObjectNameDeps(ctx context.Context, hdl SCAHandle, generated 
 			}
 
 			for _, soname := range sonames {
-				// Packages should not provide libcuda.so.1 because they will
-				// conflict with the driver injected by the host.
-				if soname == "libcuda.so.1" {
+				// Packages should not provide these shared objects because they
+				// will conflict with the driver injected by the host.
+				if isHostProvidedLibrary(soname) {
 					continue
 				}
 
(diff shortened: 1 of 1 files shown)


dannf (commented):
Ah, the reason we want to filter out these libraries is different than with `libcuda.so.1`. The reason we want to filter these additional libraries has nothing to do with the whole host passthrough business. It is because there are multiple packages that provide the same soname (one in each -cuda-X variant), and every package that depends on `libnvcuvid.so.1` should really just get 1 specific one.

So if we build a binary `cuda-app-cuda-12.6` that links against `libnvcuvid.so.1`, we want it to always get `nvidia-libnvcuvid-12.6`. Not `nvidia-libnvcuvid-11.8`, and not `libnvcuvid-12.9`. In theory, we could add an additional explicit dependency on `nvidia-libnvcuvid-12.6` and the resolver would be able to figure out that the only way to resolve both the sca-added `so:libnvcuvid.so.1` *and* `nvidia-libnvcuvid-12.6` is to use `nvidia-libnvcuvid-12.6` for both. But `apko`'s resolver doesn't work that way. Instead, `apko`'s resolver will find *some* package that satisfies `so:libnvcuvid.so.1`. Once it finds one, that's it - it's resolved. Let's say it  fixated on `nvidia-libnvcuvid-11.8`. Later, it will try to resolve `nvidia-libnvcuvid-12.6`. Well, that won't work. `nvidia-libnvcuvid-12.6` conflicts with `nvidia-libnvcuvid-11.8`. `apko` will then just refuse to continue. Slumber party over, I'm calling mom to pick me up.

We should probably keep these soname lists logically separate, with separate comments explaining their purposes. Maybe a `isHostProvidedLibrary()` and a `isLibraryWithMultipleVariants()`? fwiw - I think there are just 2 libs in the `libcuda.so.1` category - the other being `libnvidia-ml.so.1`.  While there are a *ton* of libraries in the `libnvcuvid` bucket.

pkg/sca/sca.go
```diff
@@ -99,6 +99,24 @@ func isInDir(path string, dirs []string) bool {
...
+		"libnvcuvid.so.1",
+		"libnvidia-encode.so.1",
+	}
+	
+	for _, hostLib := range hostLibs {
+		if lib == hostLib {
+			return true
+		}
+	}
+	return false
```
smoser, line 141: fwiw, this could use slices.Contains:
```go
func isHostProvidedLibrary(lib string) bool {
    return slices.Contains(
        []string{"libcuda.so.1", "libnvcuvid.so.1", "libnvidia-encode.so.1"},
        lib);
}

**Comment by smoser** [comment]
@dannf thank you for that really well written comment.

File: pkg/sca/sca.go (hunk 1 above)
Comment (murraybd) [reply pkg/sca/sca.go:141]: I think the current formatting of it is a bit easier to read and also believe it'll be easier to extend. Is using slices.Contains faster?

Comment (smoser) [reply pkg/sca/sca.go:141]: I'd be surprised if it was faster.  and i can almost promise you that no human would ever know that it was faster (or slower) when processing a list of 3 items :)

Comment (smoser) [reply pkg/sca/sca.go:141]: if you like the 'hostLibs' variable for read-friendlyness, then you could also do:

```go

func isHostProvidedLibrary(lib string) {
    hostLibs := []string{
	"libcuda.so.1",
	"libnvcuvid.so.1",
	"libnvidia-encode.so.1",
    }
    return slices.Contains(hostLibs, lib)
}
```

I was mainly just informing you of slices.Contains

dannf (approved):
I love your commit message that explains the details about where this list comes from. One minor / optional suggestion would be to also note how one can repeat the process you used to generate this list, in case it needs updating in the future.

pkg/sca/sca.go
dannf: Thanks - but, FTR, it turns out to be (no longer?) correct. I attempted to reproduce that scenario and, at least today, `apko` *does* seem to properly handle additional dependencies that disambiguate a generated `so:` dep.

(I had wondered why some packages had `options->no-provides` in the past, so i tried removing them, and that broke things in a way that matched the above scenario, but I haven't gone back to retry that experiment to confirm)

PR #2078, opened 2025-07-11
feat(sbom): Include the distro qualifier as part of the APK PURL

pdeslaur:
This change is altering the PURL generated for an APK. We are keeping track of the distro (`wolfi` or `chainguard`, generally) inside the namespace identifier. However, some scanners are expect the `distro` qualifier to be set on APKs identifying our APKs.

The real question is - was there a reason not to set the `distro` qualifier previously, or is this a safe change?

### Functional Changes

- [ ] Updates the PURL of an APK to include `distro=${distro}`.

Notes:

### SCA Changes

- [ ] I've confirmed the change changes the SBOM as desired

Notes:

### Linter

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -264,6 +264,10 @@ func newAPKPackageURL(distro, name, version, arch string) *purl.PackageURL {
 		Namespace: distro,
 		Name:      name,
 		Version:   version,
+		Qualifiers: []purl.Qualifier{{
+			Key:   "distro",
+			Value: distro,
+		}},
 	}
 
 	if arch != "" {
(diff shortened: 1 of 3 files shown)


pkg/config/config.go
```diff
@@ -264,6 +264,10 @@ func newAPKPackageURL(distro, name, version, arch string) *purl.PackageURL {
 		Namespace: distro,
 		Name:      name,
 		Version:   version,
+		Qualifiers: []purl.Qualifier{{
```
javacruft: A minor point but we might want to only provide this qualifier if the distro (which is set from the --namespace option from the melange CLI) is not set to "unknown", ensuring we always provide something of value.

javacruft (dismissed):

pkg/config/config.go
pdeslaur: Agreed - that makes a lot of sense. Updated.

PR #2079, opened 2025-07-14
Use os.DirFS over apkofs.DirFS

jonjohnsonjr:
We don't need this anymore, and the signature changed to require a context.Context, so I'm just dropping it rather than dealing with that.

### DIFF
--- a/pkg/build/test.go
+++ b/pkg/build/test.go
@@ -208,11 +208,22 @@ func (t *Test) PopulateWorkspace(ctx context.Context, src fs.FS) error {
 
 	log.Infof("populating workspace %s from %s", t.WorkspaceDir, t.SourceDir)
 
-	fsys := apkofs.DirFS(t.SourceDir, apkofs.WithCreateDir())
-
-	if fsys == nil {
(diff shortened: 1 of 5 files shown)


pkg/build/test.go
```diff
@@ -208,12 +208,23 @@ func (t *Test) PopulateWorkspace(ctx context.Context, src fs.FS) error {
...
+		log.Warn("error checking dir", "error", err)
+		return nil
+	case err != nil && os.IsNotExist(err):
+		if err := os.MkdirAll(t.SourceDir, 0o700); err != nil {
+			log.Warn("error creating dir", "dir", t.SourceDir, "error", err)
+			return nil
+		}
+	case !fi.IsDir():
+		log.Warn("not a directory", "dir", t.SourceDir)
+		return nil
```
markusthoemmes: Shouldn't these all return the respective error to remain consistent with what we had before?

jonjohnsonjr: Thanks, I missed that we just log those errors and then turn a nil response into its own error 🤷 

Let me figure out if we need the other case sensitive stuff too 🤔

PR #2099, opened 2025-07-22
Do not require .so files to be executable

murraybd:
There is no reason to require .so files to be executable and other distributions like Ubuntu do not make their .so files executable. So let's drop that requirement.

I did a melange scan of the archive with this change (details in the linked issue) and the additional provides from this change are minimal and should not cause an issue.

### DIFF
--- a/pkg/sca/sca.go
+++ b/pkg/sca/sca.go
@@ -544,10 +544,6 @@ func generateSharedObjectNameDeps(ctx context.Context, hdl SCAHandle, generated 
 			return nil
 		}
 
-		if mode.Perm()&0o555 != 0o555 {
-			return nil
-		}
-
 		basename := filepath.Base(path)
 
 		// most likely a shell script instead of an ELF, so treat any


xnox (changes requested):
we have flipped flopped this multiple times.

after doing this change, it would be nice to redo `melange scan --diff` across the whole archive and see if things generated are the expected ones; and do not conflict.

In the past based on this setting, we have many libraries that previously were not registered in provides; but now will be; and also become conflicting.

cc @justinvreeland @smoser 

So the question is, which archive wide validations were done to ensure this will not cause regressions upon package rebuilds?

murraybd:
> Setting this to draft, because I believe:
> 
> * [Expand check for libraries provided by the host #2077](https://github.com/chainguard-dev/melange/pull/2077)
> 
> Is a pre-requisite which must be merged first or at the same time as this change

The pre-req PR has been merged now.

PR #2102, opened 2025-07-26
cleanup: move packages to internal/, unexport methods

imjasonh:
Moving things to `internal/` makes the linter mad if they're unused in the module, which means we can also delete those. 🧹 

When packages are in `internal/` it's easier to tell whether API changes will break other users, since all those users must be in the same module.

Unexporting methods has the same effect; it's easier to tell whether/how things are used, and update all callers when you change APIs.


sil2100 (approved):
Went through the changes and they all look correct. No files have been lost in the move. Even the manifest.go removal looks all good with the required functionality preserved. Did a test build, ran some tests:

- Package builds work
- Package testing works
- Renovate commands for bump and copyright work

I think this is good to go. @imjasonh mentioned checking if this doesn't break any outside users too, so all good here. I'd recommend merging this when things are calm, but I see no showstoppers from the actual PR POV.

PR #2105, opened 2025-07-28
linter.go: Added manInfo linter

AbduAzim424:
A separate linter that is more specific than the current `documentationLinter` whose RegEx also includes `README`, `TODO`, `CREDITS` etc. This linter uses RegEx's that are similar to the `split/doc` and `split/info pipelines`.

Example:
```
./melange lint vim-9.1.1591-r0.apk

2025/07/28 16:40:05 INFO Required checks: [dev infodir tempdir usrmerge varempty]
2025/07/28 16:40:05 INFO Warning checks: [cudaruntimelib lddcheck maninfo object opt pkgconf python/docs python/multiple python/test setuidgid srv strip usrlocal worldwrite]
2025/07/28 16:40:05 INFO linting apk: vim (size: 11 MB)
2025/07/28 16:40:05 WARN linter "maninfo" failed on package "vim": package contains man/info files but is not a documentation package: usr/share/man/man1/evim.1; suggest: Place documentation into a separate package or remove it
```

### DIFF
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -114,6 +114,11 @@ var linterMap = map[string]linter{
 		Explain:         "This package contains intermediate object files",
 		defaultBehavior: Warn,
 	},
+	"maninfo": {
+		LinterFunc:      allPaths(manInfoLinter),
+		Explain:         "Place documentation into a separate package or remove it",
+		defaultBehavior: Warn,
+	},
 	"sbom": {
 		LinterFunc:      allPaths(sbomLinter),
 		Explain:         "Remove any files in /var/lib/db/sbom from the package",
@@ -234,6 +239,19 @@ func documentationLinter(_ context.Context, _ *config.Configuration, pkgname, pa
 	if isDocumentationFileRegex.MatchString(path) && !strings.HasSuffix(pkgname, "-doc") {
 		return errors.New("package contains documentation files but is not a documentation package")
 	}
+	return nil
+}
+
+var (
+    manRegex  = regexp.MustCompile(`(?i)^usr/(?:(?:local/)?share/man|man)/man[0-9][^/]*/[^/]+\.[0-9][^/]*(?:\.(?:gz|bz2|xz|lzma|Z))?$`)
+    infoRegex = regexp.MustCompile(`(?i)^usr/share/info/(?:dir|[^/]+\.info(?:\-[0-9]+)?(?:\.(?:gz|bz2|xz|lzma|Z))?)$`)
+)
+
+func manInfoLinter(_ context.Context, _ *config.Configuration, pkgname, path string) error {
+	if (manRegex.MatchString(path) || infoRegex.MatchString(path)) && !strings.HasSuffix(pkgname, "-doc") {
+		return errors.New("package contains man/info files but is not a documentation package")
+	}
+
 	return nil
 }
 
--- a/pkg/linter/linter_test.go
+++ b/pkg/linter/linter_test.go
@@ -19,6 +19,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"chainguard.dev/melange/pkg/config"
@@ -292,12 +293,67 @@ func TestLinters(t *testing.T) {
 		dirFunc: mkfile(t, "usr/lib/libcuda.so.560.35.05"),
 		linter:  "cudaruntimelib",
 		pass:    false,
+	}, {
+		dirFunc: mkfile(t, "usr/share/man/man1/foo.1"),
+		linter:  "maninfo-regular",
+		pass:    false,
+	}, {
+		dirFunc: mkfile(t, "usr/share/man/man1/foo.1"),
+		linter:  "maninfo-doc",
+		pass:    true,
+	}, {
+		dirFunc: mkfile(t, "usr/share/man/man8/bar.8.gz"),
+		linter:  "maninfo-regular",
+		pass:    false,
+	}, {
+		dirFunc: mkfile(t, "usr/local/share/man/man3/baz.3"),
+		linter:  "maninfo-regular",
+		pass:    false,
+	}, {
+		dirFunc: mkfile(t, "usr/man/man5/qux.5"),
+		linter:  "maninfo-regular",
+		pass:    false,
+	}, {
+		dirFunc: mkfile(t, "usr/share/info/test.info"),
+		linter:  "maninfo-regular",
+		pass:    false,
+	}, {
+		dirFunc: mkfile(t, "usr/share/info/test.info.gz"),
+		linter:  "maninfo-regular",
+		pass:    false,
+	}, {
+		dirFunc: mkfile(t, "usr/share/info/test.info-1"),
+		linter:  "maninfo-regular",
+		pass:    false,
+	}, {
+		dirFunc: mkfile(t, "usr/share/info/dir"),
+		linter:  "maninfo-regular",
+		pass:    false,
+	}, {
+		dirFunc: mkfile(t, "usr/share/info/test.info"),
+		linter:  "maninfo-doc",
+		pass:    true,
+	}, {
+		dirFunc: mkfile(t, "usr/bin/normal"),
+		linter:  "maninfo-regular",
+		pass:    true,
+	}, {
+		dirFunc: mkfile(t, "usr/lib/libfoo.so.1"),
+		linter:  "maninfo-regular",
+		pass:    true,
 	}} {
 		ctx := slogtest.Context(t)
 		t.Run(c.linter, func(t *testing.T) {
 			dir := c.dirFunc()
+
+			// Determine actual linter name
+			actualLinter := c.linter
+			if strings.HasPrefix(c.linter, "maninfo-") {
+				actualLinter = "maninfo"
+			}
+
 			// In required mode, it should raise an error.
-			err := LintBuild(ctx, c.cfg, c.linter, dir, []string{c.linter}, nil)
+			err := LintBuild(ctx, c.cfg, c.linter, dir, []string{actualLinter}, nil)
 			if c.pass {
 				assert.NoError(t, err)
 			} else {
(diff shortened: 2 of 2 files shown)


pkg/linter/linter.go
```diff
@@ -237,6 +242,19 @@ func documentationLinter(_ context.Context, _ *config.Configuration, pkgname, pa
 	return nil
 }
 
+var (
+	manRegex  = regexp.MustCompile(`(?i)^usr/(?:(?:local/)?share/man|man)/man[0-9][^/]*/[^/]+\.[0-9][^/]*(?:\.(?:gz|bz2|xz|lzma|Z))?$`)
+	infoRegex = regexp.MustCompile(`(?i)^usr/share/info/(?:dir|[^/]+\.info(?:\-[0-9]+)?(?:\.(?:gz|bz2|xz|lzma|Z))?)$`)
```
dannf: I suggest also including `usr/local` here for info pages here

pkg/linter/linter.go
```diff
@@ -237,6 +242,19 @@ func documentationLinter(_ context.Context, _ *config.Configuration, pkgname, pa
...
+)
+
+func manInfoLinter(_ context.Context, _ *config.Configuration, pkgname, path string) error {
+	if (manRegex.MatchString(path) || infoRegex.MatchString(path)) && !strings.HasSuffix(pkgname, "-doc") {
```
dannf: I suggest doing the `HasSuffix` first, and returning w/o error. Only then do we need to do the regex checks, which will be more expensive.

pkg/linter/linter.go
```diff
@@ -237,6 +242,19 @@ func documentationLinter(_ context.Context, _ *config.Configuration, pkgname, pa
...
+		return errors.New("package contains man/info files but is not a documentation package")
```
dannf: You might as well emit the path since you have it.

pkg/linter/linter_test.go
```diff
@@ -292,20 +293,75 @@ func TestLinters(t *testing.T) {
...
+		pass:    false,
+	}, {
+		dirFunc: mkfile(t, "usr/share/info/test.info-1"),
+		linter:  "maninfo-regular",
+		pass:    false,
+	}, {
+		dirFunc: mkfile(t, "usr/share/info/dir"),
+		linter:  "maninfo-regular",
+		pass:    false,
+	}, {
```
dannf, line 341: I suggest an usr/local/* test as well

pkg/linter/linter_test.go
```diff
@@ -292,20 +293,75 @@ func TestLinters(t *testing.T) {
...
 	}} {
 		ctx := slogtest.Context(t)
 		t.Run(c.linter, func(t *testing.T) {
 			dir := c.dirFunc()
+
+			// Determine actual linter name
+			actualLinter := c.linter
+			if strings.HasPrefix(c.linter, "maninfo-") {
+				actualLinter = "maninfo"
+			}
```
dannf, line 369: It is clever to use the linter name as a way to get around not having a package name parameter, but I don't think we need be clever here. I suggest modifying the code to support a package name parameter.

pkg/linter/linter_test.go
```diff
@@ -115,6 +115,7 @@ func TestLinters(t *testing.T) {
 		dirFunc func() string
 		linter  string // NB: Also the package name for this test, but that's weird!
 		pass    bool
+		pkg     string // package name (added with "maninfo" lint step)
```
dannf: If we named this variable `pkgname`, then I think we wouldn't need the comment.
(There's no need to say when it was added - that's what git history is for)

pkg/linter/linter.go
```diff
@@ -237,6 +242,33 @@ func documentationLinter(_ context.Context, _ *config.Configuration, pkgname, pa
...
+)
+
+func manInfoLinter(_ context.Context, _ *config.Configuration, pkgname, path string) error {
+	if strings.HasSuffix(pkgname, "-doc") {
+		return nil
+	}
+
+	if manRegex.MatchString(path) || infoRegex.MatchString(path) {
+		parts := strings.SplitN(path, "/", 4)
+		if len(parts) >= 3 {
```
dannf: similar to the above, you can say:
```go
if len (parts) < 3 {
  return nil
}
```
and avoid the indention

pkg/linter/linter.go
```diff
@@ -237,6 +242,33 @@ func documentationLinter(_ context.Context, _ *config.Configuration, pkgname, pa
...
+			var prefix string
+			if strings.Contains(parts[2], ".") {
+				prefix = "/" + strings.Join(parts[:2], "/")
+			} else {
+				prefix = "/" + strings.Join(parts[:3], "/")
+			}
+			return fmt.Errorf("package contains man/info files in %q but is not a documentation package", prefix)
```
dannf: I don't really follow the above (why does it matter if it contains a "."?), but it seems overcomplicated for the task IMO. Just printing an error with the path should be fine.

pkg/linter/linter_test.go
```diff
@@ -115,6 +115,7 @@ func TestLinters(t *testing.T) {
 		dirFunc func() string
 		linter  string // NB: Also the package name for this test, but that's weird!
```
dannf: Maybe change the comment to `"Also used as the fallback package name for the test if unspecified"`

pkg/linter/linter_test.go
```diff
@@ -292,20 +293,91 @@ func TestLinters(t *testing.T) {
...
+		linter:  "maninfo",
+		pkg:     "regular-package",
+		pass:    true,
 	}} {
 		ctx := slogtest.Context(t)
 		t.Run(c.linter, func(t *testing.T) {
 			dir := c.dirFunc()
+
+			packageName := c.pkg
+			if packageName == "" {
```
dannf: i'd suggest using `pkgname` here. I find it annoying when the same variable has subtly different spellings in different parts of a program.

pkg/linter/linter_test.go
```diff
@@ -292,20 +293,91 @@ func TestLinters(t *testing.T) {
...
+				packageName = c.linter // Fallback for existing tests
```
dannf: "for existing tests" helps a reviewer read this change, but it won't make a lot of sense to the majority of people in the future who read this and don't know what order things happened in. I'd suggest just saying something like "fallback to using the linter name as the fake package name"

pkg/linter/linter_test.go
```diff
@@ -292,20 +293,91 @@ func TestLinters(t *testing.T) {
...
+		pass:    false,
+	}, {
+		dirFunc: mkfile(t, "usr/local/share/man/man3/baz.3"),
+		linter:  "maninfo",
+		pkg:     "regular-package",
+		pass:    false,
+	}, {
+		dirFunc: mkfile(t, "usr/man/man5/qux.5"),
+		linter:  "maninfo",
+		pkg:     "regular-package",
```
dannf: can you call this pkgname? see above

dannf (changes requested):

pkg/linter/linter.go
AbduAzim424: Yes so this was my attempt to make `warning` message a little cleaner. For example this is how it looks with this `prefix` staff:
```
2025/07/29 14:36:07 WARN linter "maninfo" failed on package "zsh": package contains man/info files in "/usr/man" but is not a documentation package: usr/man/man.1; suggest: Place documentation into a separate package or remove it
```
As you can see even if I do not emit the path, the file name and path is printed before the `suggest` part. What I'm printing is the prefix portion as a general view of where do man pages reside, in form of `/usr/man` instead of printing `usr/man/man.1` twice.

AbduAzim424: This is how it looks without emitting `path` manually:
```
2025/07/29 14:36:07 WARN linter "maninfo" failed on package "zsh": package contains man/info files but is not a documentation package: usr/man/man.1; suggest: Place documentation into a separate package or remove it
```

AbduAzim424: I can emit path simply by just using `path` itself without any modifications and it will be cleaner but will result in duplication.

AbduAzim424: Ok I tried to build altered `vim` with man pages without my `path` changes and it looks like this after build finished:
```
2025/07/30 09:41:33 WARN linter "maninfo" failed on package "vim": package contains man/info files but is not a documentation package: usr/share/man/man1/evim.1; suggest: Place documentation into a separate package or remove it
```
So I assume I just drop emitting `path` as it gets printed by `melange`

pkg/linter/linter.go
```diff
@@ -237,6 +242,23 @@ func documentationLinter(_ context.Context, _ *config.Configuration, pkgname, pa
 	return nil
 }
 
+var (
+	manRegex  = regexp.MustCompile(`(?i)^usr/(?:local/)?share/man(?:/man[0-9][^/]*)?(?:/[^/]+\.[0-9][^/]*(?:\.(?:gz|bz2|xz|lzma|Z))?)?$|^usr/man(?:/man[0-9][^/]*)?(?:/[^/]+\.[0-9][^/]*(?:\.(?:gz|bz2|xz|lzma|Z))?)?$`)
+	infoRegex = regexp.MustCompile(`(?i)^usr/(local/)?share/info/(?:dir|[^/]+\.info(?:\-[0-9]+)?(?:\.(?:gz|bz2|xz|lzma|Z))?)$`)
```
dannf: why the difference in `(?:local/)?` and `(local/)?
and why `(?i)`?

AbduAzim424: The `(?:local/)?` is on me, I fixed and both now use `(?:local/)?`. The `(?i)` makes the regex case-insensitive. This might not be critical but I decided to add it. Should I remove it?

AbduAzim424: From what I understand the difference between `(?:local/)?` and `(local/)?` is that `(?:local/)?` does not capture the matched expression while `(local/)?` does. The `(local/)?` is good if we want to do something with that captured path, but in our case we just `warn` and exit

dannf: Thanks - that makes sense, thanks for making them consistent. And yeah, let's drop the `(?i)` unless there's a known reason for it. `/USR` would be a different directory than `/usr`. Simpler is better, and the next person won't have to look up what it means :)

But with that changed, I think this is ready :)

dannf (approved):
Thanks!

PR #2106, opened 2025-07-29
source: copy package contents when fetching source

sil2100:
Right now in most cases `patch` calls do not work as we don't have the patches present.

### DIFF
--- a/pkg/source/source_test.go
+++ b/pkg/source/source_test.go
@@ -124,10 +124,11 @@ func TestFetchSourceFromMelange(t *testing.T) {
 		fileName      string
 		expectedSteps []string
 		expectedName  string
+		expectedFiles []string
 	}{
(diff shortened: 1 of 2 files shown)


pkg/source/source_test.go
```diff
@@ -161,6 +162,16 @@ func TestFetchSourceFromMelange(t *testing.T) {
 					t.Errorf("Expected step %s, got %s", tc.expectedSteps[i], step)
 				}
 			}
+
+			// Validate the files in the destination directory
+			if tc.expectedFiles != nil {
```
dannf: you might consider:
```go
if tc.expectedFiles == nil {
  continue
}
```
to get rid of a level of indention (we're pretty nested here already)

dannf (approved):

dannf (approved):
LGTM - very good use of self-documenting names :)

PR #2116, opened 2025-07-31
Revert "SCA: Generate depends for shlibs ending in .so" and related commits

sil2100:
Temporary revert to the previous known working state regarding shlib dependency generation.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -92,10 +92,6 @@ type PackageOption struct {
 	NoCommands bool `json:"no-commands,omitempty" yaml:"no-commands,omitempty"`
 	// Optional: Don't generate versioned depends for shared libraries
 	NoVersionedShlibDeps bool `json:"no-versioned-shlib-deps,omitempty" yaml:"no-versioned-shlib-deps,omitempty"`
-	// Optional: Don't generate inter-package depends when one
-	// subpackage depends on a vendored shared library shipped by
-	// another.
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
```diff
@@ -92,10 +92,6 @@ type PackageOption struct {
 	NoCommands bool `json:"no-commands,omitempty" yaml:"no-commands,omitempty"`
 	// Optional: Don't generate versioned depends for shared libraries
 	NoVersionedShlibDeps bool `json:"no-versioned-shlib-deps,omitempty" yaml:"no-versioned-shlib-deps,omitempty"`
-	// Optional: Don't generate inter-package depends when one
-	// subpackage depends on a vendored shared library shipped by
-	// another.
-	NoVendoredCrossPackageDeps bool `json:"no-vendored-cross-package-deps,omitempty" yaml:"no-vendored-cross-package-deps,omitempty"`
```
xnox: I wonder if this alone can be or should be kept in place. (unused)

sil2100: I proposed that originally as per the slack convo, but we weren't sure if we want that or not - since a simple revert is easy to re-revert. But yes, I was thinking about this.

And now with more people saying +1, I just did that!

javacruft (approved):
I was going to suggest we leave the BUILD-FILE docs in place but I think just leaving the schema change intact is sufficient for validation checks and moving this forwards later.  Avoiding anyone using the feature and wondering why it does nothing.

xnox (approved):

PR #2118, opened 2025-08-06
Fix and improve file permission linters; update tests to mirror actual usage

egibs:
This PR fixes and improves some of the linting functions, especially `isSetUIDOrGIDLinter` and `worldWriteableLinter`. 

For starters, making both linters required will provide more immediate feedback when working on packages locally or in pre-submit. Since removing linters is not much of a lift, we can easily exclude packages that require these features.

There was also a more serious issue with the linters wherein we create a new DirFS filesystem which modifies permissions in a way that will prevent something like `worldWriteableLinter` from actually finding anything. To fix this, we need to pass the same workspace filesystem to the linters so that each runs against the actual contents that will be stored in the APK (linting an existing APK is unaffected).

Similarly, the linter tests were not following the pattern we use for builds, so I fixed that as well. We were also excluding certain paths/files from these linters.

Finally, I updated the SBOM linter to take the path and file extension into account so we can at least run that linter in warning mode.


pkg/build/build.go
```diff
@@ -859,7 +860,12 @@ func (b *Build) BuildPackage(ctx context.Context) error {
 	// perform package linting
 	for _, lt := range linterQueue {
 		log.Infof("running package linters for %s", lt.pkgName)
-		path := filepath.Join(b.WorkspaceDir, melangeOutputDirName, lt.pkgName)
+
+		// lint the in-memory filesystem which contains the correct mode bits, xattrs, etc.
+		fsys, err := apkofs.Sub(b.WorkspaceDirFS, filepath.Join(melangeOutputDirName, lt.pkgName))
```
egibs, line 865: Verifying that `Sub` does not alter permissions:
```yml
...
- runs: chmod -R 0777 "${{targets.destdir}}"
...
```
```go
err := fs.WalkDir(fsys, ".", func(path string, d fs.DirEntry, err error) error {
	if err != nil {
		return err
	}

	fi, err := d.Info()
	if err != nil {
		return err
	}
	mode := fi.Mode()
	perm := strconv.FormatUint(uint64(mode.Perm())&07777, 8)

	fmt.Println(path, perm)

	return nil
})
```
```
. 755
usr 777
usr/bin 777
usr/bin/fping 777
```

pkg/linter/linter_test.go
```diff
@@ -362,112 +363,116 @@ func TestLinters(t *testing.T) {
...
 	ctx := slogtest.Context(t)
 
 	dir := t.TempDir()
+	fsys := apkofs.DirFS(ctx, dir)
 
 	linters := []string{"python/test"}
 
 	// Base dir
-	pythonPathdir := filepath.Join(dir, "usr", "lib", "python3.14", "site-packages")
+	pythonPathdir := filepath.Join("usr", "lib", "python3.13", "site-packages")
```
justinvreeland, line 448: Why did we change python versions here?

pkg/linter/linter_test.go
```diff
@@ -362,112 +363,116 @@ func TestLinters(t *testing.T) {
...
 func Test_pythonMultiplePackagesLinter(t *testing.T) {
 	ctx := slogtest.Context(t)
 	dir := t.TempDir()
+	fsys := apkofs.DirFS(ctx, dir)
 
 	linters := []string{"python/multiple"}
 
 	// Base dir
-	pythonPathdir := filepath.Join(dir, "usr", "lib", "python3.14", "site-packages")
+	pythonPathdir := filepath.Join("usr", "lib", "python3.13", "site-packages")
```
justinvreeland: Here too but the other python examples stayed the same.

pkg/linter/linter_test.go
egibs: I think it's arbitrary but I can leave it as-is since it's not critical that the version be a specific value.

pkg/linter/linter_test.go
egibs: Reverted in [`a25de39` (#2118)](https://github.com/chainguard-dev/melange/pull/2118/commits/a25de39e991e6c1baa8a7b01f8b1502207634cec).

pkg/linter/linter_test.go
egibs: Reverted in [`a25de39` (#2118)](https://github.com/chainguard-dev/melange/pull/2118/commits/a25de39e991e6c1baa8a7b01f8b1502207634cec).

Issue #2133, opened 2025-08-19
Melange test-e2e doesn't work by default and it's hard to figure out how to run it


justinvreeland:
When trying to test fixes for https://github.com/chainguard-dev/melange/issues/2131 I noticed I can't run the e2e tests locally. I can't see to get things running with docker correctly and it looks like the e2e runner has been set to qemu without adding code to download and set the kernel image. There's a number of failures but in the end it looks like this:

```
2025/08/19 16:40:26 INFO qemu: generating ssh key pairs for ephemeral VM
2025/08/19 16:40:26 ERRO could not prepare rootfs: qemu: /boot/vmlinuz not found, specify a kernel path with env variable QEMU_KERNEL_IMAGE
2025/08/19 16:40:26 ERRO ERROR: failed to test package. the test environment has been preserved:
2025/08/19 16:40:26 ERRO failed to test package: unable to start pod: qemu: /boot/vmlinuz not found, specify a kernel path with env variable QEMU_KERNEL_IMAGE
ERROR[1]: Testing xcover-nopkg-test from xcover-nopkg-test.yaml for test
FAIL:  capabilities-add-drop-build.yaml/build capabilities-add-drop-nopkg-test.yaml/test env-vars-build-test.yaml/build env-vars-build-test.yaml/test git-checkout-build.yaml/build greeter-build-test.yaml/build greeter-build-test.yaml/test numpy-test.yaml/test php-8.2-msgpack-nopkg-test.yaml/test php-8.2-msgpack-test.yaml/test py3-pandas-nopkg-test.yaml/test py3-pandas-test.yaml/test py3-setuptools-test.yaml/test xcover-nopkg-test.yaml/test
make: *** [Makefile:162: test-e2e] Error 1
```

I think this is caused by: https://github.com/chainguard-dev/melange/commit/ece051e1b85548685b4a8881c2bce7dd68e84c99

PR #2134, opened 2025-08-20
e2e-test: Fix local test running by using alpine kernel

justinvreeland:
This adds a `fetch-kernel` target to the Makefile like wolfi used to have. This is roughly the same as what CI is doing and putting in `/tmp/` but keeping the kernel in the working dir instead.  Since we're downloading the kernel anyway we might as well run the tests on the native archs which makes tests a bit quicker on my mac.  I've tested this on an aarch64 mac with aarch64 and x86_64 `ARCH` settings and on x86_64 linux with both `ARCH`s as well. I have another commit that runs the e2e-tests with every backend but there are a small number of failures. We also set melange to the top level melange here.  The test-e2e target now explicitly sets melange to the project binary. It seems strange to me that anyone would want to silently use the melange in their path melange but that is how it's set up.

The qemu runner detection in one of our tests isn't reliable and won't work for qemu runner on macOS.  I added a kernel cmdline parameter to check if we're in the qemu runner. It's global now but I can understand wanting to pull it into a flag or environment option if that's preferred.

Also included in here is a shortened timeout for development but I do think an hour is way to long to wait for a VM to start up and I've been bitten by a bad setting and not knowing for an hour.

fixes: https://github.com/chainguard-dev/melange/issues/2133

### DIFF
--- a/e2e-tests/capabilities-add-drop-build.yaml
+++ b/e2e-tests/capabilities-add-drop-build.yaml
@@ -21,17 +21,17 @@ pipeline:
   - name: Test default effective capability
     runs: |
       # Skip test if on QEMU runner, since the runner does not support process capabilities add/drop Melange feature.
-      { test -f /sys/class/dmi/id/sys_vendor && grep -E "^QEMU$" /sys/class/dmi/id/sys_vendor && exit 0; } || true
+      grep "melange_qemu_runner=1" /proc/cmdline && { echo "SKIP - running inside melange runner"; exit 0; }
       capsh --decode=$(grep CapEff /proc/self/status | cut -d ':' -f2 | xargs) | grep -i cap_dac_override
 
   - name: Test added non-default effective capability
     runs: |
       # Skip test if on QEMU runner, since the runner does not support process capabilities add/drop Melange feature.
-      { test -f /sys/class/dmi/id/sys_vendor && grep -E "^QEMU$" /sys/class/dmi/id/sys_vendor && exit 0; } || true
+      grep "melange_qemu_runner=1" /proc/cmdline && { echo "SKIP - running inside melange runner"; exit 0; }
       capsh --decode=$(grep CapEff /proc/self/status | cut -d ':' -f2 | xargs) | grep -i cap_net_admin
 
   - name: Test dropped default effective capability
     runs: |
       # Skip test if on QEMU runner, since the runner does not support process capabilities add/drop Melange feature.
-      { test -f /sys/class/dmi/id/sys_vendor && grep -E "^QEMU$" /sys/class/dmi/id/sys_vendor && exit 0; } || true
+      grep "melange_qemu_runner=1" /proc/cmdline && { echo "SKIP - running inside melange runner"; exit 0; }
       capsh --decode=$(grep CapEff /proc/self/status | cut -d ':' -f2 | xargs) | grep -vi cap_sys_chroot
--- a/e2e-tests/capabilities-add-drop-nopkg-test.yaml
+++ b/e2e-tests/capabilities-add-drop-nopkg-test.yaml
@@ -23,17 +23,17 @@ test:
     - name: Test default effective capability
       runs: |
         # Skip test if on QEMU runner, since the runner does not support process capabilities add/drop Melange feature.
-        { test -f /sys/class/dmi/id/sys_vendor && grep -E "^QEMU$" /sys/class/dmi/id/sys_vendor && exit 0; } || true
+        grep "melange_qemu_runner=1" /proc/cmdline && { echo "SKIP - running inside melange runner"; exit 0; }
         capsh --decode=$(grep CapEff /proc/self/status | cut -d ':' -f2 | xargs) | grep -i cap_dac_override
 
     - name: Test added non-default effective capability
       runs: |
         # Skip test if on QEMU runner, since the runner does not support process capabilities add/drop Melange feature.
-        { test -f /sys/class/dmi/id/sys_vendor && grep -E "^QEMU$" /sys/class/dmi/id/sys_vendor && exit 0; } || true
+        grep "melange_qemu_runner=1" /proc/cmdline && { echo "SKIP - running inside melange runner"; exit 0; }
         capsh --decode=$(grep CapEff /proc/self/status | cut -d ':' -f2 | xargs) | grep -i cap_net_admin
 
     - name: Test dropped default effective capability
       runs: |
         # Skip test if on QEMU runner, since the runner does not support process capabilities add/drop Melange feature.
-        { test -f /sys/class/dmi/id/sys_vendor && grep -E "^QEMU$" /sys/class/dmi/id/sys_vendor && exit 0; } || true
+        grep "melange_qemu_runner=1" /proc/cmdline && { echo "SKIP - running inside melange runner"; exit 0; }
         capsh --decode=$(grep CapEff /proc/self/status | cut -d ':' -f2 | xargs) | grep -vi cap_sys_chroot
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -675,7 +675,7 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 	// panic=-1 ensures that if the init fails, we immediately exit the machine
 	// Add default SSH keys to the VM
 	sshkey := base64.StdEncoding.EncodeToString(pubKey)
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -156,10 +156,27 @@ integration:
...
+	mkdir -p kernel/${ARCH}
+	tar -xf ./linux-virt.apk -C kernel/${ARCH}
+
 .PHONY: test-e2e
-test-e2e: generate # This is invoked by a separate GHA workflow, so not combining it with the other test targets.
-	go test -tags e2e ./... -race
-	cd e2e-tests && ./run-tests
+test-e2e: fetch-kernel generate # This is invoked by a separate GHA workflow, so not combining it with the other test targets.
+	$(eval KERNEL_VER := $(shell ls kernel/lib/modules))
+	cd e2e-tests && \
```
justinvreeland, line 211: Looks like I dropped the go test line

e2e-tests/capabilities-add-drop-build.yaml
```diff
@@ -21,17 +21,17 @@ pipeline:
...
       # Skip test if on QEMU runner, since the runner does not support process capabilities add/drop Melange feature.
-      { test -f /sys/class/dmi/id/sys_vendor && grep -E "^QEMU$" /sys/class/dmi/id/sys_vendor && exit 0; } || true
+      {  grep -E "melange_qemu_runner=1" /proc/cmdline && exit 0; } || true
       capsh --decode=$(grep CapEff /proc/self/status | cut -d ':' -f2 | xargs) | grep -i cap_net_admin
 
   - name: Test dropped default effective capability
     runs: |
       # Skip test if on QEMU runner, since the runner does not support process capabilities add/drop Melange feature.
-      { test -f /sys/class/dmi/id/sys_vendor && grep -E "^QEMU$" /sys/class/dmi/id/sys_vendor && exit 0; } || true
+      {  grep -E "melange_qemu_runner=1" /proc/cmdline && exit 0; } || true
```
smoser: `melange_qemu_runner=1` isn't really a extended regex (`-E`). I'm guessing you were trying something like `--word-regexp` (full word match).

e2e-tests/capabilities-add-drop-build.yaml
```diff
@@ -21,17 +21,17 @@ pipeline:
   - name: Test default effective capability
     runs: |
       # Skip test if on QEMU runner, since the runner does not support process capabilities add/drop Melange feature.
-      { test -f /sys/class/dmi/id/sys_vendor && grep -E "^QEMU$" /sys/class/dmi/id/sys_vendor && exit 0; } || true
+      {  grep -E "melange_qemu_runner=1" /proc/cmdline && exit 0; } || true
```
smoser: this would be much more clearly written as:

```
grep "melange_qemu_runner=1" /proc/cmdline && { echo "SKIP - running inside melange runner"; exit 0; }
```

justinvreeland:
> My feelling here is we should not bother using an alpine kernel that we do not control or intend to "fix" if it breaks. Rather we should use the artifact that we use in wolfi.

I'm not particularly concerned about the alpine kernel breaking. We're not using any strange features and the virt kernel is pretty straightforward. Maybe we should pin it instead of pulling the newest all the time. I'm not a huge fan of the `eval.....ls` stuff so that fixes that too.  I would in general like to have a kernel we control more though if only because I'd rather our test kernel be slightly smaller and mostly builtins.

I am not in favor of the auth strategy by default, people should be able to run `make test-e2e` without being auth'd imo. Our infra should be more than happy to test with our asset though.

smoser:
> I am not in favor of the auth strategy by default, people should be able to run make test-e2e without being auth'd imo. 

I agree with that, but we're not doing that in wolfi.

smoser:
```diff
commit 4a103d69fc3247eb596fc647de8dd6644d024035
Author: Scott Moser <smoser@brickies.net>
Date:   Wed Aug 20 15:58:36 2025 -0400

    more like wolfi

diff --git a/Makefile b/Makefile
index 5e0ab3df..6ab74c07 100644
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 # Some nice defines for the "make install" target
+TOP_D := $(patsubst %/,%,$(dir $(abspath $(lastword $(MAKEFILE_LIST)))))
 
 OS := $(shell uname -s)
 ifeq ($(OS), Darwin)
@@ -162,21 +163,55 @@ ifeq (${ARCH}, arm64)
 endif
 
 
+curlget = mkdir -p $$(dirname $(2)) && \
+  echo "curl[$(1)] from $(2)" && \
+  mkdir -p $(dir $(1)) && \
+  curl --fail -LS --silent -o $(1).tmp.$$$$ $(2) && \
+	mv "$(1).tmp.$$$$" "$(1)"
+
 .PHONY: fetch-kernel
 fetch-kernel:
-	$(eval KERNEL_PKG := $(shell curl -sL https://dl-cdn.alpinelinux.org/alpine/edge/main/$(ARCH)/APKINDEX.tar.gz | tar -Oxz APKINDEX | awk -F':' '$$1 == "P" {printf "%s-", $$2} $$1 == "V" {printf "%s.apk\n", $$2}' | grep "linux-virt" | grep -v dev))
-	curl -s -LSo linux-virt.apk "https://dl-cdn.alpinelinux.org/alpine/edge/main/$(ARCH)/$(KERNEL_PKG)"
-	mkdir -p kernel/${ARCH}
-	tar -xf ./linux-virt.apk -C kernel/${ARCH}
+	rm -rf kernel/$(ARCH)
+	$(MAKE) kernel/$(ARCH)/vmlinuz
+
+QEMU_KERNEL_REPO = https://dl-cdn.alpinelinux.org/alpine/edge/main/
+kernel/%/APKINDEX.tar.gz:
+	@$(call curlget,$@,$(QEMU_KERNEL_REPO)/$(ARCH)/APKINDEX.tar.gz)
+
+kernel/%/APKINDEX: kernel/%/APKINDEX.tar.gz
+	tar -x -C kernel --to-stdout -f $< APKINDEX > $@.tmp.$$$$ && mv $@.tmp.$$$$ $@
+	touch $@
+
+kernel/%/chosen: kernel/%/APKINDEX
+	# Extract lines with 'P:linux-qemu-generic' and the following line that contains the version
+	# This approach is compatible with both GNU and BSD sed
+	awk '/^P:linux-virt$$/ {print; getline; print}' $< > kernel/$*/available
+	grep '^V:' kernel/$*/available | sed 's/V://' | \
+	  sort -V | tail -n1 > $@.tmp
+	# Sanity check that this looks like an apk version
+	grep -E '^([0-9]+\.)+[0-9]+-r[0-9]+$$' $@.tmp
+	mv $@.tmp $@
+
+kernel/%/linux.apk: kernel/%/chosen
+	@$(call curlget,$@,$(QEMU_KERNEL_REPO)/$*/linux-virt-$(shell cat kernel/$*/chosen).apk)
+
+# kernel/<arch>/vmlinuz makes kernel/<arch>/modules as a side effect
+kernel/%/vmlinuz: kernel/%/linux.apk
+	tmpd=kernel/.$$$$ && mkdir -p $$tmpd $(dir $@) && \
+		rm -Rf $(dir $@)/vmlinuz $(dir $@)/modules && \
+		tar -x -C $$tmpd -f $< 2> /dev/null && \
+	    mv -v $$tmpd/boot/vmlinuz* $(dir $@)/vmlinuz && \
+		mv -v $$tmpd/lib/modules $(dir $@) ; \
+		rc=$$?; rm -Rf $$tmpd; exit $$rc
+	touch $@
 
 .PHONY: test-e2e
-test-e2e: fetch-kernel generate # This is invoked by a separate GHA workflow, so not combining it with the other test targets.
+test-e2e: kernel/$(ARCH)/vmlinuz melange generate # This is invoked by a separate GHA workflow, so not combining it with the other test targets.
 	go test -tags e2e ./... -race
-	$(eval KERNEL_VER := $(shell ls kernel/lib/modules))
 	cd e2e-tests && \
-	QEMU_KERNEL_IMAGE=$(realpath kernel/${ARCH}/boot/vmlinuz-virt) \
-	QEMU_KERNEL_MODULES=$(realpath kernel/${ARCH}/lib/modules/) \
-	MELANGE=$(realpath melange) \
+	QEMU_KERNEL_IMAGE=$(TOP_D)/kernel/$(ARCH)/vmlinuz \
+	QEMU_KERNEL_MODULES=$(TOP_D)/kernel/$(ARCH)/modules/ \
+	MELANGE=$(TOP_D)/melange \
 	./run-tests
 
 .PHONY: clean
```

I did the above, which is more of a copy-pasta from wolfi with the change from 'authget' to 'curlget'

smoser (approved):

PR #2138, opened 2025-08-22
Qemu runner: support builds as non-root user in guest

stevebeattie:
During the development of the QEMU runner, the ability for builds and package tests running under that runner lost the ability to be run as non-root, due to some of the environmental controls happening under the same user as the build or tests. This merge request attempts to fix that, with the intended future direction that QEMU runner builds and tests happen as non-root user by default, with overrides in the packaging where privileged operations are necessary.

The branch takes the following steps:

1. Introduces e2e_tests to be run via the run-tests script exercising builds as root, the build user, and a non-standard added user. The QEMU runner fails under the second and third, while the bubblewrap and docker ones fail on the third.
2. Fixes the bubblewrap failure, which is due it using the `uid` of the run-as user as the `gid` as well, which is not guaranteed to exist. It addresses this by looking up the primary group for the run-as user and using that as the `gid`.
3. Renames the SSH client structures to better indicate which one is the build and which one is the control session, the latter of which should always be privileged. No functional changes are introduced by this commit.
4. Introduces a 3rd SSH client session, which is a control session in the chrooted build environment, and adjusts which one is used for each interaction with the QEMU guest such that, when specifying a non-root user either in the build phase or the test phase, the build will succeed unless the build or test itself attempts a privileged operation.
5. Update the documentation around the yaml build file configuration.

I have not (yet) addressed the docker failure in the non-standard unprivileged user configuration; this is due to `/home/build` being owned by the `build` user and also being mode 700; in the bubblewrap runner, `/home/build` is similarly owned but has 777 permissions, so a different user can write there. In QEMU, `/home/build` gets its ownwership set to the user passed via run-as.

One more thing to note is that it seems the `run-tests` script is not actually run as part of the PR CI actions.

v2: fix up bubblewrap runner to fall back to GID == UID if no GID is specified for the user, to fix up a failing unit test bubblewrap; this may fail if there is no corresponding GID that matches UID in the build/test environment. Also add a test to check bubblewrap startup with GID != UID.

v3: updated to
- created a check-accounts pipeline to reduce the copy-pasta in the tests a bit
- include tests where the run-as user is specified by numeric value instead of by name
- added a check for the default user/group builds and tests are run under, which fails under bubblerwrap because the default is different (build/1000) there
- fixed up an added ssh invocation that I added in a different PR to report the qemu guest's kernel version during the build/test

v4: added subpackage tests to verify that they can run as different users than the parent packaghe builds are run as. Melange does not support this for subpackage builds.

v5: update the documentation around the yaml file format to describe the accounts subsection, along with a brief description about when adjusting the user to build/test under might make sense.

### DIFF
--- a/docs/BUILD-FILE.md
+++ b/docs/BUILD-FILE.md
@@ -4,16 +4,18 @@ This documents the melange build file structure, fields, when, and why to use va
 
 # High level structure overview
 
-The following are the high level sections for the build file, with detailed descriptions for each of them, and their fields in the sections following. 
+The following are the high level sections for the build file, with detailed descriptions for each of them, and their fields in the sections following.
 
- ## Required
- ### package
+## Required
+### package
 
    Package metadata about this package, name, version, etc.
- ### environment
+
+### environment
 
     Specification for the packages build environment
- ### pipeline
+
+### pipeline
 
     Ordered list of pipelines that produce this package
 
@@ -21,18 +23,23 @@ The following are the high level sections for the build file, with detailed desc
 ### subpackages
 
    List of subpackages that this package also produces. For example, docs.
+
 ### data
 
    Arbitrary list of data available for templating in the pipeline.
+
 ### [update](./UPDATE.md)
 
    Defines how this package is auto updated
+
 ### vars
 
    Map of arbitrary variables available for templating in the pipeline.
+
 ### [var-transforms](./VAR-TRANSFORMS.md)
 
    List of transformations to create for the builtin template variables.
+
 ### options
 
    Deviations to the build
@@ -43,20 +50,20 @@ Details about the particular package that will be used to find and use it.
 
 ### name
 Unique name for the package. Convention is to use the same name as the YAML file without extension. This is what people will search for, so it's a good idea to keep it consistent with how the package is named in other distributions. for example:
-```
+```yaml
 name: python-3.10
 ```
 
 ### version
 Version of the package. For example:
-```
+```yaml
 version: 3.10.12
 ```
 
 ### epoch
 Monotonically increasing value (starting at 0) indicating same version of the
 package, but with changes (security patches for example) applied to it.
-```
+```yaml
 epoch: 0
 ```
 
@@ -66,7 +73,7 @@ form: `<name>-<version>-r<epoch>.apk` for our example above, this would be:
 
 ### description
 Human readable description of the package. Make this meaningful, as this information shows up when searching for the package with apk, for example:
-```
+```yaml
 description: "the Python programming language"
 ```
 
@@ -100,7 +107,7 @@ The license paths that this license applies to
 Attestations for this license.
 
 For example, saying that this entire package has license `PSF-2.0`
-```
+```yaml
 copyright:
   - license: PSF-2.0
 ```
@@ -113,7 +120,7 @@ List of packages that this package depends on at runtime, but not during build
 time. These will get installed by apk as system dependencies when the package is
 installed. For example, saying that a package depends on `openssl`, `socat`, and
 `curl` at runtime:
-```
+```yaml
 dependencies:
   runtime:
     - openssl
@@ -126,7 +133,7 @@ Provides allows you to create "aliases" for a package. If your `package.name` is
 for example `php-8.1`, but you want somebody be able to get this package by
 `php`, you could provide a section like this:
 
-```
+```yaml
   dependencies:
     provides:
       - php=8.1.23
@@ -137,7 +144,7 @@ provide a floating version, so that when the package gets upgraded, the user
 will get the latest one. For that melange provides a `${{package.full-version}}`
 variable. It gets expanded to `${{package.version}}-r${{package-epoch}}`. So for
 the example above, you could do this
-```
+```yaml
   dependencies:
     provides:
       - php=${{package.full-version}}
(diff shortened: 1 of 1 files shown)


dannf (commented):
I love this! Thanks for the clear commit messages and commit organization.

Do we need a test that YAML w/ unspecified `accounts` get the expected defaults?

Very optional, but we could create a pipeline .yaml for this test so each copy can just:

```
- uses: validate-user-ids
  with:
     id: "12345"
     username: "nonroot"
     gid: "12346
     groupname: "nongroup"
```

Mind adding docs to tell developers when they should use nonroot/root/build? Do we want to explicitly have this filled out in each YAML file? Do we want to eventually drop "build" and have everything use "root" or "nonroot"?

pkg/container/qemu_runner.go
```diff
@@ -126,7 +126,7 @@ func (bw *qemu) Debug(ctx context.Context, cfg *Config, envOverride map[string]s
 	if pubKey != nil {
 		command := fmt.Sprintf("echo %q | tee -a /root/.ssh/authorized_keys /home/*/.ssh/authorized_keys", strings.TrimSpace(string(pubKey)))
 		err := sendSSHCommand(ctx,
-			cfg.WorkspaceClient,
+			cfg.SSHControlClient,
```
maxgio92, line 129: Shouldn't we use the build client here (i.e. `SSHBuildClient`)? I would expect in this connection to mimic what happens at build/test time. For instance, when specifying a non-root run-as user, I would expect to debug with the same user. That would apply also to the chroot-ed root, I think.

stevebeattie: This step is injecting the user's public keys into the VM guest so that one can ssh inside beyond the debug session; the debugging session that one is dropped into in an interactive session is setup later in the function.

This injection needs to be done via the control connection because of the way the ssh daemon is configured in the guest; successful connections through the default ssh port (22) get chrooted into `/mount` where the build environment occurs, and sshd should not honor `.ssh/` directories in that chroot tree.

maxgio92: Thank you @stevebeattie for clarifying it, and thank you for doing this. This is one step further into unprivileged builds with QEMU, love it

stevebeattie:
@dannf thanks for the great feedback!

> Do we need a test that YAML w/ unspecified `accounts` get the expected defaults?

Yes, the challenge for the test is that the defaults are different across the runners: qemu and docker are currently root/0, whereas bubblewrap's default is build/1000. I had intended to add the default check when I get to the step of defaulting the qemu runner to match the bubblewrap runner, build/1000, but went ahead and added it as a separate commit in this merge request.

> Very optional, but we could create a pipeline .yaml for this test so each copy can just:
> 
> ```
> - uses: validate-user-ids
>   with:
>      id: "12345"
>      username: "nonroot"
>      gid: "12346
>      groupname: "nongroup"
> ```

Oh, nice idea! I went ahead and did this with uh different names (I forgot your suggestions while poking at the details of pipelines and how they're used). I don't feel strongly about the names, so can change them if someone has stronger opinions.

> Do we want to eventually drop "build" and have everything use "root" or "nonroot"?

I just chose the username "nonroot" to inject for testcases as something different from the default root/build users that get automatically injected into the build environment. I expect using a custom user/uid for the build phase to be rarely used, if ever, but it is more useful for some testcases where there's an added user in the environment that should be used at runtime; as an example, this change was made to logstash-9 https://github.com/wolfi-dev/os/commit/d0981aea5fa5d2d9ff379d7c817b9385be151c97#diff-bed3014e41445e3d861ca6b9018aaf657e5d44a2f16e12d28fc6e2b4f8d0f62eL301-R302 I think because of non-root tests not working under QEMU.

Suffice to say, I don't really care what the name of the non-privileged user is, I want to stop having QEMU builds run as root by default.

> Do we want to explicitly have this filled out in each YAML file?

I strongly prefer sensible and safe defaults, which would mean unprivileged builds in the QEMU guest by default, with the ability to override for situations that need privilege in the build/install process. As such, I don't think we should require adding boilerplate in each YAML file. Is there a reason we might want to do that?

Once melange defaults to non-privileged builds by default after adjusting the yaml files for packages that need to build as root because they do things like create setuid binaries or set fscaps, then we can look at alternate ways of specifying such things in the melange yaml configuration to further reduce the set of package builds that need privilege.

> Mind adding docs to tell developers when they should use nonroot/root/build?

Don't mind at all, will work on that.

egibs (approved):
This is great. LGTM, but deferring to OS for a separate approval.

sanmai-NL:
The same fault happens, through a different conditional branch, when `run-as` has a `uid` instead of username value.

stevebeattie:
@sanmai-NL I can reproduce the issue both in bubblewrap and the qemu runner. The thing that is triggering it is in the package's account config:

```yaml
  accounts:
    groups:
      - gid: 10001
        groupname: unprivileged
    run-as: unprivileged
    users:
      - shell: /bin/nologin
        uid: 10001
        username: unprivileged
```
the `unprivileged` user does not have an associated membership in the `unprivileged` group (unlike in the account setup for the test pipeline). This used to work accidentally because under the bubblewrap runner, the group ID was always set to the UID of the run-as user. The commit a3612ee3 ("bubblewrap: when running as custom UID, use corrsponding GID") changed that to run as the primary group/GID as specified in the user setup, but didn't handle the case where one was not configured, hence the SEGV and traceback.

While it's strongly recommended to be explicit about setting a gid for the user if setting one up, melange should handle this better and fall back to the old behavior of running with gid==uid if one is not specified, and not crash. Will put up a PR for this.

Thanks again for the report!

Issue #1548, opened 2024-10-08
A nested pipeline's environment does not pass through to a `uses` step.


smoser:
I was trying to use `environment` from a nested `pipeline` in a `uses` stanza.  The environment variable didnt' pass through as I wanted.

```
$ melange version | grep ^[A-Z]
GitVersion:    v0.12.1-7-g94476be
GitCommit:     94476be97a34c3d0b0dece6c4b2ec596960b1b34
GitTreeState:  clean
BuildDate:     '2024-09-26T17:58:32Z'
GoVersion:     go1.23.1
Compiler:      gc
Platform:      linux/amd64

```

Given the two files:
1. `test-me.yaml`
```yaml
# test-me.yaml
package:
  name: test-me
  version: 1.0
  epoch: 0

environment:
  contents:
    packages:
      - busybox
  environment:
    BAR: "the other day I saw a BAR"

pipeline:
  - environment:
      FOO: "WINNER WINNER"
    pipeline:
      - runs: |
          echo "==== FOO=$FOO ==="
          echo "==== BAR=$BAR ==="
      - uses: test/dump-env
```

2. `pipelines/test/dump-env`
```yaml
name: dump your environment
pipeline:
  - name: "show env"
    runs: |
      fails=0
      [ -n "$FOO" ] && echo "PASS: FOO got $FOO" ||
        { echo "FAIL: FOO was empty"; fails=$((fails+1)); }
      [ -n "$BAR" ] && echo "PASS: BAR got $BAR" ||
        { echo "FAIL: BAR was empty"; fails=$((fails+1)); }
      exit $fails
```

build command and output are below.  you can see that the 'run' section gets both the global environment (BAR) and the "local" environment ("FOO"), but the `uses` pipeline does not get access to the local variable (FOO).

```
$ melange build test-me.yaml --repository-append=./packages \
  --keyring-append=local-melange.rsa.pub \
  --signing-key=local-melange.rsa \
  --arch=x86_64 --pipeline-dir=./pipelines/  \
  -k https://packages.wolfi.dev/os/wolfi-signing.rsa.pub \
  -r https://packages.wolfi.dev/os
...
2024/10/08 15:00:11 INFO ==== FOO=WINNER WINNER ===
2024/10/08 15:00:11 INFO ==== BAR=the other day I saw a BAR ===
2024/10/08 15:00:11 INFO running step "test/dump-env"
2024/10/08 15:00:11 INFO running step "show env" uses=test/dump-env
[...]

PR #2139, opened 2025-08-22
Propagate environment between uses and pipelines

justinvreeland:
If you specify the environment of a uses pipeline it will be ignored when the pipeline is loaded.

You can see grpc-1.76 doing this here:

https://github.com/wolfi-dev/os/blob/0c6b8959949e816f3dbce1af2237ad81ee579618/grpc-1.74.yaml#L120

These variables aren't ever actually set the uses pipeline overwrites it. So we need to fix those places so that the variables are propegated. An easy fix for this when it's needed is to make then environment variables at the top level with the packages. This also explains some additional strange behavior we've had trying to override variables for individual steps.

### Functional Changes

Notes:

We should at a minimum verify that the grpc package still build correctly since some of them vendor things incidentally a spot check will be needed for packages that also build this.

Manual checking doesn't show problems with grpc packages or other that display this.

### DIFF
--- a/e2e-tests/env-propagation-uses-build.yaml
+++ b/e2e-tests/env-propagation-uses-build.yaml
@@ -1,0 +1,29 @@
+package:
+  name: uses-env-propagation
+  version: "0.0.1"
+  epoch: 0
+  description: Test environment propagation to external pipeline files via uses
+  copyright:
+    - license: APACHE-2.0
+
+environment:
+  contents:
+    repositories:
+      - https://packages.wolfi.dev/os
+    keyring:
+      - https://packages.wolfi.dev/os/wolfi-signing.rsa.pub
+    packages:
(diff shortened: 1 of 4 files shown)


justinvreeland:
> Maybe you're fixing https://github.com/chainguard-dev/melange/pull/1785 ?
I don't think so. It doesn't seem to rely on environment variables.

justinvreeland:
Packages I've see that do this seem to be building fine like this. Not sure where we might hit other sharp edges.

justinvreeland:
> > > Maybe you're fixing #1785 ?
> > 
> > 
> > I don't think so. It doesn't seem to rely on environment variables.
> 
> what about if I paste the right issue ? #1548

Yeap, it fixes that. Added it to the pr description.

PR #2146, opened 2025-08-28
pipelines/go/build: Add ignore-untracked-files to prevent some +dirty…

justinvreeland:
… builds

Right now we have some files that are added to the build env that may cause binaries to be built with +dirty in the verison when inappropriate. This adds and option to ignore those loose files.

It doesn't cover modifications to tracked files and commits will still be appended to the version.

### DIFF
--- a/pkg/build/pipelines/go/build.yaml
+++ b/pkg/build/pipelines/go/build.yaml
@@ -101,8 +101,21 @@ inputs:
       If true, "go mod tidy" will run before the build
     default: "false"
 
+  ignore-untracked-files:
+    description: |
+      If true, we will provide a gitignore that ignore all untracked files.
+    default: "true"
+
 pipeline:
   - runs: |
+
+      # We sometimes have stray files in our build environment
+      # That we don't neccesarily want surfaced in the build version
+      if [ "${{inputs.ignore-untracked-files}}" = "true" ]; then
+        mkdir -p ~/.config/git/
+        echo '**' > ~/.config/git/ignore
+      fi
+
       cd "${{inputs.modroot}}"
 
       # check if modroot is set correctly by checking go.mod file exist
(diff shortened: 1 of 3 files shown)


pkg/build/pipelines/go/build.yaml
```diff
@@ -135,3 +149,8 @@ pipeline:
       [ -e /home/build/go.sum.local ] && cp /home/build/go.sum.local go.sum
 
       GOAMD64="${{inputs.amd64}}" GOARM64="${{inputs.arm64}}" GOEXPERIMENT="${{inputs.experiments}}" go build -o "${{targets.contextdir}}"/${BASE_PATH} -tags "${{inputs.toolchaintags}},${{inputs.tags}}" -ldflags "${LDFLAGS}" -trimpath -buildmode ${{inputs.buildmode}} ${{inputs.extra-args}} ${{inputs.packages}}
+
+      # Clean up the gitignore file we created to avoid interfering with subsequent builds
+      if [ "${{inputs.ignore-untracked-files}}" == "true" ]; then
```
sergiodj:
```suggestion
      if [ "${{inputs.ignore-untracked-files}}" = "true" ]; then
```

pkg/build/pipelines/go/build.yaml
```diff
@@ -101,8 +101,22 @@ inputs:
...
+    description: |
+      If true, we will provide a gitignore that ignore all untracked files.
+    default: "true"
+
 pipeline:
   - runs: |
+
+      # We sometimes have stray files in our build environment
+      # That we don't neccesarily want surfaced in the build version
+      if [ "${{inputs.ignore-untracked-files}}" == "true" ]; then
```
sergiodj:
```suggestion
      if [ "${{inputs.ignore-untracked-files}}" = "true" ]; then
```

Pedantically speaking, string comparisons are done with `=` in shell scripts.

pkg/build/pipelines/go/build.yaml
```diff
@@ -101,8 +101,22 @@ inputs:
...
+        mkdir -p ~/.config/git/
+        echo "**" > ~/.config/git/ignore
+        echo ".*" >> ~/.config/git/ignore
```
sergiodj:
```suggestion
        echo '**' > ~/.config/git/ignore
        echo '.*' >> ~/.config/git/ignore
```

Might want to use single quote here to avoid any unwanted shell expansion.

Also, from the [documentation](https://git-scm.com/docs/gitignore#_pattern_format), I believe `**` already matches everything, so maybe we don't need `.*`?

sergiodj (changes requested):

pkg/build/pipelines/go/build.yaml
justinvreeland: Sounds good i'll check this.

sergiodj (dismissed):

PR #2154, opened 2025-09-10
Qemu runner: debug initial connection

stevebeattie:
We are seeing some build failures in production where melange does not 
appear to be able to get the openssh protocol version string to know
that the ssh service has successfully started, though log messages
indicate that the ssh daemon has started enough to notice melange's
attempts to make initial contact.

To try to address this there are two separate changes here:
- bump the logging level from debug to info on the error message when the initial ssh check fails
- increase the connection Dial timeout from 50ms to 150ms to see if resource contention is causing the initial detection that the ssh daemon is started to fail

These are created independently so that if we'd like to cherrypick one or the other, we can do so.


89luca89 (approved):

89luca89:
looks good to me, I'd even have raised the timeout to 500ms just to be sure, but I assume 150ms can be enough to detect resource starvation

PR #2155, opened 2025-09-10
qemu runner: make virtfs 9p bindings read-only

stevebeattie:
The QEMU runner uses virtfs 9p mount bindings to inject some artifacts into the build environment, but unlike the docker and bubblewrap runners, does not use any bind mounts to extract out artifacts to produce apks; instead tar over ssh is used for artifact extraction.

Thus, to reduce risk of a malicious build being able to write artifacts outside of the QEMU guest, make the virtfs 9p bindings be read-only. (There has also has been at least one vulnerability the virtfs 9p implementation that allowed writes outside of the specified path.)

Note that in order for this to land, a [change to the microvm-init package in wolfi-dev/os]( **must** land that removes a chmod that happens on the workspace directory (`/mount/mnt`) or else builds will fail.

My longer term plan is to drop the use of the 9p virtfs mounts entirely in the QEMU runner and inject per-package additions through other means, to reduce the mechanisms that the QEMU guest can influence the hosting melange environment.

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -707,11 +707,11 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 	// kill all performances (lots of small files)
 	// instead we will copy back the finished workspace artifacts when done.
 	// this dramatically improves compile time, making them comparable to bwrap or docker runners.
-	baseargs = append(baseargs, "-fsdev", "local,security_model=mapped,id=fsdev100,path="+cfg.WorkspaceDir)
+	baseargs = append(baseargs, "-fsdev", "local,security_model=mapped,id=fsdev100,readonly=on,path="+cfg.WorkspaceDir)
 	baseargs = append(baseargs, "-device", "virtio-9p-pci,id=fs100,fsdev=fsdev100,mount_tag=defaultshare")
 
 	if cfg.CacheDir != "" {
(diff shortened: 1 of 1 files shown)


pkg/container/qemu_runner.go
```diff
@@ -941,24 +942,77 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
...
+		if err != nil {
+			return err
+		}
+	}
+
+	clog.FromContext(ctx).Info("qemu: unmounting host workspace from guest")
+	// now that we've copied files from the host filesystem into the
+	// guest's workspace, we can umount the filesystem, as nothing in
+	// the build should reference the original location and it reduces
+	// the ability of later build steps to possibly intteract with the
```
egibs:
```suggestion
	// the ability of later build steps to possibly interact with the
```

egibs (approved):
The new changes make sense to me.

- If `microvm-init` handles the initial `defaultshare` mount, we'll remount it during the cutover period
- Once `microvm-init` no longer creates the `defaultshare` mount, Melange will handle it going forward
- The file copying/ownership is now handled by a new, intermediate command
- We drop the mount at the very end once everything is copied over

egibs (approved):

egibs (approved):

89luca89 (approved):

PR #2165, opened 2025-09-30
cmake pipeline: add verbosty to cmake/install

stevebeattie:
When debugging behaviors in the install phase of a build, it's useful to have verbose output.

Also, quote the directory name argument to cmake --install

### DIFF
--- a/pkg/build/pipelines/cmake/install.yaml
+++ b/pkg/build/pipelines/cmake/install.yaml
@@ -13,4 +13,4 @@ inputs:
 
 pipeline:
   - runs: |
-      DESTDIR="${{targets.contextdir}}" cmake --install ${{inputs.output-dir}}
+      VERBOSE=1 DESTDIR="${{targets.contextdir}}" cmake --install "${{inputs.output-dir}}"


pkg/build/pipelines/cmake/install.yaml
```diff
@@ -13,4 +13,4 @@ inputs:
 
 pipeline:
   - runs: |
-      DESTDIR="${{targets.contextdir}}" cmake --install ${{inputs.output-dir}}
+      VERBOSE=1 DESTDIR="${{targets.contextdir}}" cmake --install "${{inputs.output-dir}}"
```
egibs, line 16: Should `VERBOSE` be a configurable input that defaults to `0`?

I don't have a strong preference one way or the other.

stevebeattie: Hrm, actually, in trying to demonstrate the difference between the default and VERBOSE=1 I uh have discovered in the case that I am investigating that there's not any difference in the output and I'm getting the info reported that I was trying to get which led me down this path to begin with. So I no longer think this change is necessary. Closing.

(Fixing up the quoting would be good as a separate change evaluating all of the pipelines.)

PR #2166, opened 2025-10-03
Set $TERM when entering debug shell

dannf:
Boy am I tired of typing `apk add emacs; export TERM=xterm-256color` each time I enter a debug shell.


justinvreeland:
IMO. This should be a configuration option in a file somewhere not something we hard code into melange. I know the probability of someone having a terminal that doesn't conform is low. But i don't like that idea of hard coding one.   Maybe pull the variable from the env melange is run in? Also doesn't this also assume that we have the adaquate terminfo files in the debug env somewhere?

dannf:
Using the host TERM setting seems reasonable to me, I'll switch to that. My use cases all work w/o the terminfo files, so I think this is generally an improvement.

dustinkirkland (approved):

dannf:
> Maybe pull the variable from the env melange is run in?

Turns out, the qemu runner uses this logic already:
https://github.com/chainguard-dev/melange/blob/e4b918fc8756f62b85d934e9c883d65de5bdc6d3/pkg/container/qemu_runner.go#L222

I'm doing it the same way now.

PR #2191, opened 2025-10-21
split/lib: add new pipeline for shared libraries

javacruft:
Add a new pipeline to help build shared library subpackages alongside
-{dev,static,bin}.

This supports general migration towards a shared library naming policy
that enables seamless transitions between soversions.


pkg/build/pipelines/split/lib.yaml
EyeCantCU, line 1: Could we parameterize the library path so that private libraries could be split too?

EyeCantCU (approved):

PR #2192, opened 2025-10-21
Add additional linters for duplicate files, DLLs, unsupported architectures, and more

egibs:
This PR adds seven new, non-blocking (for now) linters which focus on:
- Static archives (`.a` files) in non-dev packages
- Windows DLLs (caveat, only DLLs that don't exist in places like `.jar` files or other archives)
- Mach-O DYLIBs (same caveat as above)
- References to unsupported architectures
- References to non-Linux platforms
- Duplicate files (same name and contents in different directories)
- Binary architectures

### DIFF
--- a/docs/md/melange_build.md
+++ b/docs/md/melange_build.md
@@ -54,7 +54,7 @@ melange build [flags]
   -k, --keyring-append strings                                  path to extra keys to include in the build environment keyring
       --license string                                          license to use for the build config file itself (default "NOASSERTION")
       --lint-require strings                                    linters that must pass (default [dev,infodir,setuidgid,tempdir,usrmerge,varempty,worldwrite])
-      --lint-warn strings                                       linters that will generate warnings (default [cudaruntimelib,dll,duplicate,lddcheck,maninfo,nonlinux,object,opt,pkgconf,python/docs,python/multiple,python/test,sbom,srv,staticarchive,strip,unsupportedarch,usrlocal])
+      --lint-warn strings                                       linters that will generate warnings (default [binaryarch,cudaruntimelib,dll,duplicate,dylib,lddcheck,maninfo,nonlinux,object,opt,pkgconf,python/docs,python/multiple,python/test,sbom,srv,staticarchive,strip,unsupportedarch,usrlocal])
       --memory string                                           default memory resources to use for builds
       --namespace string                                        namespace to use in package URLs in SBOM (eg wolfi, alpine) (default "unknown")
       --out-dir string                                          directory where packages will be output (default "./packages/")
--- a/docs/md/melange_lint.md
+++ b/docs/md/melange_lint.md
@@ -30,7 +30,7 @@ melange lint [flags]
 ```
   -h, --help                   help for lint
       --lint-require strings   linters that must pass (default [dev,infodir,setuidgid,tempdir,usrmerge,varempty,worldwrite])
-      --lint-warn strings      linters that will generate warnings (default [cudaruntimelib,dll,duplicate,lddcheck,maninfo,nonlinux,object,opt,pkgconf,python/docs,python/multiple,python/test,sbom,srv,staticarchive,strip,unsupportedarch,usrlocal])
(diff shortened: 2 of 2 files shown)


pkg/linter/linter.go
```diff
@@ -234,6 +269,221 @@ func objectLinter(_ context.Context, _ *config.Configuration, _, path string) er
...
+
+func dllLinter(_ context.Context, _ *config.Configuration, _, path string) error {
+	if filepath.Ext(path) == ".dll" {
+		return fmt.Errorf("package contains DLL file %q", path)
+	}
+	return nil
+}
+
+func nonLinuxLinter(_ context.Context, _ *config.Configuration, _, path string) error {
+	names := []string{"darwin", "osx", "windows"}
```
justinvreeland: might as well add macOS/macos here, it would not be unreasonable to find a ton of these in `-dev` packages that just pull their whole set of headers into the install.

pkg/linter/linter.go
```diff
@@ -234,6 +269,221 @@ func objectLinter(_ context.Context, _ *config.Configuration, _, path string) er
 	return nil
 }
 
+func staticArchiveLinter(_ context.Context, _ *config.Configuration, pkgname, path string) error {
+	// Static archives are expected in -dev packages
+	if strings.HasSuffix(pkgname, "-dev") {
```
justinvreeland: We have some packages that are `-static` because alpine does so we can add that here too.

pkg/linter/linter.go
```diff
@@ -699,19 +949,94 @@ func lddcheckTestLinter(_ context.Context, cfg *config.Configuration, pkgname, p
...
-func lintPackageFS(ctx context.Context, cfg *config.Configuration, pkgname string, fsys fs.FS, linters []string) error {
-	errs := []error{}
+func formatLintResults(results []LintResult, severity string, useColor bool) string {
+	if len(results) == 0 {
+		return ""
+	}
+
+	// ANSI color codes
+	const (
+		colorReset  = "\033[0m"
```
justinvreeland: Nit: 

I approve of the formatting. I also have Opinions on how we do output from these things but given that as far as I'm aware the majority of logging in melange is slog/charm we should stick with that until we come up with a different solution. I'm not a huge fan of embedding our own coloring into the loggin that also colors things for us. Though I do think this will be easier to read/find than just slog/charm.

pkg/linter/linter.go
```diff
@@ -205,6 +215,31 @@ var linterMap = map[string]linter{
 		Explain:         "CUDA driver-specific libraries should be passed into the container by the host. Installing them in an image could override the host libraries and break GPU support. If this library is needed for build-time linking or ldd-check tests, please use a package containing a stub library instead. For libcuda.so, use nvidia-cuda-cudart-$cuda_version. For libnvidia-ml.so, use nvidia-cuda-nvml-dev-$cuda_version.",
 		defaultBehavior: Warn,
 	},
+	"dll": {
```
justinvreeland, line 211: If we're checking for dlls we should check for dylibs too.

justinvreeland (changes requested):
Always happy to see new linters.  Mixed feelings about the new formatting.

If we're already iterating over all the paths it might be worth checking the architecture of binary files as well. There's probably a go library somewhere that extracts the architecture from elf files.

pkg/linter/linter.go
egibs: Added in [`91cdddd` (#2192)](https://github.com/chainguard-dev/melange/pull/2192/commits/91cddddfc3ac14af01f746ac4450c89480e07f36).

pkg/linter/linter.go
egibs: Reverted most of the changes in [`91cdddd` (#2192)](https://github.com/chainguard-dev/melange/pull/2192/commits/91cddddfc3ac14af01f746ac4450c89480e07f36).

pkg/linter/linter.go
egibs: Updated in [`91cdddd` (#2192)](https://github.com/chainguard-dev/melange/pull/2192/commits/91cddddfc3ac14af01f746ac4450c89480e07f36).

pkg/linter/linter.go
egibs: Added in [`91cdddd` (#2192)](https://github.com/chainguard-dev/melange/pull/2192/commits/91cddddfc3ac14af01f746ac4450c89480e07f36).

stevebeattie (approved):
This is great, thanks @egibs !

Long term, we might want to look at de-duplicating license files, but they're relatively small in the scheme of things.

PR #2196, opened 2025-10-23
feat: Reorganize linters; improve linter output structure; add support for storing JSON results

egibs:
We currently display linter findings as warning/error logs during a package build. This isn't very useful unless one happens to be looking at build logs and these logs are eventually deleted after a brief period of time (or cleared when building locally) so any historical reference is minimal at best. 

The current output is also hard to parse and does not include all of the findings. Likewise, the `linter.go` file grows each time we add a new linter and it's a bit hard to scroll through (~1300 SLOC before these changes).

So, this PR:
- reorganizes the current linters into their own package
- moves existing code into smaller, more maintainable files
- retains the current linter stdout functionality (and slightly improves it due to the use of the structured results) but adds an opt-in flag to write the results as structured JSON to a local file (loosely inspired by `--create-build-log`).
- adds the same results output functionality to the `melange lint` command and also supports providing a custom output directory

For both commands, the default output directory is `packages/<arch>` and the file naming mirrors the APK name with `lint-` prefixed to the full package version.

The net-new code can be found in these files:
- `pkg/build/build.go`
- `pkg/build/options.go`
- `pkg/cli/build.go`
- `pkg/cli/lint.go`
- `pkg/linter/linters/common.go` (`AllPaths` now handles the structured logging and we don't pass the linter functions via `allPaths(...)`)
- `pkg/linter/types/types.go`
- `pkg/linter/log.go`
- `pkg/linter/results.go`

Example of the new linter output: https://github.com/chainguard-dev/melange/actions/runs/18759749787/job/53520911139?pr=2196#step:12:120

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -36,6 +36,7 @@ require (
 	go.opentelemetry.io/otel v1.38.0
 	go.opentelemetry.io/otel/exporters/stdout/stdouttrace v1.37.0
 	go.opentelemetry.io/otel/sdk v1.38.0
+	go.yaml.in/yaml/v2 v2.4.2
 	golang.org/x/crypto v0.42.0
 	golang.org/x/exp v0.0.0-20250819193227-8b4c13bb791b
 	golang.org/x/sync v0.17.0
@@ -58,7 +59,6 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus v1.1.0 // indirect
 	github.com/grpc-ecosystem/go-grpc-middleware/v2 v2.3.2 // indirect
 	go.opencensus.io v0.24.0 // indirect
-	go.yaml.in/yaml/v2 v2.4.2 // indirect
 	golang.org/x/tools v0.36.0 // indirect
 	k8s.io/klog/v2 v2.130.1 // indirect
 )


pkg/linter/linter.go
```diff
@@ -99,1014 +75,360 @@ func DefaultWarnLinters() []string {
...
 	}
 
-	return nil
+	return lintErr
 }
 
 // Lint the given APK at the given path
-func LintAPK(ctx context.Context, path string, require, warn []string) error {
+// If outputDir is provided, lint results will be saved to JSON files
+func LintAPK(ctx context.Context, path string, require, warn []string, outputDir string) error {
```
stevebeattie: I think because outDir is defined be default in the cobra command, outputDir will always be non-nil when passed here, which means that `melange lint` will always generate the json output files; it seems the only way to not get files generated if linter warnings exist is to do `melange lint --out-dir='' APK` . I'm not sure if that's the desired behavior.

egibs: Correct, it is the desired behavior currently. I think I'll make it opt-in for both `melange build` and `melange lint` for consistency.

egibs: Updated in [`c5e2260` (#2196)](https://github.com/chainguard-dev/melange/pull/2196/commits/c5e2260515db41c5c18912c7a7a405e3dbedb05a).

PR #2204, opened 2025-10-29
chore: migrate to golangci-lint v2; add config file; address findings

egibs:
I noticed that we don't have a `.golangci.yml` file for Melange like we do elsewhere. This PR adds the same config that we use for Apko, addresses the findings, and also updates our `verify` Workflow to use the new v2 schema.

This should help keep additions consistent (historically, I'd inadvertently format the entire codebase or something like that) and following best-practices.

None of the changes should affect behavior except for where we write files as `0644` instead of `0755` in `docs/cmd/pipeline-reference-gen/main.go`.

### DIFF
--- a/pkg/build/build.go
+++ b/pkg/build/build.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"io/fs"
 	"log/slog"
+	"maps"
 	"math"
 	"os"
 	"path/filepath"
@@ -47,7 +48,6 @@ import (
 	"github.com/yookoala/realpath"
 	"github.com/zealic/xignore"
 	"go.opentelemetry.io/otel"
-	"golang.org/x/exp/maps"
 	"golang.org/x/sys/unix"
 	"sigs.k8s.io/release-utils/version"
 
@@ -280,9 +280,7 @@ func New(ctx context.Context, opts ...Option) (*Build, error) {
 		log.Infof("applying configuration patches for build option %s", optName)
 
 		if opt, ok := b.Configuration.Options[optName]; ok {
-			if err := b.applyBuildOption(opt); err != nil {
-				return nil, err
-			}
+			b.applyBuildOption(opt)
 		}
 	}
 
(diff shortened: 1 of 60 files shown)


pkg/build/build.go
```diff
@@ -280,9 +280,7 @@ func New(ctx context.Context, opts ...Option) (*Build, error) {
 		log.Infof("applying configuration patches for build option %s", optName)
 
 		if opt, ok := b.Configuration.Options[optName]; ok {
-			if err := b.applyBuildOption(opt); err != nil {
-				return nil, err
-			}
+			b.applyBuildOption(opt)
```
justinvreeland, line 283: This isn't a behavior change?

pkg/source/source.go
```diff
@@ -115,15 +118,15 @@ func FetchSourceFromMelange(ctx context.Context, filePath, destDir string) (*con
...
+		return nil, fmt.Errorf("failed to create temporary directory: %w", err)
 	}
 	defer os.RemoveAll(tmpDir)
 
 	// Check if the file is an apk package
 	isApk := false
 	if filepath.Ext(filePath) == ".apk" {
 		if err := extractMelangeYamlFromTarball(filePath, tmpDir); err != nil {
-			log.Fatalf("Failed to extract melange yaml from apk package: %v", err)
+			return nil, fmt.Errorf("failed to extract melange yaml from apk package: %w", err)
```
justinvreeland, line 129: This seems strange for an automated fix. I'm guess the assuming is that the caller will log the error?

pkg/build/build.go
egibs: The original function only ever returned nil at the very end:
```go
// applyBuildOption applies a patch described by a BuildOption to a package build.
func (b *Build) applyBuildOption(bo config.BuildOption) error {
	// Patch the variables block.
	if b.Configuration.Vars == nil {
		b.Configuration.Vars = make(map[string]string)
	}

	for k, v := range bo.Vars {
		b.Configuration.Vars[k] = v
	}

	// Patch the build environment configuration.
	lo := bo.Environment.Contents.Packages
	b.Configuration.Environment.Contents.Packages = append(b.Configuration.Environment.Contents.Packages, lo.Add...)

	for _, pkg := range lo.Remove {
		pkgList := b.Configuration.Environment.Contents.Packages

		for pos, ppkg := range pkgList {
			if pkg == ppkg {
				pkgList[pos] = pkgList[len(pkgList)-1]
				pkgList = pkgList[:len(pkgList)-1]
			}
		}

		b.Configuration.Environment.Contents.Packages = pkgList
	}
	
	return nil
}

pkg/source/source.go
egibs: Right. `log.Fatalf` will prevent the `defer` from running and everything else was already returning an error.

justinvreeland: Thanks!

PR #2207, opened 2025-10-31
qemu: support injecting kernel cmdline args via QEMU_KERNEL_CMDLINE

stevebeattie:
Support debugging the kernel build environment by allowing the passing of additional kernel cmdline arguments via the environment variable QEMU_KERNEL_CMDLINE. These will be injected as is, so multiple arguments need to be space delimited in the variable; e.g:

  QEMU_KERNEL_CMDLINE="slub_debug=- apparmor=1 lsm=apparmor"

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -707,6 +707,13 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 		baseargs = append(baseargs, "-cpu", "host")
 	}
 
+	cmdlineVar, ok := os.LookupEnv("QEMU_KERNEL_CMDLINE")
+	if ok {
+		clog.FromContext(ctx).Infof("qemu: QEMU_KERNEL_CMDLINE set to: %s", cmdlineVar)
+	} else {
+		cmdlineVar = ""
+	}
+
 	// ensure we disable unneeded devices, this is less needed if we use microvm machines
 	// but still useful otherwise
 	baseargs = append(baseargs, "-display", "none")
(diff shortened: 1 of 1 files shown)


pkg/container/qemu_runner.go
```diff
@@ -699,6 +699,13 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 		baseargs = append(baseargs, "-cpu", "host")
 	}
 
+	cmdlineVar, ok := os.LookupEnv("QEMU_KERNEL_CMDLINE")
```
justinvreeland, line 710: Doesn't cobra support commandline flags and environment variables? If so can't we configure this through cobra instead of a random env variable that's no discoverable?

stevebeattie: Yeah, the co-developed viper module supports loading environment variables with cobra integration. Covnerting the half dozen environment variables melange already looks for into viper is probably worthwhile.

I'm on the fence about QEMU_KERNEL_CMDLINE ; it makes it really easy to inject kernel cmd line arguments for debugging, but I'm leery of having it potentially be used in  production environments, as it's a configuration of the build environment that won't get captured.

justinvreeland: That seems like we should not have it rather than not document it. If we're doing to have it we should probably log all or the sensitive settings we use somewhere. I opened https://github.com/chainguard-dev/melange/issues/2212 to track this though.  My first pass assumption is that we should just sync the flags and env variables and print them out before building so we can pull them from the logs.

eslerm (approved):

89luca89 (approved):

PR #2209, opened 2025-11-03
fix: pin cosign-installer to v3.10.1 to avoid issues with GoReleaser

egibs:
The v4 release of the `cosign-installer` Action failed with this error during our scheduled release:
```
  ⨯ release failed after 2m1s                       
    error=
    │ sign: cosign failed: exit status 1: Error: must provide --bundle with --signing-config or --use-signing-config
    │ error during command execution: must provide --bundle with --signing-config or --use-signing-config
```

This PR pins the Action to the latest version that still supports cosign v2.x to avoid this.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -65,7 +65,8 @@ jobs:
           check-latest: true
 
       # Cosign is used by goreleaser to sign release artifacts.
-      - uses: sigstore/cosign-installer@7e8b541eb2e61bf99390e1afd4be13a184e9ebc5 # v3.10.1 # latest cosign v2.x release; v3.x has issues with GoReleaser
+      # The Action is pinned to v3.10.x (cosign binary v2.x) because v4 (cosign binary v3.x) has issues with GoReleaser
+      - uses: sigstore/cosign-installer@7e8b541eb2e61bf99390e1afd4be13a184e9ebc5 # v3.10.1
         if: steps.check.outputs.need_release == 'yes'
 
       - uses: goreleaser/goreleaser-action@e435ccd777264be153ace6237001ef4d979d3a7a # v6.4.0


.github/workflows/release.yaml
```diff
@@ -61,11 +61,11 @@ jobs:
       - uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
         if: steps.check.outputs.need_release == 'yes'
         with:
-          go-version-file: './go.mod'
+          go-version-file: "./go.mod"
           check-latest: true
 
       # Cosign is used by goreleaser to sign release artifacts.
-      - uses: sigstore/cosign-installer@faadad0cce49287aee09b3a48701e75088a2c6ad # v4.0.0
+      - uses: sigstore/cosign-installer@7e8b541eb2e61bf99390e1afd4be13a184e9ebc5 # v3.10.1 # latest cosign v2.x release; v3.x has issues with GoReleaser
```
cjwatson: Shouldn't this be more like this?

```suggestion
      - uses: sigstore/cosign-installer@7e8b541eb2e61bf99390e1afd4be13a184e9ebc5 # v3.10.1 # latest cosign v3.x release; v4.x has issues with GoReleaser
```

egibs: Ah, this is a little confusing. The v3.x Action releases map to cosign v2.x and the v4.x release maps to cosign v3.x.

cjwatson: Ah, got it!

egibs: Improved the wording in [`e656eab` (#2209)](https://github.com/chainguard-dev/melange/pull/2209/commits/e656eabe9cd8ec36579825077b2dae3fad55b87d).

stevebeattie: Is there a reference to an open bug report with cosign that we can perhaps include, so that we know what to look for when to determine when to unpin?

egibs: There's actually a migration Issue in the GoReleaser repository. I'll open a new PR with the changes.

PR #2213, opened 2025-11-03
chore: pin cosign to v2.x

egibs:
Relates to: #2209

This PR pins Cosign to v2.x for the time being. The v3 migration seems relatively straightforward but would also change how we sign release artifacts, so we can just retain the original functionality while we work out those details.


stevebeattie (approved):
LGTM, thanks!

PR #2218, opened 2025-11-06
Refactor sbom generation into its own package.

wlynch:
Refactors SBOM generation into an interface so it's easier to hook in alternative implementations of SBOM generation.

This change is not intended to be a breaking change, and should (ideally) create the same SBOMs as before (though I don't know how comprehensive our testing is here).

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

### DIFF
--- a/pkg/build/sbom/spdx/spdx.go
+++ b/pkg/build/sbom/spdx/spdx.go
@@ -167,7 +167,8 @@ func (g *Generator) GenerateSBOM(ctx context.Context, gc *build.GeneratorContext
 		})
 	}
 
-	// Add upstream source packages from main package pipelines
+	// Add upstream source packages from main package pipelines to main package SBOM
+	// and to all subpackage SBOMs (since subpackages are derived from the main source)
 	for i, p := range gc.Configuration.Pipeline {
 		uniqueID := strconv.Itoa(i)
 		upstreamPkg, err := p.SBOMPackageForUpstreamSource(gc.Configuration.Package.LicenseExpression(), gc.Namespace, uniqueID)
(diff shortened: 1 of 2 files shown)


sil2100 (changes requested):
+1 on the approach! This is a good start for sure. There's some issues I see before we can merge it.

I have experimented with this branch locally, building a few packages here and there. For most it works flawlessly, but I'm getting a crash whenever there's a package that has subpackage `git-checkout` stanzas. Example: wolfi unbound package. Getting a:

```
panic: value method chainguard.dev/melange/pkg/sbom.Package.ID called using nil *Package pointer
```

Looking at the code, I'm not sure currently the upstream source fetching from subpackage code fetches/git-checkouts are handled. I see `GenerateSBOM` doing `SBOMPackageForUpstreamSource()` for the main pipeline, but I don't think that handles subpackage pipelines?

wlynch:
Done! (I think this was mostly a bad rebase from https://github.com/chainguard-dev/melange/pull/2171)

pkg/build/sbom/spdx/spdx.go
```diff
@@ -0,0 +1,235 @@
...
+		if err != nil {
+			return fmt.Errorf("creating SBOM package for upstream source: %w", err)
+		}
+
+		if upstreamPkg == nil {
+			// This particular pipeline step doesn't tell us about the upstream source code.
+			continue
+		}
+
+		pSBOM.AddUpstreamSourcePackage(upstreamPkg)
```
sil2100, line 185: I think this right now causes a regression. For subpackages, right now this basically adds the main package upstream source only to the main package SBOM (`pSBOM`). Up until now, and I like that, we added the main/origin upstream source to all subpackages + the subpackage source next to it. A good example is `unbound`. I guess this suggestion might fix it, not sure if that's the cleanest. Tests will have to be adjusted as well.

```suggestion
		sg.AddUpstreamSourcePackage(upstreamPkg)
```

wlynch: I'm surprised the existing tests didn't catch this. Done!

sil2100:
@wlynch might be that a git push has been missed? I don't think I see any new changes pushed since my last review! Or am I just confused?

wlynch:
whoops! check again :shipit:

sil2100 (approved):

PR #2221, opened 2025-11-07
feat: record git cherry-picks in SBOMs

sil2100:
Currently cherry-picks, similar to patches, go unnoticed in our apk SBOMs. But unlike patches, which are hard to describe in SBOMs as their contents can be custom, cherry-picks come from upstream and can be registered. There's multiple ways to do it, but here I proposed recording those in `externalRefs` for the source. SPDX 2.3 supports an arbitrary `"OTHER"` category which I propose to use here for this. There is also `"SECURITY"`, which is nice as it is something that tools already know how to parse, but it has some shortcomings: first, it's purpose is for security fixes. Second, the locator needs to be a direct url - which is nice, but redundant and might be a bit arbitrary.

Thoughts?

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -550,9 +550,8 @@ func SHA256(text string) string {
 // Each cherry-pick is represented as an ExternalRef with the locator being the
 // branch/commit format
 func parseCherryPicksToExternalRefs(cherryPicks string) []spdx.ExternalRef {
-	var refs []spdx.ExternalRef
-
 	lines := strings.Split(cherryPicks, "\n")
+	refs := make([]spdx.ExternalRef, 0, len(lines))
 	for _, line := range lines {
 		line = strings.Split(line, "#")[0]
 		line = strings.TrimSpace(line)
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -916,8 +916,8 @@ func TestParseCherryPicksToExternalRefs(t *testing.T) {
 			expected: 0,
 		},
 		{
-			name: "single cherry-pick",
-			input: "3.10/c62c9e518b784fe44432a3f4fc265fb95b651906: CVE-2024-4032",
+			name:     "single cherry-pick",
+			input:    "3.10/c62c9e518b784fe44432a3f4fc265fb95b651906: CVE-2024-4032",
 			expected: 1,
 			checkRef: func(t *testing.T, refs []spdx.ExternalRef) {
 				require.Equal(t, "OTHER", refs[0].Category)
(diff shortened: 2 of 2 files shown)


pkg/config/config.go
```diff
@@ -543,10 +544,52 @@ func SHA256(text string) string {
...
+// parseCherryPicksToExternalRefs parses cherry-pick entries and converts them
+// to SPDX ExternalRef entries with category "OTHER" and type "cherry-pick".
+// The format expected is: [branch/]commit-id: comment
+// Each cherry-pick is represented as an ExternalRef with the locator being the
+// branch/commit format
+func parseCherryPicksToExternalRefs(cherryPicks string) []spdx.ExternalRef {
+	lines := strings.Split(cherryPicks, "\n")
+	refs := make([]spdx.ExternalRef, 0, len(lines))
+	for _, line := range lines {
+		line = strings.Split(line, "#")[0]
```
sergiodj, line 556: Not sure I understand this split.  Are you looking for comments here?

pkg/config/config.go
```diff
@@ -543,10 +544,52 @@ func SHA256(text string) string {
...
+		line = strings.TrimSpace(line)
+		if line == "" {
```
sergiodj, line 558:
```suggestion
		if line == "" || strings.HasPrefix(line, "#") {
```

IIUC we can also have comments in the cherry-pick list, which should be ignored as well.

sergiodj (changes requested):
Thanks, @sil2100!  Leaving a small nit for now; I haven't had the chance to do an in-depth review yet.

pkg/config/config.go
sil2100: My thinking was: certain lines can have inline comments, right? Like in `master/b67d508a93bf1ba231c18dce3894cfee25c16e0d: some description  # This is temporary`. If we don't support this, then I guess we can drop it.

pkg/config/config.go
sil2100: So this should be already handled by the above `line = strings.Split(line, "#")[0]` line. If a line starts with a `#`, then the first split will end up the first result (left side from the `#` comment) to be empty (whitespaces will be eliminated with `TrimSpace`), so the line will be skipped (due to being `== ""`). So your suggestion shouldn't be necessary! And, as mentioned above, it also handles the inline comments. Well, depends if inline comments are even a thing...

sergiodj: Ah, gotcha, that makes sense.

pkg/config/config.go
sergiodj: That's a good question.  Looking at the definition of `cherry-picks`:

https://github.com/chainguard-dev/melange/blob/7e104169ebc8bc825a330263c700a46472e21da7/pkg/build/pipelines/git-checkout.yaml#L37-L53

It indeed says any content on a line after `#` is ignored.  Which means you're correct and we should also ignore this content when composing the SBOM.

sergiodj:
> @sergiodj sent some replies to your comments. Tell me what you think! Would you have some time to review this in more depth?

Thanks, @sil2100.  I just had those two questions that you've already answered.  I sat down and looked at the code more carefully, and it seems great.  Thanks a lot for doing this in such short notice!  You rock.

PR #2223, opened 2025-11-08
build: always perform build

xnox:
isBuildLess check is incorrect, it produces empty packages
successfully, if there is no main pipeline, and all computation is
performed in sub-packages pipelines.

This is dangerous, as the build is "successful" when it should not be.

Also linters are skipped and not executed, when they might be needed.


xnox:
Diff is best viewed in the terminal, ignoring whitespace changes.

EyeCantCU (approved):

PR #2237, opened 2025-11-19
sca: scan linux kernels for version strings 

a-crate:
Linux kernels will provide linux:version. In the future modules can be scanned
for the version they were built with using vermagic and depend on this.

Ref VMS-190


sergiodj (commented):
I'm a bit concerned about the size of the SCA file now with these extra helper functions . What do you think about putting them in another file and import it instead?

sergiodj (approved):
Thanks, @a-crate.

LGTM.  There are a few places where we could potentially deduplicate code, but overall I like the organization and I'm OK with this.  I read through the functions and understood everything they're doing, which I think is a good sign!

+1 from my side.

PR #2239, opened 2025-11-21
linter:tempdir: Consider directories too

cjwatson:
For the project to migrate `/var/run` to `/run`, it's inconvenient that
the `tempdir` linter ignores directories under `/var/run`.  Make this
stricter.

### Linter

Notes:

Related to https://docs.google.com/document/d/1pydotE6GRdgWP9xmO9YFFX6_f7PdMa7wMWRQSzH8bFw.  At the moment this does introduce some errors in Wolfi, which will need to be either fixed or ignored before we can land this.

### DIFF
--- a/pkg/linter/linter.go
+++ b/pkg/linter/linter.go
@@ -101,7 +101,7 @@ var linterMap = map[string]linter{
 	},
 	"tempdir": {
 		LinterFunc:      linters.TempDirLinter,
-		Explain:         "Remove any offending files in temporary dirs in the pipeline",
+		Explain:         "Remove any offending files in temporary dirs in the pipeline, or add a /usr/lib/tmpfiles.d/ snippet in the case of dirs under /run/",
 		defaultBehavior: Require,
 	},
 	"usrlocal": {
--- a/pkg/linter/linters/tempdir.go
+++ b/pkg/linter/linters/tempdir.go
@@ -24,14 +24,43 @@ import (
 )
 
 func TempDirLinter(ctx context.Context, _ *config.Configuration, pkgname string, fsys fs.FS) error {
+	hasTmpFilesSnippet := false
+	err := fs.WalkDir(fsys, ".", func(path string, d fs.DirEntry, err error) error {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+		if err != nil {
+			return err
+		}
+		if strings.HasPrefix(path, "usr/lib/tmpfiles.d/") {
+			hasTmpFilesSnippet = true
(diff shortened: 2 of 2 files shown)


cjwatson:
Also if I've missed any packages then please let me know as I'm fairly new to using apk-db.  I noticed that there were some old version streams of packages listed by apk-db that have had their corresponding YAML files garbage-collected from Wolfi/enterprise-packages; I assume we just need to skip those ones.

a-crate:
Wrong button, sorry.

cjwatson:
I believe this should now be ready to go.

pkg/linter/linter.go
```diff
@@ -101,7 +101,7 @@ var linterMap = map[string]linter{
 	},
 	"tempdir": {
 		LinterFunc:      linters.TempDirLinter,
-		Explain:         "Remove any offending files in temporary dirs in the pipeline",
+		Explain:         "Remove any offending files in temporary dirs in the pipeline, or add a /usr/lib/tmpfiles.d/ snippet",
```
a-crate: I think this error message is not really clear to an end user, if I put a directory in `var/run/something` I'll get an error saying 'add a /usr/lib/tmpfiles.d/ snippet,' but that won't actually fix anything. We should be clear that that's only allowed for dirs under `/run`

pkg/linter/linters/tempdir.go
```diff
@@ -18,13 +18,46 @@ import (
...
+	})
+	if err != nil {
+		return err
+	}
+
 	return AllPaths(ctx, pkgname, fsys,
-		func(path string) bool { return IsTempDirRegex.MatchString(path) },
+		func(path string, d fs.DirEntry) bool {
+			// Anything in /tmp/, /var/tmp/, or /var/run/ is an error.
+			if IsTempDirRegex.MatchString(path) || strings.HasPrefix(path, "var/run/") {
```
a-crate: We should maintain the old (permitting directories) behavior for `^(var/)?tmp/` since we're not going to check for or fix any package with files there: `if !d.IsDir() && IsTempDirRegex.MatchString(path)`.

pkg/linter/linter.go
cjwatson: OK, I tried to clarify this.

pkg/linter/linters/tempdir.go
cjwatson: Done.

a-crate (approved):
lgtm once tempdir exceptions are merged

PR #2245, opened 2025-11-28
fix(sca): handle broken shebang properly

wojciechka:


cpanato (approved):
seems ok, maybe @xnox also can take a look?

xnox (commented):
Do you have explicit examples where this has happened?

Normally executable file with an empty "#!" would fail to execute by the kernel. Thus this should not happen, and maybe we should emit a warning or even fail the build?

I wonder if this happend due to unsuccessful string substitution, rather than intentionally.

Unless it is some file that starts with such a comment, and it is never intended to be executed - but then why would it be executable.

PR #2249, opened 2025-12-02
sca: scan linux modules for vermagic

a-crate:
Module counterpart to #2237, scan the modules and depend on the version string.

Fixes VMS-190

### DIFF
--- a/pkg/sca/kernel_sca.go
+++ b/pkg/sca/kernel_sca.go
@@ -172,6 +172,7 @@ func tryDecompressModule(ctx context.Context, r io.ReadSeeker) (io.ReadSeeker, e
 	// gzip header
 	gzr, err := gzip.NewReader(r)
 	if err == nil {
+		defer gzr.Close()
 		b, err := io.ReadAll(gzr)
 		if err != nil {
 			return nil, err


pkg/sca/kernel_sca.go
```diff
@@ -102,6 +126,91 @@ func generateKernelDeps(ctx context.Context, hdl SCAHandle, generated *config.De
...
+	}()
+
+	if _, err := r.Seek(0, io.SeekStart); err != nil {
+		return nil, err
+	}
+	// Reader header on NewReader() and returns err if it's not a valid
+	// gzip header
+	gzr, err := gzip.NewReader(r)
+	if err == nil {
+		b, err := io.ReadAll(gzr)
```
sergiodj, line 176:
```suggestion
		defer gzr.Close()
		b, err := io.ReadAll(gzr)
```

pkg/sca/kernel_sca.go
```diff
@@ -102,6 +126,91 @@ func generateKernelDeps(ctx context.Context, hdl SCAHandle, generated *config.De
...
+		log.Debugf("found gzipped module")
+		return bytes.NewReader(b), nil
+	}
+
+	if _, err := r.Seek(0, io.SeekStart); err != nil {
+		return nil, err
+	}
+	xzr, err := xz.NewReader(r)
+	if err == nil {
+		b, err := io.ReadAll(xzr)
```
sergiodj, line 189: As far as I've checked the `xz` Reader [doesn't have a `Close` method](https://pkg.go.dev/github.com/ulikunitz/xz#Reader), so we should be OK here.

pkg/sca/kernel_sca.go
```diff
@@ -28,20 +28,24 @@ import (
...
 // well for bootc, which suggests sticking kernel images at
-// /usr/lib/modules/`uname -r`/vmlinuz
+// /usr/lib/modules/`uname -r`/vmlinuz.
 var BootDirs = []string{"boot/...", "lib/modules/...", "usr/lib/modules/..."}
 
-// TODO: a-crate: scan kernel modules for vermagic and depend on linux:version
-// TODO: a-crate: maybe also parse modules.dep for module dependencies
+// ModuleDirs is the list of directories to search for kernel modules.
+// This is exported so that callers can append to it as needed.
+var ModuleDirs = []string{"usr/lib/modules/...", "lib/modules/..."}
```
sergiodj, line 47: I should have asked this before, but just out of curiosity: what happens with the `...` at the end of each path?  Does `isInDir` disconsider them?

sergiodj (changes requested):
Thanks, @a-crate.

Leaving some comments and a suggestion, but otherwise LGTM.

pkg/sca/kernel_sca.go
a-crate: Yea, a bit odd IMO but there's no `Close`.

pkg/sca/kernel_sca.go
a-crate: `...` means `isInDir` should also consider sub-directories rather than just direct children.

sergiodj: I may be wrong here, but I don't think this works...?  Here's the implementation of `isInDirs`:

```go
func isInDir(path string, dirs []string) bool {
	mydir := filepath.Dir(path)
	for _, d := range dirs {
		if mydir == d || mydir+"/" == d {
			return true
		}
	}
	return false
}
```

We're just doing a regular string comparison here, not a path expansion.  Or am I missing something?

a-crate: The version committed to `main` is different from that code: https://github.com/chainguard-dev/melange/blob/main/pkg/sca/sca.go#L102

sergiodj: D'oh.  Sorry about the noise.

PR #2250, opened 2025-12-02
Refactor Build to generate build options from flag struct.

wlynch:
Also Rebuild and Test.

This change groups build option creation logic so any logic related to flag parsing / manipulation can be consolidated together consistently.

### DIFF
--- a/pkg/cli/build.go
+++ b/pkg/cli/build.go
@@ -154,6 +154,111 @@ func ParseBuildFlags(args []string) (*BuildFlags, []string, error) {
 	return flags, fs.Args(), nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/cli/build.go
```diff
@@ -28,6 +28,7 @@ import (
 	apko_types "chainguard.dev/apko/pkg/build/types"
 	"github.com/chainguard-dev/clog"
 	"github.com/go-git/go-git/v5"
+	"github.com/google/go-containerregistry/cmd/crane/cmd"
```
jonjohnsonjr: lol omg

wlynch: Fixed 😅

PR #2264, opened 2025-12-09
feat: add TESTING environment variable feature gate for microvm-init

antitree:
## Summary

Add support for `TESTING` environment variable that gets passed to the kernel command line as `melange.testing=<value>`. This allows us to feature within staging to test things. In particular, we'd like to make some changes to  microvm-init. 

## Changes

- Modified `pkg/container/qemu_runner.go` to read `TESTING` environment variable
- Validates that `TESTING` is a number (or empty string) to prevent injection issues
- Passes the value to kernel command line as `melange.testing=<value>`
- Adds debug/warning logging for the feature gate

## Usage

```bash
TESTING=1 melange build mypackage.yaml
```

## Security

The implementation validates that `TESTING` only accepts numeric values (or empty string) to prevent kernel command line injection attacks. String values are rejected with a warning message.

## Test plan

- [x] Set `TESTING=1` and verify `melange.testing=1` appears in kernel cmdline
- [x] Set `TESTING=foo` and verify warning is logged and value is not passed
- [x] Set `TESTING=` and verify `melange.testing=` is passed to kernel
- [x] Run without `TESTING` env var and verify no changes to existing behavior

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -722,13 +722,10 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 	kernelArgs := kernelConsole + " nomodeset random.trust_cpu=on panic=-1 sshkey=" + sshkey + " melange_qemu_runner=1"
 
 	// Check for TESTING environment variable and pass it to microvm-init
-	// TESTING must be a number (or empty string)
+	// TESTING must be a number (0 for disabled, non-zero for enabled)
 	if testingValue, ok := os.LookupEnv("TESTING"); ok {
-		if testingValue == "" {
-			log.Debugf("qemu: TESTING env set to empty string, passing to microvm-init via kernel cmdline")
-			kernelArgs += " melange.testing="
-		} else if _, err := strconv.Atoi(testingValue); err == nil {
-			log.Debugf("qemu: TESTING env set to %s, passing to microvm-init via kernel cmdline", testingValue)
+		if _, err := strconv.Atoi(testingValue); err == nil {
+			log.Infof("qemu: TESTING env set to %s, passing to microvm-init via kernel cmdline", testingValue)
 			kernelArgs += " melange.testing=" + testingValue
 		} else {
 			log.Warnf("qemu: TESTING env must be a number, ignoring invalid value: %s", testingValue)


pkg/container/qemu_runner.go
```diff
@@ -717,7 +717,25 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
...
+	// Build kernel command line arguments
+	kernelArgs := kernelConsole + " nomodeset random.trust_cpu=on panic=-1 sshkey=" + sshkey + " melange_qemu_runner=1"
+
+	// Check for TESTING environment variable and pass it to microvm-init
+	// TESTING must be a number (or empty string)
+	if testingValue, ok := os.LookupEnv("TESTING"); ok {
+		if testingValue == "" {
+			log.Debugf("qemu: TESTING env set to empty string, passing to microvm-init via kernel cmdline")
+			kernelArgs += " melange.testing="
+		} else if _, err := strconv.Atoi(testingValue); err == nil {
```
egibs: Do we want to make sure the converted value is either `0` or `1` or is it fine to pass through any `int`?

e.g., `else if s, err := strconv.Atoi(testingValue); err == nil && (s == 0 || s == 1) { ...`

antitree: It was at least intended to handle any int so TESTING=1 or TESTING=9 is possible but you can tell me I shouldn't and happy to rip it out. In practice blank, 0, and 1 are all we need today. LMK

egibs: I think this is fine. I guess the primary check elsewhere will be whether the value is zero or not so the non-zero value doesn't really matter much.

pkg/container/qemu_runner.go
```diff
@@ -717,7 +717,25 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
...
+
+	// Build kernel command line arguments
+	kernelArgs := kernelConsole + " nomodeset random.trust_cpu=on panic=-1 sshkey=" + sshkey + " melange_qemu_runner=1"
+
+	// Check for TESTING environment variable and pass it to microvm-init
+	// TESTING must be a number (or empty string)
+	if testingValue, ok := os.LookupEnv("TESTING"); ok {
+		if testingValue == "" {
+			log.Debugf("qemu: TESTING env set to empty string, passing to microvm-init via kernel cmdline")
+			kernelArgs += " melange.testing="
```
stevebeattie: What do you expect to use the empty space setting for? Is that expected to map to/be treated by microvm-init as zero or non-zero?

Personally, I'd stick with environment variable unset, 0, and non-zero numeric as possible values. Trying to keep the logic around this as simple as possible is probably for the best.

(I'm trying hard not to bikeshed on the variable name(s).)

antitree: This was more input validation than a valid use case but yes I can reduce this down to set number or not set

pkg/container/qemu_runner.go
```diff
@@ -717,7 +717,25 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
...
+		} else if _, err := strconv.Atoi(testingValue); err == nil {
+			log.Debugf("qemu: TESTING env set to %s, passing to microvm-init via kernel cmdline", testingValue)
```
stevebeattie: I'm not **adamant** about this but probably we want this at an informational level, rather than debug, given that it's an input into the build environment, and it makes it easier to capture than running `--log-level=debug`. 

(OTOH, if we start to capture build logs then the dmesg/syslog/journal would contain the actual passed command line flags, so it might be redundant. But we don't do that yet.)

stevebeattie:
Modulo the comments, this looks good and works as expected. Thanks!

pkg/container/qemu_runner.go
antitree: fixed

PR #2266, opened 2025-12-10
feat: add QEMU_ADDITIONAL_PACKAGES environment variable

antitree:
## Summary

This PR adds support for the `QEMU_ADDITIONAL_PACKAGES` environment variable, which allows users to specify additional packages to **include in the initramfs image** during QEMU VM initialization. This complements the existing `TESTING` environment variable feature gate.

## Changes

- Added `QEMU_ADDITIONAL_PACKAGES` support in `pkg/container/qemu_runner.go`
- Packages are added to the initramfs image build specification
- Accepts comma-separated list of package names (e.g., `hello-wolfi,nginx-stable,strace`)
- Packages are baked into the initramfs for immediate availability at boot
- Input validation prevents injection attacks (regex: `^[a-zA-Z0-9_,.-]+$`)
- Cache invalidation: different package lists create separate cached initramfs files

## How It Works

**Without QEMU_ADDITIONAL_PACKAGES:**
```go
spec := apko_types.ImageConfiguration{
    Contents: apko_types.ImageContents{
        Packages: []string{"microvm-init"},
    },
}
```

**With QEMU_ADDITIONAL_PACKAGES=hello-wolfi,strace:**
```go
spec := apko_types.ImageConfiguration{
    Contents: apko_types.ImageContents{
        Packages: []string{"microvm-init", "hello-wolfi", "strace"},
    },
}
```

Packages are installed into the initramfs during the apko build, so they're available immediately when the VM boots - no runtime `apk add` needed!

## Usage

```bash
# Single package
QEMU_ADDITIONAL_PACKAGES=hello-wolfi melange build mypackage.yaml --runner qemu

# Multiple packages
QEMU_ADDITIONAL_PACKAGES=strace,gdb,tcpdump melange build mypackage.yaml --runner qemu
```

## Use Cases

- **Testing/debugging**: Install tools like `strace`, `gdb`, `tcpdump` for debugging builds
- **Runtime dependencies**: Install packages needed during the build that aren't in the build environment  
- **Development**: Quick iteration without modifying build configurations\
- **Init.d hooks**: Install packages that provide init.d scripts for custom initialization

## Security

The implementation validates input to prevent shell injection attacks. Only alphanumeric characters, hyphens, underscores, commas, and dots are allowed in package names. Suspicious input is rejected with a warning.

## Test Results

Verified that packages are successfully added to the initramfs:
```
2025/12/10 12:05:59 INFO qemu: QEMU_ADDITIONAL_PACKAGES env set to hello-wolfi, adding to initramfs
2025/12/10 12:06:00 INFO     packages:     [microvm-init hello-wolfi]
2025/12/10 12:06:02 INFO installing hello-wolfi (2.12.2-r2)
```

The package is installed in the initramfs and available at boot time ✅

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -1795,9 +1795,6 @@ func generateBaseInitramfs(ctx context.Context, cfg *Config, initramfsPath, cach
 	if err != nil {
 		return fmt.Errorf("unable to create dest directory: %w", err)
 	}
-	
-	// Get existing additional packages from context
-	additionalPkgs := getAdditionalPackages(ctx)
 
 	// Start with base packages and add any additional packages from environment
 	packages := []string{"microvm-init"}


pkg/container/qemu_runner.go
```diff
@@ -1784,14 +1782,34 @@ func generateBaseInitramfs(ctx context.Context, cfg *Config, initramfsPath, cach
 		return fmt.Errorf("unable to create dest directory: %w", err)
 	}
 
+	// Start with base packages
+	packages := []string{"microvm-init"}
+
+	// Check for QEMU_ADDITIONAL_PACKAGES environment variable
+	// Add packages to the initramfs image so they're available during boot
+	if additionalPkgs, ok := os.LookupEnv("QEMU_ADDITIONAL_PACKAGES"); ok && additionalPkgs != "" {
```
egibs: Would it be worth moving this to a new function (`getAdditionalPackages`) and writing a test to make sure this does what it's supposed to?

egibs: (and likely the new `TESTING` code as well although it's not doing nearly as much)

pkg/container/qemu_runner.go
```diff
@@ -1756,9 +1742,21 @@ func generateCpio(ctx context.Context, cfg *Config) (string, error) {
...
 	cacheDir = filepath.Join(cacheDir, "melange-cpio")
 
+	// Include additional packages in cache filename to invalidate cache when they change
+	additionalPkgs := os.Getenv("QEMU_ADDITIONAL_PACKAGES")
+	cacheSuffix := ""
+	if additionalPkgs != "" {
+		// Use sanitized package list as cache key (replace commas and dots with dashes)
+		sanitized := strings.NewReplacer(",", "-", ".", "-").Replace(additionalPkgs)
+		if len(sanitized) > 32 {
+			sanitized = sanitized[:32] // Limit length for reasonable filenames
```
egibs: Can we foresee hitting the edge case where we install enough packages to make this consistent even though we're installing new/different packages after the cutoff?

Would something like a checksum work better?

pkg/container/qemu_runner.go
89luca89: Agree on the separate function and tests
Also I'd avoid shadowing variables in general

pkg/container/qemu_runner.go
89luca89: Agree on checksum, makes it simpler and less likely to have strange chars in it

antitree: Yeah I think that smart and gets us out of dealing with the 4K cmdline problem I think

pkg/container/qemu_runner.go
antitree: On it. Added a new function and tests for it

pkg/container/qemu_runner.go
```diff
@@ -1790,14 +1796,16 @@ func generateBaseInitramfs(ctx context.Context, cfg *Config, initramfsPath, cach
 		return fmt.Errorf("unable to create dest directory: %w", err)
 	}
 
+	// Start with base packages and add any additional packages from environment
+	packages := []string{"microvm-init"}
+	packages = append(packages, getAdditionalPackages(ctx)...)
```
egibs:
```suggestion
	packages = append(packages, additionalPkgs...)
```

89luca89: Seems to me the only thing I'd change 👍

antitree (commented):
factored in @egibs change

pkg/container/qemu_runner.go
```diff
@@ -1790,14 +1796,16 @@ func generateBaseInitramfs(ctx context.Context, cfg *Config, initramfsPath, cach
 		return fmt.Errorf("unable to create dest directory: %w", err)
 	}
 
+	// Start with base packages and add any additional packages from environment
+	packages := []string{"microvm-init"}
+	packages = append(packages, additionalPkgs...)
```
egibs, line 1801: Ah, we need to pass this in on L1791 via L1780. The diff made this look like the same function. 😅

89luca89: lol yea it looked like the same function, it was working correctly before

PR #2268, opened 2025-12-14
pipeline: fetch: populate cache

dannf:
Studies show that a good way to improve cache hit %'s is to actually put things in the cache.

Also: misc code improvements - see individual commits.


nissessenap:
I ran melange localy and it works as expected. 
This will save me lots of time <3, lets hope it will get merged as well.

PR #2278, opened 2025-12-16
linter: Added new fetch/templating linter

AbduAzim424:
This PR introduces a new linter, `fetch/templating`, which validates that package `fetch` sources and `git checkouts` use proper version templating and reproducibility safeguards. It is designed to prevent version drift, silent breakage and non-reproducible builds, including the class of issues that previously allowed outdated sudo sources to go unnoticed.

It operates directly on the raw YAML (pre-template expansion), so it sees the actual `uri`, `tag` and `ref` strings as written in the recipe. It enforces a small set of rules:
- Rule `A`: At least one source must use templates (e.g., ${{package.version}})
- Rule `B`: Fetch URLs with hardcoded package versions should use version templates
- Rule `C`: Git tags with hardcoded versions should use version templates
- Rule `D`: Git branches/refs require expected-commit for reproducibility

- Examples:
```
2025/12/16 00:14:30 WARN [fetch/templating] source lacks templated variables: fetch URL: https://www.sudo.ws/dist/sudo-1.9.17p1.tar.gz
2025/12/16 00:14:30 WARN  Consider using ${{package.version}} to ensure URL updates with version changes
2025/12/16 00:14:30 WARN  → At least one fetch URL should use version-related template variables (like ${{package.version}}) to ensure source URLs are updated when package versions change
```

```
2025/12/16 00:15:35 WARN [fetch/templating] source lacks templated variables: git tag: v9.1.1868
2025/12/16 00:15:35 WARN   Consider using ${{package.version}} to ensure URL updates with version changes
2025/12/16 00:15:35 WARN   → At least one fetch URL should use version-related template variables (like ${{package.version}}) to ensure source URLs are updated when package versions change
```

### DIFF
--- a/pkg/linter/linters/fetch.go
+++ b/pkg/linter/linters/fetch.go
@@ -1,0 +1,401 @@
+// Copyright 2025 Chainguard, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package linters
+
+import (
+	"context"
+	"fmt"
+	"io/fs"
+	"regexp"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+
+	"chainguard.dev/melange/pkg/config"
+)
+
+/*
+Rules:
+- Rule A: At least one source must use templates (e.g., ${{package.version}})
+- Rule B: Fetch URLs with hardcoded package versions should use version templates
+- Rule C: Git tags with hardcoded versions should use version templates
+- Rule D: Git branches/refs require expected-commit for reproducibility
+
+Supports version transforms: ${{package.version | replace: ".", "_"}} and ${{vars.version}}
+*/
+
+// Holds git checkout reference information for Rule D validation
+type gitRefInfo struct {
+	Branch         string
+	Ref            string
+	Tag            string
+	ExpectedCommit string
+}
+
+// Holds compiled regex patterns for a specific package
+type packagePatterns struct {
+	exactVersionURL   *regexp.Regexp
+	anyVersionURL     *regexp.Regexp
+	exactVersionTag   *regexp.Regexp
+	packageVersionTag *regexp.Regexp
+}
+
+// Holds extracted pipeline data
+type sourceData struct {
+	fetchURLs   []string
+	gitTags     []string
+	gitBranches []gitRefInfo
+}
+
+// Handles validation logic with pre-compiled patterns
+type validator struct {
+	pkg      packageInfo
+	patterns *packagePatterns
+}
+
+// Holds package metadata
+type packageInfo struct {
+	name    string
+	version string
+}
+
+var (
+	anyTemplateRegex     = regexp.MustCompile(`\$\{\{[^}]+\}\}`)
+	versionTemplateRegex = regexp.MustCompile(`\$\{\{\s*(package\.(version|full-version)|vars\.[^}]*\bversion\b[^}]*)[^}]*\}\}`)
+	shaRegex             = regexp.MustCompile(`^[0-9a-fA-F]{12,40}$`)
+)
+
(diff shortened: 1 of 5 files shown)


pkg/linter/linters/fetch.go
```diff
@@ -0,0 +1,401 @@
...
+	"strings"
+
+	"gopkg.in/yaml.v3"
+
+	"chainguard.dev/melange/pkg/config"
+)
+
+/*
+Rules:
+- Rule A: At least one source must use templates (e.g., ${{package.version}})
```
egibs, line 31: Should this warn if there are multiple sources and only one source uses templating in addition to failing if none of the sources use templating?

pkg/linter/linters/fetch.go
```diff
@@ -0,0 +1,401 @@
...
+	if packageName == "" || packageVersion == "" {
+		return nil
+	}
+
+	escName := regexp.QuoteMeta(packageName)
+	escVer := regexp.QuoteMeta(packageVersion)
+
+	return &packagePatterns{
+		exactVersionURL: regexp.MustCompile(
+			`(?:^|/)` + escName +
```
egibs, line 95: Should these patterns be defined as variables for easier reference? Would it make sense to also fuzz/test each to ensure sure they do or do not match what we'd expect?

pkg/linter/linters/fetch.go
```diff
@@ -0,0 +1,401 @@
...
+
+	"chainguard.dev/melange/pkg/config"
+)
+
+/*
+Rules:
+- Rule A: At least one source must use templates (e.g., ${{package.version}})
+- Rule B: Fetch URLs with hardcoded package versions should use version templates
+- Rule C: Git tags with hardcoded versions should use version templates
+- Rule D: Git branches/refs require expected-commit for reproducibility
```
egibs, line 34: IMO, we should require `expected-sha256` or an equivalent for fetches (unless this already exists somewhere and I'm misremembering).

pkg/linter/linters/fetch.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+
+	getString := func(mapping *yaml.Node, key string) string {
+		if node := findValue(mapping, key); node != nil && node.Kind == yaml.ScalarNode {
+			return node.Value
+		}
+		return ""
+	}
+
+	var processPipelines func(*yaml.Node)
```
egibs, line 145: Would it make sense to have `findValue`, `getString`, and `processPipelines` exist as separate functions for maintainability and to improve future testing?

pkg/linter/linters/fetch.go
AbduAzim424: Yes it's tricky. I was basing this off of `One challenge is that a package can have multiple fetch / origins, not all of which would be parameterized on the package version or other template variable, so it would need to be something like "flag if none of the fetch locations include a variable` comment

pkg/linter/linters/fetch.go
```diff
@@ -0,0 +1,401 @@
...
+	}
+
+	validator := newValidator(pkg)
+	return validator.validateAll(sources)
+}
+
+// Runs all validation rules and returns formatted errors
+func (v *validator) validateAll(sources sourceData) error {
+	// Rule A: Template requirement
+	if err := v.validateRuleA(sources); err != nil {
```
justinvreeland, line 240: why is A different than the other issues?

justinvreeland (commented):
I'd like to see the results of this run on all the files we have now.

I have some concerns about files that should be fixed up that seem like they'd not work with this. `gn.yaml` and `jenkins-2.yaml` in wolfi should ideally use git checkout. `fluent-plugin-newrelic.yaml` will fail this I think but that might be a poorly formed package in general. I think the matching might be a bit too strict. We have `ingress-nginx-controller-1.13.yaml` and `opentelemetry-cpp-contrib.yaml` which define their own variables to old the versions in that should pass this ideally but i think will fail; though I guess they're already opting out of a number of git related pipelines.

It'd be nice to get a list of all the current warnings and see if they can be fixed or pre-emptively add exceptions so this can be turned on as more than a warning.

I'm not sure about the placement of the linter here. IIRC melange linters are mostly linting packages not the melange files themselves. I thought there was work or plans to begin unifying them but looking at this I'm pretty sure that the build configuration you're being passed here as config is the fully processed build configuration struct not the raw yaml from the file.

justinvreeland:
> You do call the same function to parse the configuration but it does look like it should be performing variable substituion:
> 
> https://github.com/chainguard-dev/melange/blob/82f1210fcf7bd0dbf3f327a3474452334b80c3d1/pkg/config/config.go#L1529

Missed that cfg.Root preserves the original tree.

PR #2320, opened 2026-01-30
Revert "fix: add and leverage helper functions for [safely] setting c…

jonjohnsonjr:
…apacity (#2318)"

This reverts commit ae3c12f0267c6ea5a37fa6a8f367d630e09b88d1.

This package adds unnecessary noise around a very common operation.

If a melange program actually has anywhere close to `math.MaxInt` values of anything, we're never going to actually finish executing it, or we'll run out of memory pretty quickly.

Go is a memory safe language. If you attempt to allocate a slice with negative capacity, it will panic, which is the only reasonable thing to do. Attempting to "safely" allocate, instead, `math.MaxInt` of anything just delays the problem very slightly. If we actually need that capacity, we're not going to be able to execute the program.


egibs:
We discussed this offline and I agree entirely. The source of the original finding has been dismissed to avoid over-engineered solutions in the future.

PR #2323, opened 2026-02-02
feat: add --pipeline-dirs flag to build command

debasishbsws:
Adds --pipeline-dirs flag to accept multiple pipeline directories, matching the test command functionality. The existing --pipeline-dir (singular) is kept for backward compatibility.

Directories are searched in order with first match winning. For example, if both custom1/fetch.yaml and custom2/fetch.yaml exist, custom1 wins:

  `melange build --pipeline-dirs custom1 --pipeline-dirs custom2 --pipeline-dir original config.yaml`
  
  **Searches: `original/` -> `custom1/` -> `custom2/` -> `/usr/share/melange/pipelines/`**

This enables layered pipeline overrides for teams with shared repositories.

Search order:
1. --pipeline-dir value (if specified)| (for backward compatibility)
2. --pipeline-dirs values (left to right)
3. Built-in pipeline directory

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

Reason for this change is explained : https://github.com/chainguard-dev/stereo/pull/15448

This also make the `melange test` and `melaneg build` command flags more consistent with each other

### DIFF
--- a/docs/md/melange_build.md
+++ b/docs/md/melange_build.md
@@ -62,6 +62,7 @@ melange build [flags]
       --package-append strings                                  extra packages to install for each of the build environments
       --persist-lint-results                                    persist lint results to JSON files in packages/{arch}/ directory
       --pipeline-dir string                                     directory used to extend defined built-in pipelines
+      --pipeline-dirs strings                                   directories used to extend defined built-in pipelines
   -r, --repository-append strings                               path to extra repositories to include in the build environment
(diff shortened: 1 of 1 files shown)


kranurag7 (approved):
this looks good to me and is pretty similar to what we've here https://github.com/chainguard-dev/melange/blob/d3bebaf0bb5d91e27c39341d55a0349a2960a1d0/pkg/cli/test.go?plain=1#L38

PR #2326, opened 2026-02-03
better sbom

xnox:
Previously we had: SPDX-Package be described SPDX-Package and generated from SPDX-Package.

With this change we have: SPDX-Packge-**apk** with PrimaryPurpose **Application** described by SPDX-Packge-**Melange** with PrimaryPurpose **Install**, and generated from SPDX-Packge-**Source** with PrimaryPurpose **Source**.

This uses SPDX semantics to better quality the SPDX IDs and their purpose, allow to clearly differentiate melange.yaml build recipes, from the binary apk, from the upstream source code.

The net change is best shown in the last commit, expand in full to see the changed fields:
- https://github.com/chainguard-dev/melange/pull/2326/changes/0e8695aa32b36a536f6f0551467ef42dde0916f8

- **sbom: set PrimaryPurpose APPLICATION on APK SBOM**
  Set PrimaryPurpose APPLICATION on the APK SBOM packages for easier
  identification and filtering.
  

- **sbom: rename APK SBOM packages from Package to Package-apk**
  Rename apk sbom package from Package to Package-apk, such that one can
  identify and filter these SPDX packages.
  

- **sbom: set primary purpose SOURCE on the source references**
  Set primary purpose SOURCE on the source references
  

- **sbom: rename source packages from Package to Package-Source**
  Rename SOURCE packages from Package to Package-Source.
  

- **sbom: rename subpackages to Package-APK**
  Rename subpackages to Package-apk too
  

- **sbom: rename melange yamls to Package-Melange with PrimaryPurpose INSTALL**
  Whilst primary purpose INSTALL is ambigious, it is a close enough
  approximation, as it describes the INSTALL steps done to create
  APPLICATION from SOURCE. Which are the other purposes we use.
  

- **update gold files**

### DIFF
--- a/pkg/build/testdata/goldenfiles/sboms/7zip-two-fetches-2301-r3.spdx.json
+++ b/pkg/build/testdata/goldenfiles/sboms/7zip-two-fetches-2301-r3.spdx.json
@@ -13,7 +13,7 @@
   "dataLicense": "CC0-1.0",
   "documentNamespace": "https://spdx.org/spdxdocs/chainguard/melange/35a84e53c8a7af192be547e9f8c6eac6",
   "documentDescribes": [
-    "SPDXRef-Package-7zip-two-fetches-2301-r3"
+    "SPDXRef-Package-apk-7zip-two-fetches-2301-r3"
   ],
   "packages": [
     {
@@ -30,7 +30,7 @@
       "primaryPackagePurpose": "OPERATING-SYSTEM"
     },
     {
-      "SPDXID": "SPDXRef-Package-7zip-two-fetches-2301-r3",
+      "SPDXID": "SPDXRef-Package-apk-7zip-two-fetches-2301-r3",
       "name": "7zip-two-fetches",
       "versionInfo": "2301-r3",
       "filesAnalyzed": false,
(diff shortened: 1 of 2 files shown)


pkg/build/sbom/spdx/spdx.go
```diff
@@ -146,25 +148,29 @@ func (g *Generator) GenerateSPDX(ctx context.Context, gc *build.GeneratorContext
...
 	if gc.ConfigFile != nil {
 		sg.AddBuildConfigurationPackage(&sbom.Package{
+			IDComponents:    []string{"Melange", gc.ConfigFile.Path, gc.ConfigFile.Commit},
 			Name:            gc.ConfigFile.Path,
 			Version:         gc.ConfigFile.Commit,
 			LicenseDeclared: gc.ConfigFile.License,
 			Namespace:       gc.Namespace,
 			Arch:            "", // This field doesn't make sense in this context
 			PURL:            gc.ConfigFile.PURL,
+			PrimaryPurpose:  "INSTALL",
```
sil2100, line 173: This is what I'm not entirely sure about. SPDX 2.3 says `INSTALL` is "if the package is used to install software on disk", which the melange yaml does not do. It's like the package build recipe. Maybe just `FILE`? Since it's "if the package is a single file which can be independently distributed (configuration file, statically linked binary, Kubernetes deployment, etc)", which, even though still unfitting, sort of feel like it? Shame there's no `METADATA` or something...

xnox: So spdxv3 has more types we can choose from. The v2.3 doesn't have a perfect match. I am trying to say that "SOURCE did INSTALL to create APK" as the pagadim. As our melange.yaml does ultimately do "make install" right?

Issue #2329, opened 2026-02-05
Example fails to build


codysoyland:
When building the example in the readme using the following command:

```
melange build examples/gnu-hello.yaml
```

This fails with the following error:

```
2026/02/04 17:43:50 ERRO failed to build package: unable to build guest: unable to lock image configuration: resolving apk packages: for arch "loong64": error getting repository indexes: reading index https://packages.wolfi.dev/os/loongarch64/APKINDEX.tar.gz: unexpected status code 404
```

The expected file for loongarch64 is not found at `https://packages.wolfi.dev/os/loongarch64/APKINDEX.tar.gz`.

PR #2335, opened 2026-02-05
Limit architectures specified in example

codysoyland:

This limits the architectures specified in the example to just arm64 and x86_64. This speeds up the build and removes arches not supported by wolfi.

Fixes https://github.com/chainguard-dev/melange/issues/2329

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

Notes:

### DIFF
--- a/examples/gnu-hello.yaml
+++ b/examples/gnu-hello.yaml
@@ -3,6 +3,9 @@ package:
   version: 2.12
   epoch: 0
   description: "the GNU hello world program"
+  target-architecture:
+    - x86_64
+    - arm64
   copyright:
     - attestation: |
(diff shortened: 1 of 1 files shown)


codysoyland:
> Not sure I understand the e2e workflow change. What issue are you trying to address with the e2e workflow change?

Sorry, I meant to come back and update this after CI ran -- the intent with the e2e workflow was so that we would catch this error if the architecture section is removed from the example workflow (the workflow sets the architecture explicitly and removing that causes it to break as I observed when trying to run the example manually).

I will actually back out that workflow change as I broke the workflow as it references several other unrelated examples that I would rather not touch and this is unlikely to become a regression.

PR #2338, opened 2026-02-06
Add clang support for package metadata ELF notes

dannf:
This adds Package Metadata ELF notes for binaries built with Clang, like we already have for GCC. We do this by creating user configuration files for clang (`~/.config/clang-*/*.cfg`) to add the necessary linker flag.
    
User config files override system config files. The system config files provide our hardening config, so the user config files we generate include the corresponding system config files.

The warts:
- Generating ~/.config/ in the workspace can potentially conflict with source that has a top level `.config` file, or uses one at build-time, which will fail the build. One of `melange`'s test cases hit that, and I've fixed it up here by moving the build to a subdirectory. There are a couple of other CG OS packages I've found that also do that that I'm preparing fixes for, but I haven't done a thorough investigation.
- clang errors out if its config includes a file that doesn't exist. We can't check in advance if a given system config file exists, so we're counting on the fact that `openssf-compiler-options` is always installed in a melange environment, and provides config files for all of our clang versions.
- Our clang config files are per-clang-major-version[*], and we don't know in advance what clang version(s) will be installed. Here we take a shotgun approach and just install configs for all versions we currently have, and all versions that are likely to exist in the next decade.
 
Alternatives considered:
- Scanning for clang system configs in the workspace, and only creating corresponding config files for those. We don't have a good way to insert files after the image has been created.
- Examining the apko resolution tree for clang packages
- Using a pipeline to populate the config. That'd avoid potential `~/.config` conflicts w/ non-clang packages. But I think we'd prefer this to just be on by default.

[*] That's a build-time config (https://github.com/chainguard-dev/stereo/pull/13823), so it could be changed, but following the system config pattern is what I felt made the most sense for our users.

### DIFF
--- a/e2e-tests/git-checkout-build.yaml
+++ b/e2e-tests/git-checkout-build.yaml
@@ -273,8 +273,8 @@ pipeline:
       find . -user "$runner_user" $exclude_args | sort > /tmp/runner-owned.lst
 
       if [ "${runner_user}" = "${melange_user}" ]; then
-        cat melange-created.lst runner-created.lst | \
-          sort -u > /tmp/expected-melange-owned.lst
+        sort -u melange-created.lst runner-created.lst > \
+          /tmp/expected-melange-owned.lst
         cp /tmp/expected-melange-owned.lst /tmp/expected-runner-owned.lst
       else
         cp melange-created.lst /tmp/expected-melange-owned.lst


pkg/build/compiler_config.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	return nil
+}
+
+func (b *Build) createClangConfigFiles(ctx context.Context) error {
+	log := clog.FromContext(ctx)
+
+	systemConfigs, err := b.detectClangSystemConfigFiles()
+	if err != nil {
```
xnox: just use a fixed list here

pkg/build/compiler_config.go
```diff
@@ -0,0 +1,192 @@
...
+		log.Warnf("failed to detect clang system configs: %v", err)
+		return nil
+	}
+
+	if len(systemConfigs) == 0 {
+		log.Debugf("no clang system config files detected, skipping user config file creation")
```
xnox: no, we always write.

pkg/build/compiler_config.go
dannf: done

e2e-tests/git-checkout-build.yaml
```diff
@@ -267,35 +267,19 @@ pipeline:
...
-          echo "Expected $expected_melange files owned by $melange_user, found $found_melange"
-          find . -user $melange_user $exclude_args
-          mismatch=true
-        fi
+      find . -user "$melange_user" $exclude_args | sort > /tmp/melange-owned.lst
+      find . -user "$runner_user" $exclude_args | sort > /tmp/runner-owned.lst
+
+      if [ "${runner_user}" = "${melange_user}" ]; then
+        cat melange-created.lst runner-created.lst | \
+          sort -u > /tmp/expected-melange-owned.lst
```
sergiodj: Being pedantic here, `sort -u` will accept filenames so there's no need to `cat | sort`.

dannf: fixed!

PR #2345, opened 2026-02-10
feat: optimize QEMU microVM startup time and resource usage

89luca89:
- Disable virtio console and serial output in non-debug builds to reduce boot time and memory overhead
- Add x-option-roms=off to microvm machine type to skip option ROM loading
- Add kernel boot optimizations: tsc=reliable, no_timer_check, cryptomgr.notests, rcupdate.rcu_expedited=1
- Simplify disk drive config to use aio=threads uniformly across platforms

This improves boot time for the microvm

Ref OS-251

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -786,7 +786,7 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 	sshkey := base64.StdEncoding.EncodeToString(pubKey)
 
 	// Build kernel command line arguments
-	kernelArgs := kernelConsole + " nomodeset random.trust_cpu=on tsc=reliable no_timer_check cryptomgr.notests rcupdate.rcu_expedited=1 panic=-1 " + cmdlineVar + " sshkey=" + sshkey + " melange_qemu_runner=1"
+	kernelArgs := kernelConsole + " nomodeset random.trust_cpu=on no_timer_check cryptomgr.notests rcupdate.rcu_expedited=1 panic=-1 " + cmdlineVar + " sshkey=" + sshkey + " melange_qemu_runner=1"
 
 	// Check for TESTING environment variable and pass it to microvm-init
 	// TESTING must be a number (0 for disabled, non-zero for enabled)


pkg/container/qemu_runner.go
```diff
@@ -778,7 +786,7 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 	sshkey := base64.StdEncoding.EncodeToString(pubKey)
 
 	// Build kernel command line arguments
-	kernelArgs := kernelConsole + " nomodeset random.trust_cpu=on panic=-1 " + cmdlineVar + " sshkey=" + sshkey + " melange_qemu_runner=1"
+	kernelArgs := kernelConsole + " nomodeset random.trust_cpu=on tsc=reliable no_timer_check cryptomgr.notests rcupdate.rcu_expedited=1 panic=-1 " + cmdlineVar + " sshkey=" + sshkey + " melange_qemu_runner=1"
```
EyeCantCU: The default clocksource is `kvm-clock` and not `tsc` so I believe we can remove `tsc=reliable` from our kargs

pkg/container/qemu_runner.go
```diff
@@ -860,12 +868,7 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 	// append raw disk, init will take care of formatting it if present.
 	baseargs = append(baseargs, "-object", "iothread,id=io1")
 	baseargs = append(baseargs, "-device", "virtio-blk-pci,drive=disk0,iothread=io1,packed=on,num-queues="+fmt.Sprintf("%d", max(1, nproc/2)))
-	if runtime.GOOS == "linux" {
-		baseargs = append(baseargs, "-drive", "if=none,id=disk0,cache=unsafe,cache.direct=on,format=raw,aio=native,file="+diskFile)
-	}
-	if runtime.GOOS == "darwin" {
-		baseargs = append(baseargs, "-drive", "if=none,id=disk0,cache=unsafe,format=raw,aio=threads,file="+diskFile)
-	}
```
EyeCantCU, line 868: I'd like to know a little more as both have tradeoffs. Why did we decide to go with threads over native on Linux?

89luca89: Empirically i've found it to be better for builds involving many files
using aio=native needs either cache=unsafe or cache=disabled, which does not use host's i/o cache, while threads allow for this
also it unifies the code path with macos reducing complexity

pkg/container/qemu_runner.go
89luca89: right, will remove :+1:

89luca89: done

pkg/container/qemu_runner.go
EyeCantCU: Thanks for explaining!

PR #2365, opened 2026-02-20
Use namespace instead of ReleaseData for SBOM generation.

wlynch:
This was noticed in some packages - we were using `/etc/os-release` to populate the `SPDXRef-OperatingSystem` package, but these do not have to match since this is pulling the host OS release data, but we want to populate the intended target namespace OS info.

### DIFF
--- a/pkg/build/sbom/spdx/spdx_test.go
+++ b/pkg/build/sbom/spdx/spdx_test.go
@@ -24,7 +24,6 @@ import (
 	"time"
 
 	apkofs "chainguard.dev/apko/pkg/apk/fs"
-	apko_build "chainguard.dev/apko/pkg/build"
 	"chainguard.dev/apko/pkg/sbom/generator/spdx"
 	"github.com/google/go-cmp/cmp"
 	purl "github.com/package-url/packageurl-go"
@@ -76,10 +75,6 @@ func TestSBOMGeneration(t *testing.T) {
 			License: "Apache-2.0",
 			PURL:    &testPURL,
 		},
-		ReleaseData: &apko_build.ReleaseData{
-			ID:        "test-os",
-			VersionID: "1.0",
-		},
 	}
 
 	gen := &Generator{}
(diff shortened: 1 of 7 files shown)


xnox:
@wlynch thank you! we will fix this up and release it. Enjoy :skier:

pkg/build/testdata/goldenfiles/sboms/7zip-two-fetches-2301-r3.spdx.json
```diff
@@ -13,13 +13,12 @@
...
   "documentDescribes": [
-    "SPDXRef-Package-apk-7zip-two-fetches-2301-r3"
+    "SPDXRef-Package-7zip-two-fetches-2301-r3"
   ],
   "packages": [
     {
       "SPDXID": "SPDXRef-OperatingSystem",
       "name": "wolfi",
-      "versionInfo": "20230201",
       "filesAnalyzed": false,
```
xnox, line 22: i guess we want to possibly plumb or hardcode versions back in. but it is true that we want to be versionless really. because we are rolling.

pkg/build/sbom/spdx/spdx_test.go
```diff
@@ -100,12 +95,11 @@ func TestSBOMGeneration(t *testing.T) {
...
 			DataLicense:       "CC0-1.0",
 			Namespace:         "https://spdx.org/spdxdocs/chainguard/melange/e43c05eed89f57b011808279db234a08",
-			DocumentDescribes: []string{"SPDXRef-Package-apk-test-pkg-1.2.3-r2"},
+			DocumentDescribes: []string{"SPDXRef-Package-test-pkg-1.2.3-r2"},
 			Packages: []spdx.Package{
 				{
 					ID:               "SPDXRef-OperatingSystem",
-					Name:             "test-os",
-					Version:          "1.0",
+					Name:             "test-ns",
```
xnox, line 102: @luhring do you think lack of version field will trip up grype NAK support?

PR #2368, opened 2026-02-23
Add `shallow-submodules` and `submodule-jobs` options for `git-checkout`

Taffer:
The `shallow-submodules` option might be redundant with `--depth 1`. Clickhouse in particular seems to benefit from this, and adding `--jobs=$(nproc)` as the default stops us from having to add it in various places (mostly Clickhouse, of course).

Defaults (`shallow-submodules: true` and `submodule-jobs: $(nproc)` seem reasonable (and useful) to me, but there are probably use cases I'm not considering.

### DIFF
--- a/pkg/build/pipelines/git-checkout.yaml
+++ b/pkg/build/pipelines/git-checkout.yaml
@@ -286,11 +286,12 @@ pipeline:
           vr git config --global --add safe.directory "$dest_fullpath"
 
           msg "Attempting git clone with retry (max_retries=$max_retries, initial_backoff=${initial_backoff}s, max_backoff=${max_backoff}s)"
+          # shellcheck disable=SC2086
           retry_with_backoff "$max_retries" "$initial_backoff" "$max_backoff" \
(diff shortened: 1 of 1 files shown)


pkg/build/pipelines/git-checkout.yaml
```diff
@@ -34,6 +34,16 @@ inputs:
...
+  shallow-submodules:
+    description: |
+      Whether to use --shallow-submodules when recurse-submodules is true.
+      Ignored if recurse-submodules is false.
+    default: true
+  submodule-jobs:
+    description: |
+      The number of concurrent jobs to use when recurse-submodules is true.
+      Ignored if recurse-submodules is false.
+    default: The number of cores returned by nproc.
```
sergiodj: I think this needs to contain an actual value.  I'd go with `1` to align with `git-submodule`'s default.

Taffer: OMG yes it does :laughing:

PR #2370, opened 2026-02-24
FDO package metadata improvements

xnox:
Relocatable links currently break with package-metadata, or result in duplicates, or result in false positives.

Also publish package-metadata integration for static translation units.

References:
- https://src.fedoraproject.org/rpms/package-notes/c/fc99dac13b7aba2053296be6b1166dbcaceda2df?branch=rawhide

### DIFF
--- a/pkg/build/compiler_config.go
+++ b/pkg/build/compiler_config.go
@@ -34,6 +34,38 @@ var gccLinkTemplate = `*link:
 var clangConfigTemplate = `-Xlinker --package-metadata='` + packageMetadataTemplate + `'
 `
 
+// fdoNoteTemplate is a C header, generated alongside the gcc spec file, that
+// projects can #include to embed the very same FDO packaging metadata that the
+// --package-metadata linker flag would emit. This is useful for statically
+// linked or vendored code (e.g. a bundled allocator), which leaves no trace in
+// the dynamic dependency graph and is therefore invisible to SBOM / CVE
+// scanners: the note travels with the object into the final binary regardless.
+//
+// The note layout (NT_FDO_PACKAGING_METADATA, owner "FDO") follows
+// https://systemd.io/COREDUMP_PACKAGE_METADATA/ . The %s is the C-string-escaped
+// metadata JSON, identical to what the linker flag carries.
+var fdoNoteTemplate = `#ifndef MELANGE_PACKAGE_NOTE_H
+#define MELANGE_PACKAGE_NOTE_H
+#if defined(__ELF__)
+#define MELANGE_PACKAGE_NOTE_JSON "%s"
+__attribute__((used, retain, section(".note.package"), aligned(4)))
+static const struct {
+	unsigned int namesz;
+	unsigned int descsz;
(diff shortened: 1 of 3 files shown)


antitree (dismissed):

pkg/build/compiler_config_test.go
```diff
@@ -106,6 +107,52 @@ func TestCreateGccSpecFileError(t *testing.T) {
...
+
+	content, err := os.ReadFile(headerPath)
+	require.NoError(t, err)
+	s := string(content)
+
+	// The generated header carries the note attribute and the well-known
+	// variable name, in the FDO .note.package section.
+	assert.Contains(t, s, `__attribute__((used, retain, section(".note.package"), aligned(4)))`)
+	assert.Contains(t, s, "melange_package_note")
+	assert.Contains(t, s, "0xcafe1a7e")
```
a-crate, line 142: This is really opaque even with the spec linked by the header generation, deserves a comment like `// FDO note type` IMO

pkg/build/compiler_config.go
```diff
@@ -58,6 +90,37 @@ func (b *Build) createGccSpecFile() error {
...
+
+	// Escape for embedding in a C string literal: backslash first so we don't
+	// double-escape the backslashes we introduce for the quotes.
+	escaped := strings.ReplaceAll(metadata, `\`, `\\`)
+	escaped = strings.ReplaceAll(escaped, `"`, `\"`)
+
+	content := fmt.Sprintf(fdoNoteTemplate, escaped)
+
+	// #nosec G306 -- header should be world-readable so builds can include it
+	return os.WriteFile(filepath.Join(b.WorkspaceDir, ".melange.fdo.h"), []byte(content), 0o644)
```
a-crate, line 121: Do we generate any other header files? I would like if we injected `-I ${workspaceDir}/.melange/include` somewhere so people could just `#include <melange_fdo.h>` instead of a magic location but maybe that's not realistic.

a-crate (commented):
Basically LGTM, if my desire to not make people hardcode the header location is not realistic I'm fine approving as is. I guess we probably already hard code the gcc spec location so I don't know how much a dedicated include directive would improve the situation.

xnox:
> Basically LGTM, if my desire to not make people hardcode the header location is not realistic I'm fine approving as is

Unfortunately it is not realistic to make this thing a public header; or shove it into system path; or somehow inject things - I tried hard to inject custom include paths for the eco-2.28 toolchain and it's basically impossible to control. This solution is the most universal for all types of builds.

But yes will address other comments.

pkg/build/compiler_config.go
xnox: This is a first header like thing.

And yes it is impossible to pass through '-I' reliably.

pkg/build/compiler_config_test.go
xnox: Yeah it does look like a magic value.

markusboehme (approved):

PR #2388, opened 2026-03-03
melange: fix git-checkout pipeline

zmarano:
Adding `set -u` will cause the pipeline to fail when a variable is not set. And we often don't have the SOURCE_DATE_EPOCH variable set in some pipelines. In particular, the iamguarded-test pipeline.


a-crate:
Wolfi APK repositories are updated immediately, there is just a small delay in making the source available at wolfi-dev/os. It's already run-merged, we just need to do runmerge for the test cases in melange (which I forgot about).

PR #2394, opened 2026-03-03
feat(qemu): optional build observability event logging

antitree:
## Summary

- Adds optional build observability support to the QEMU runner users can decide if they want to have an observability / telemetry step inside their builds
- When a `microvm-observability-hook` package is installed in the build VM, eBPF-based observability events (process execution, network connections) are captured during the build
- After the build completes, events are retrieved via SSHControlClient (port 2223) and logged to stdout with an `[OBSERVABILITY]` prefix
- Fully optional — builds without the hook produce zero observability output and zero errors

## Files changed

| File | Change |
|------|--------|
| `pkg/build/build.go` | 9 lines: call `RetrieveObservabilityEvents` + `LogObservabilityEvents` after workspace retrieval |
| `pkg/container/observability.go` | New: event retrieval via SSH, NDJSON parsing, network connection extraction, stdout logging |
| `pkg/container/observability_test.go` | New: 11 unit tests + integration/regression stubs |

## How it works

1. After `retrieveWorkspace` completes, probes the VM for an events file at known paths
2. If no file exists (hook not installed), returns nil silently — one SSH command overhead
3. If found, retrieves NDJSON events via `cat` over SSH, parses network connections
4. Logs summary + network connections + full raw events to stdout with `[OBSERVABILITY]` prefix

## Test plan

- [x] Unit tests pass (`go test ./pkg/container/ -race`)
- [x] Full test suite passes (`go test ./... -race`)
- [x] `golangci-lint` clean (zero new findings)
- [x] `make checkfmt` clean
- [x] E2E: default build without observability — zero output, zero errors
- [x] E2E: hello-test (shell) with observability — 8,890 events, 9 network connections
- [x] E2E: hello-wolfi (C/autoconf/gcc) with observability — 17,449 events, 10 network connections
- [x] E2E: age (Go) with observability — 9,640 events, 40 network connections
- [x] Build output (APKs) identical with and without observability

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,7 @@ require (
 	github.com/docker/docker v28.5.2+incompatible
 	github.com/dprotaso/go-yit v0.0.0-20250513224043-18a80f8f6df4
 	github.com/github/go-spdx/v2 v2.4.0
-	github.com/go-git/go-git/v5 v5.17.0
+	github.com/go-git/go-git/v5 v5.17.1
 	github.com/google/go-cmp v0.7.0
 	github.com/google/go-containerregistry v0.21.3
 	github.com/google/licenseclassifier/v2 v2.0.0
--- a/go.sum
+++ b/go.sum
@@ -131,8 +131,8 @@ github.com/go-git/go-billy/v5 v5.8.0 h1:I8hjc3LbBlXTtVuFNJuwYuMiHvQJDq1AT6u4DwDz
 github.com/go-git/go-billy/v5 v5.8.0/go.mod h1:RpvI/rw4Vr5QA+Z60c6d6LXH0rYJo0uD5SqfmrrheCY=
 github.com/go-git/go-git-fixtures/v4 v4.3.2-0.20231010084843-55a94097c399 h1:eMje31YglSBqCdIqdhKBW8lokaMrL3uTkpGYlE2OOT4=
 github.com/go-git/go-git-fixtures/v4 v4.3.2-0.20231010084843-55a94097c399/go.mod h1:1OCfN199q1Jm3HZlxleg+Dw/mwps2Wbk9frAWm+4FII=
-github.com/go-git/go-git/v5 v5.17.0 h1:AbyI4xf+7DsjINHMu35quAh4wJygKBKBuXVjV/pxesM=
-github.com/go-git/go-git/v5 v5.17.0/go.mod h1:f82C4YiLx+Lhi8eHxltLeGC5uBTXSFa6PC5WW9o4SjI=
+github.com/go-git/go-git/v5 v5.17.1 h1:WnljyxIzSj9BRRUlnmAU35ohDsjRK0EKmL0evDqi5Jk=
+github.com/go-git/go-git/v5 v5.17.1/go.mod h1:pW/VmeqkanRFqR6AljLcs7EA7FbZaN5MQqO7oZADXpo=
 github.com/go-jose/go-jose/v3 v3.0.4 h1:Wp5HA7bLQcKnf6YYao/4kpRpVMp/yf6+pJKV8WFSaNY=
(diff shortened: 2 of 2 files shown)


pkg/build/build.go
```diff
@@ -784,6 +784,15 @@ func (b *Build) BuildPackage(ctx context.Context) error {
 	}
 	log.Infof("retrieved and wrote post-build workspace to: %s", b.WorkspaceDir)
 
+	// Retrieve and log build observability events if the observability hook
+	// is installed. Fully optional — if the hook is not in the VM, the probe
+	// finds no events file and returns nil with no impact on the build.
+	if obsEvents, err := container.RetrieveObservabilityEvents(ctx, cfg); err != nil {
```
89luca89: You call container.RetrieveObservabilityEvents(ctx, cfg) unconditionally. The only guard inside RetrieveObservabilityEvents is cfg.SSHControlClient ==
  nil. While this happens to be nil for bubblewrap/docker runners, it's an implicit contract rather than an explicit check. If another runner ever populates SSHControlClient, this would  unexpectedly try to probe event paths. Consider gating on b.Runner.Name() == container.QemuName.

pkg/container/observability.go
```diff
@@ -0,0 +1,236 @@
...
+
+	log := clog.FromContext(ctx)
+
+	// Probe known event file locations. If none exist, the observability
+	// hook is not installed and we silently return nil.
+	eventsPath := ""
+	for _, path := range observabilityEventPaths {
+		var checkBuf bytes.Buffer
+		err := sendSSHCommand(ctx, cfg.SSHControlClient, cfg, nil, nil, &checkBuf, false,
+			[]string{"sh", "-c", fmt.Sprintf("test -f %s && echo exists || echo missing", path)})
```
89luca89: quote the path: fmt.Sprintf("test -f '%s' && ...", path). Better yet, use []string{"test", "-f", path} and check the exit code.

pkg/container/observability.go
```diff
@@ -0,0 +1,236 @@
...
+		// This is the normal case for default builds; return silently.
+		return nil, nil
+	}
+
+	log.Infof("qemu: found observability events at %s", eventsPath)
+
+	// Retrieve the events file
+	var eventsBuf bytes.Buffer
+	err := sendSSHCommand(ctx, cfg.SSHControlClient, cfg, nil, nil, &eventsBuf, false,
+		[]string{"cat", eventsPath})
```
89luca89: real builds generate 9k-17k+ events. A pathological or compromised VM   could produce gigabytes of data maybe we can do a fetch+read step or stream them?

pkg/container/observability_test.go
```diff
@@ -0,0 +1,545 @@
...
+		if sock.DPort != wantPort {
+			t.Errorf("for IP %s: dport = %d, want %d", sock.DAddr, sock.DPort, wantPort)
+		}
+	}
+}
+
+func TestTracingPolicyYAMLValid(t *testing.T) {
+	policyPath := filepath.Join("..", "..", "..", "..", "stereo", "os", "microvm-observability-hook", "network-monitor.yaml")
+	if _, err := os.Stat(policyPath); os.IsNotExist(err) {
+		policyPath = "/home/mark-manning/projects/stereo/os/microvm-observability-hook/network-monitor.yaml"
```
89luca89: hardcoded path

pkg/container/observability_test.go
```diff
@@ -0,0 +1,545 @@
...
+		if !strings.Contains(content, field) {
+			t.Errorf("TracingPolicy missing required field: %q", field)
+		}
+	}
+}
+
+func TestSystemdUnitFileValid(t *testing.T) {
+	unitPath := filepath.Join("..", "..", "..", "..", "stereo", "os", "microvm-observability-hook", "tetragon.service")
+	if _, err := os.Stat(unitPath); os.IsNotExist(err) {
+		unitPath = "/home/mark-manning/projects/stereo/os/microvm-observability-hook/tetragon.service"
```
89luca89: hardcoded path

89luca89 (dismissed):

PR #2397, opened 2026-03-04
Remove erroneous variable quoting

Taffer:
Scott noticed a couple of spots where I shouldn't have quoted variables; we don't seem to have a test that would have triggered on this, but there's probably something in the repos that would!

### DIFF
--- a/.github/workflows/wolfi-presubmit.yaml
+++ b/.github/workflows/wolfi-presubmit.yaml
@@ -105,7 +105,7 @@ jobs:
           melange version
 
       # needed to run tests for Bubblewrap packages
-      - uses: docker/setup-docker-action@e43656e248c0bd0647d3f5c195d116aacf6fcaf4 # v4.7.0
+      - uses: docker/setup-docker-action@1a6edb0ba9ac496f6850236981f15d8f9a82254d # v5.0.0
 
       # this need to point to main to always get the latest action
       - uses: wolfi-dev/actions/install-wolfictl@d8faf0b2bf2a7c6eefc571567ef370faae5baed2 # last commit that had the scan command


pkg/build/pipelines/git-checkout.yaml
```diff
@@ -292,15 +293,16 @@ pipeline:
               "--config=user.name=Melange Build" \
               "--config=user.email=melange-build@cgr.dev" \
               $flags \
-              ${depthflag:+"$depthflag"} "$repo" "$workdir" ||
+              "$depthflag" "$repo" "$workdir" ||
```
smoser: this is only good if depthflag _always_ has a value other than "" here.

Taffer: ... wait, what's wrong with putting an empty string there?

```sh
🐚  chrish@asteroid ❯ function z() { local foo="" ; echo a "$foo" b ; }

🐚  chrish@asteroid ❯ z
a  b
```

has the same output as:

```sh
🐚  chrish@asteroid ❯ function z() { local foo="" ; echo a "${foo:+"$foo"}" b ; }

sparse_paths via 🐹 v1.25.6
🐚  chrish@asteroid ❯ z
a  b
```

smoser: it will go as an argument to git.
An example:

```console
$ nodecorate="--no-decorate"
$ git log --oneline "$nodecorate" HEAD^..
44654e54 chore(dependabot): allow for minor gomod updates (#2386)
$ nodecorate=""
$ git log --oneline "$nodecorate" HEAD^..
fatal: ambiguous argument '': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

but if you use it  with `:+`, then it works in both ways:

```console
$ flag="--no-decorate"
$ git log --oneline ${flag:+"$flag"} HEAD^..
44654e54 chore(dependabot): allow for minor gomod updates (#2386)

$ flag=""
$ git log --oneline ${flag:+"$flag"} HEAD^..
44654e54 (HEAD, tag: v0.43.6, upstream/main, upstream/HEAD) chore(dependabot): allow for minor gomod updates (#2386)
```

Taffer: :exploding_head: I know nothing of shell arg parsing.

PR #2412, opened 2026-03-10
qemu: add SSH keepalives, handshake timeouts, and error wrapping

jonjohnsonjr:
- Add 30s keepalive requests on all SSH clients with a max of 3 consecutive misses before closing the connection
- Replace ssh.Dial with sshDialWithTimeout to set a deadline covering the full SSH handshake, not just TCP connect (golang/go#19338)
- Wrap previously-swallowed errors with %w for debuggability
- Add tests using synctest for keepalive timing logic

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -68,6 +68,20 @@ const QemuName = "qemu"
 
 const (
 	defaultDiskSize = "50Gi"
+
+	// sshDialTimeout is the maximum time allowed for TCP connect + SSH handshake.
+	sshDialTimeout = 30 * time.Second
+
+	// sshKeepaliveInterval is how often keepalive requests are sent.
+	sshKeepaliveInterval = 30 * time.Second
+
+	// sshKeepaliveMaxMissed is the number of consecutive missed keepalives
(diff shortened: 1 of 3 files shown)


pkg/container/qemu_runner.go
```diff
@@ -1147,29 +1154,125 @@ func setupSSHClients(ctx context.Context, cfg *Config) error {
...
+	startSSHKeepalive(ctx, cfg.SSHControlClient, "control")
+	startSSHKeepalive(ctx, cfg.SSHControlBuildClient, "control-build")
+
 	return nil
 }
 
+// sshDialWithTimeout dials an SSH connection with a deadline covering both the
+// TCP connect and the SSH handshake. This prevents hangs when the remote end
+// accepts the TCP connection but stalls during the handshake (golang/go#19338).
+func sshDialWithTimeout(network, addr string, config *ssh.ClientConfig, timeout time.Duration) (*ssh.Client, error) {
```
EyeCantCU:
```suggestion
func sshDialWithTimeout(network string, addr string, config *ssh.ClientConfig, timeout time.Duration) (*ssh.Client, error) {
```

PR #2414, opened 2026-03-11
qemu: allow selection of alternate source for microvm initramfs

a-crate:
It's currently impossible to select an alternate microvm-init sourcefor the qemu runner, which makes it very difficult to test. I don't love the growing number of environment variables necessary to configure the qemu runner but I don't have a better solution. I don't think turning all of these into flags is better.

I've also added a command to generate an initramfs for a given configuration.
Since the modules are attached the kernel (which we don't know until runtime) instead of the initramfs, I moved injecting them to the same time we inject SSH host keys.

### DIFF
--- a/docs/md/melange.md
+++ b/docs/md/melange.md
@@ -25,6 +25,7 @@ toc: true
 * [melange compile](/docs/md/melange_compile.md)	 - Compile a YAML configuration file
 * [melange completion](/docs/md/melange_completion.md)	 - Generate completion script
 * [melange index](/docs/md/melange_index.md)	 - Creates a repository index from a list of package files
+* [melange initramfs](/docs/md/melange_initramfs.md)	 - Build a base initramfs for the QEMU runner
 * [melange keygen](/docs/md/melange_keygen.md)	 - Generate a key for package signing
 * [melange license-check](/docs/md/melange_license-check.md)	 - Gather and check licensing data
 * [melange lint](/docs/md/melange_lint.md)	 - EXPERIMENTAL COMMAND - Lints an APK, checking for problems and errors
--- a/docs/md/melange_initramfs.md
+++ b/docs/md/melange_initramfs.md
@@ -1,0 +1,66 @@
+---
+title: "melange initramfs"
+slug: melange_initramfs
+url: /docs/md/melange_initramfs.md
+draft: false
+images: []
+type: "article"
+toc: true
+---
+## melange initramfs
+
+Build a base initramfs for the QEMU runner
+
+### Synopsis
+
+Build a base initramfs that can be used with the QEMU runner.
+
+The generated initramfs can be reused across multiple builds by setting
+the QEMU_BASE_INITRAMFS environment variable to point to the output file.
+
+The generated initramfs does NOT contain SSH host keys.
+Keys and modules are injected at runtime for each build.
+
+```
+melange initramfs [flags]
+```
+
+### Examples
+
+```
+  # Generate default initramfs for x86_64
+  melange initramfs --arch x86_64 --output ./initramfs.cpio
+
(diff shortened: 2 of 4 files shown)


pkg/container/qemu_runner.go
```diff
@@ -2233,6 +2234,57 @@ func getAdditionalPackages(ctx context.Context) []string {
 	return packages
 }
 
+// microvmInitConfig holds configuration for the microvm-init package source.
```
egibs: Thinking out loud for later -- is it time to split this file up into a more manageable `qemu` package given that this is now ~2500 lines?

a-crate: It is quite large. Up to you if you want that in this PR or to refactor separately.

egibs: Totally fine to leave for a future PR 🙂

pkg/cli/build_qemu_initramfs.go
```diff
@@ -0,0 +1,113 @@
...
+		outputPath   string
+		archStr      string
+		initPackage  string
+		repositories []string
+		packages     []string
+		extraKeys    []string
+	)
+
+	cmd := &cobra.Command{
+		Use:   "build-qemu-initramfs",
```
jonjohnsonjr: This is a bit of a mouthful. Do you think just `initramfs` would be sufficiently unambiguous here to save typing?

pkg/container/qemu_runner.go
```diff
@@ -2088,65 +2087,103 @@ func secureDelete(ctx context.Context, path string) {
...
+		 * cache it.
+		 * if present, we nop and return, else we build it.
+		 */
+		cacheDir := filepath.Join(
+			"kernel",
+			cfg.Arch.ToAPK(),
+			"melange-cpio")
+
+		err := os.MkdirAll(cacheDir, 0o755)
 		if err != nil {
```
jonjohnsonjr: combine these lines

pkg/container/qemu_runner.go
```diff
@@ -2088,65 +2087,103 @@ func secureDelete(ctx context.Context, path string) {
...
+			err = GenerateBaseInitramfs(ctx, cfg.Arch, microvmCfg, baseInitramfs)
+			if err != nil {
+				return "", err
+			}
 		}
 	}
 
-	return injectHostKeyIntoCpio(ctx, cfg, baseInitramfs)
+	// Inject SSH host keys and modules
+	result, err := injectRuntimeData(ctx, cfg, os.Getenv("QEMU_KERNEL_MODULES"), baseInitramfs)
```
jonjohnsonjr: just return here

pkg/container/qemu_runner.go
```diff
@@ -2301,6 +2323,42 @@ func injectHostKeyIntoCpio(ctx context.Context, cfg *Config, baseInitramfs strin
...
+				f.Close()
+				if err != nil {
+					return err
+				}
+				moduleRecords = append(moduleRecords, cpio.StaticFile(archivePath, string(contents), 0o755))
+			}
+			return nil
+		}
+		err := fs.WalkDir(modfs, ".", walkModules)
+		if err != nil {
```
jonjohnsonjr: combine these lines

pkg/container/qemu_runner.go
```diff
@@ -2301,6 +2323,42 @@ func injectHostKeyIntoCpio(ctx context.Context, cfg *Config, baseInitramfs strin
...
+			return "", fmt.Errorf("failed to generate module cpio records: %w", err)
+		}
+		err = cpio.WriteRecordsAndDirs(cpioWriter, moduleRecords)
+		if err != nil {
```
jonjohnsonjr: combine these lines

PR #2423, opened 2026-03-16
pipeline: execute them with set -eo pipefail

aborrero:
If a pipeline is 'cmd1 | cmd2' we want to fail the execution if cmd1 failed.

Previous to this patch, we had a raising number of pipelines that are manually encoding
'set -eo pipefail' on each instance.

Signed-off-by: Arturo Borrero Gonzalez <arturo.borrero@chainguard.dev>

### DIFF
--- a/pkg/build/pipeline.go
+++ b/pkg/build/pipeline.go
@@ -181,11 +181,15 @@ func buildEvalRunCommand(_ *config.Pipeline, debugOption rune, workdir string, f
 	// Using go-shellquote library for robust shell escaping
 	safeWorkdir := quoteShellArg(workdir)
 
-	script := fmt.Sprintf(`set -e%c
+	xFlag := ""
+	if debugOption == 'x' {
+		xFlag = "x"
+	}
+	script := fmt.Sprintf(`set -e%so pipefail
 [ -d %s ] || mkdir -p %s
 cd %s
 %s
-exit 0`, debugOption, safeWorkdir, safeWorkdir, safeWorkdir, fragment)
+exit 0`, xFlag, safeWorkdir, safeWorkdir, safeWorkdir, fragment)
 	return []string{"/bin/sh", "-c", script}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/build/pipeline.go
```diff
@@ -181,11 +181,15 @@ func buildEvalRunCommand(_ *config.Pipeline, debugOption rune, workdir string, f
 	// Using go-shellquote library for robust shell escaping
 	safeWorkdir := quoteShellArg(workdir)
 
-	script := fmt.Sprintf(`set -e%c
+	xFlag := ""
+	if debugOption == 'x' {
+		xFlag = "x"
+	}
```
markusboehme, line 187: Nit: I think a better change would be to remove `debugOption rune` from the function signature and replace it with something like `enableTracing bool`. Call sites don't have to know about `x`/`-xtrace`.

pkg/build/pipeline.go
```diff
@@ -181,11 +181,15 @@ func buildEvalRunCommand(_ *config.Pipeline, debugOption rune, workdir string, f
 	// Using go-shellquote library for robust shell escaping
 	safeWorkdir := quoteShellArg(workdir)
 
-	script := fmt.Sprintf(`set -e%c
+	xFlag := ""
+	if debugOption == 'x' {
+		xFlag = "x"
```
justinvreeland, line 186: This is going to cause a lot of spam we don't need an debugs run with -x? Why would we enabled it for all pipelines? @smoser didn't we disable this behavior in the past?

aborrero: Previous to this patch we were already enabling -x for pipelines in debug mode no? the behavior should be unchanged in this patch.

smoser: just seeing this now. I do find `set -x` an anti-pattern. it makes the output of better tests, that provide reasonable output all of a sudden unreadable.  but, as aborrero pointed out, it already was.

smoser:
@aborrero , 
this caused some passing tests to fail.
https://github.com/chainguard-dev/melange/issues/2478

PR #2438, opened 2026-03-26
feat(harden-runner): move to inline block policies

egibs:
This PR moves harden-runner to block mode similar to other repositories. The endpoints are derived from stable baseline data in StepSecurity and should be representative of the usual activity for these Workflows.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -2,9 +2,9 @@ name: ci
 
 on:
   push:
-    branches: [ "main" ]
+    branches: ["main"]
   pull_request:
-    branches: [ "main" ]
+    branches: ["main"]
 
 permissions: {}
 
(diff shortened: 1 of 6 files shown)


_stepsec_fetch.py
```diff
@@ -0,0 +1,45 @@
+import urllib.request, json, subprocess, sys
+
+api_key = subprocess.check_output(['op', 'read', 'op://employee/StepSecurity API/notesPlain']).decode().strip()
+
+all_jobs = []
+page = 1
+while True:
+    req = urllib.request.Request(
+        f'https://agent.api.stepsecurity.io/v1/github/chainguard-dev/terraform-provider-imagetest/jobs/baseline?page={page}',
```
antitree: 0_o am I reading this wrong?

antitree: Just double checking this is the right name. Not melange?

egibs: Removed. Good catch 🙂

PR #2471, opened 2026-04-09
pkg/sca: add static:lib<name> provides and dependency tracking

smoser:
Scan .a files in standard library directories and emit static:lib<name> provides via SCA, mirroring how so: and pc: provides are generated.

Also parse Libs.private fields from .pc files to discover static:lib<name> runtime dependencies. This generation is gated behind the MELANGE_GENERATE_STATIC_DEPS environment variable (accepted values: "true" or "false"); when disabled, candidate dependencies are only logged so the impact can be evaluated before enabling.

### DIFF
--- a/pkg/sca/e2e_test.go
+++ b/pkg/sca/e2e_test.go
@@ -131,6 +131,8 @@ func TestAnalyze(t *testing.T) {
 				"so:libwebpdemux-df9b36c7.so.2.0.14=2.0.14",
 				"so:libwebpmux-9fe05867.so.3.0.13=3.0.13",
 				"so:libxcb-f0538cc0.so.1.1.0=1.1.0",
+				"static:libnpymath",
+				"static:libnpyrandom",
 			},
 		},
 	}, {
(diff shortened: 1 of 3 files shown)


xnox (changes requested):
For `make package/yara`, does this produce what you expect?

I got this:

```
$ cat .PKGINFO 
# Generated by melange
pkgname = yara-dev
pkgver = 4.5.5-r4
arch = x86_64
size = 1327707
origin = yara
pkgdesc = yara dev
url = 
commit = 4a947e9ea1a780700c9a08b6bf3fd6f1b39564a4
maintainer = wolfi
builddate = 1776449776
license = BSD-3-Clause
depend = pc:libcrypto
depend = so:libyara.so.10
depend = yara
provides = pc:yara=4.5.5-r4
provides = static:libyara
datahash = 6189486c4c5d7d42897842cc3c72ec6031391da3a4457278f5c4726659eb8259
```

And the pc file is this https://apk.chaindag.dev/https/packages.wolfi.dev/os/x86_64/yara-dev-4.5.5-r4.apk@sha1:d9a54cb0de2b4589b0da207a183e1464c0c70036/usr/lib/pkgconfig/yara.pc

```
prefix=/usr
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=/usr/lib

Name: yara
Description: YARA library
URL: https://virustotal.github.io/yara/
Version: 4.5.5
Requires.private:  libcrypto
Cflags: -I${includedir}
Libs: -L${libdir} -lyara
Libs.private:  -lm -lmagic 
```

Should there be static dependency on magic?

Also not sure how you want to handle Requires.private - as that says that for static linking on needs what ever static linking flags are specified in libcrypto.pc as in like a dep on static:libcrypto.

PR #2481, opened 2026-04-14
feat(qemu): add DNS search domains

egibs:
There are certain cases where we need to set DNS search domains in a more configurable way.

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -755,7 +755,18 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 	baseargs = append(baseargs, "-nodefaults")
 	baseargs = append(baseargs, serialArgs...)
 	// use -netdev + -device instead of -nic, as this is better supported by microvm machine type
-	baseargs = append(baseargs, "-netdev", "user,id=id1,hostfwd=tcp:"+cfg.SSHAddress+"-:22,hostfwd=tcp:"+cfg.SSHControlAddress+"-:2223")
+	netdevArgs := "user,id=id1,hostfwd=tcp:" + cfg.SSHAddress + "-:22,hostfwd=tcp:" + cfg.SSHControlAddress + "-:2223"
+	// QEMU_DNS_SEARCH allows configuring DNS search domains inside the guest VM.
+	// This is useful for builds that need to resolve short hostnames via search
+	// domains, or when the build environment requires specific DNS resolution
+	// behavior. The search domain is passed to SLIRP which includes it in DHCP
+	// responses, so the guest receives it naturally via DHCP.
+	// Example: QEMU_DNS_SEARCH=example.com
+	if dnsSearch, ok := os.LookupEnv("QEMU_DNS_SEARCH"); ok {
+		log.Infof("qemu: QEMU_DNS_SEARCH set to %s, adding to SLIRP network config", dnsSearch)
+		netdevArgs += ",dnssearch=" + dnsSearch
+	}
+	baseargs = append(baseargs, "-netdev", netdevArgs)
 	// Set host_mtu to avoid silent packet drops in nested environments (e.g.,
 	// QEMU inside GKE pods). SLIRP defaults to 1500 MTU but the host path MTU
 	// may be lower due to encapsulation (GCP VPC uses 1460, pod networks can be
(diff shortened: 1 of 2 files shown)


antitree (dismissed):

smoser (changes requested):
One request - add documentation.

As a comment, It is somewhat weird to expect the guest to get some network config from a dhcp request and some from the kernel command line. You could also do this with qemu user networking (`-nic user,dnssearch=....`) and then the network is set up to do the right thing.  You could also pass full static networking on the kernel command line.

egibs:
> One request - add documentation.
> 
> As a comment, It is somewhat weird to expect the guest to get some network config from a dhcp request and some from the kernel command line. You could also do this with qemu user networking (`-nic user,dnssearch=....`) and then the network is set up to do the right thing. You could also pass full static networking on the kernel command line.

Added a comment + made the functionality more consistent in [`9fb5981` (this PR)](https://github.com/chainguard-dev/melange/pull/2481/commits/9fb5981d69e9ece4356feb3c5525a60c10d43da6).

pkg/container/qemu_runner.go
```diff
@@ -792,7 +750,18 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
...
+	// QEMU_DNS_SEARCH allows configuring DNS search domains inside the guest VM.
+	// This is useful for builds that need to resolve short hostnames via search
+	// domains, or when the build environment requires specific DNS resolution
+	// behavior. The search domain is passed to SLIRP which includes it in DHCP
+	// responses, so the guest receives it naturally via DHCP.
+	// Example: QEMU_DNS_SEARCH=example.com
+	if dnsSearch, ok := os.LookupEnv("QEMU_DNS_SEARCH"); ok {
+		log.Infof("qemu: QEMU_DNS_SEARCH set to %s, adding to SLIRP network config", dnsSearch)
+		netdevArgs += ",dnssearch=" + dnsSearch
+	}
```
smoser, line 773: it looks like in order to specify multiple dns search domains you repeat 'dnssearch='.
so we should be explicity and have this be comma (or space) delimited.  So either

    QEMU_DNS_SEARCH="my.domainname.com my.otherdomainname.com"
    or 
    QEMU_DNS_SEARCH="m.domainname.com,myother.com"

and then validate that each token only has valid domain name chars.

otherwise the user could sneak in other options to netdev here.

pkg/container/qemu_runner.go
```diff
@@ -2330,6 +2268,69 @@ func getAdditionalPackages(ctx context.Context) []string {
...
+// dnsSearchDomainRegex matches valid DNS search domain characters.
+// Only allows alphanumeric characters, dots, and hyphens.
+// This prevents injection of QEMU netdev options via malicious domain names.
+var (
+	// dnsSearchDomainRegex matches valid DNS search domain characters.
+	// Only allows alphanumeric characters, dots, and hyphens.
+	// This prevents injection of QEMU netdev options via malicious domain names.
+	dnsSearchDomainRegex = regexp.MustCompile(`^[a-zA-Z0-9.-]+$`)
+	// dnsSearchDelimiterRegex splits on comma or whitespace (space, tab, newline).
+	dnsSearchDelimiterRegex = regexp.MustCompile(`[,\s]+`)
```
smoser: i'm sorry if i wasn't clear. i was suggesting picking `,` or whitespace.  Did you have a reason why we'd pick both?  I guess i'd prefer `,` just to avoid there being newlines and such.

pkg/container/qemu_runner.go
```diff
@@ -792,7 +750,25 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
...
+	// Supports multiple domains: comma or space separated.
+	// Example: QEMU_DNS_SEARCH="example.com my.domain.org"
+	// Example: QEMU_DNS_SEARCH="example.com,my.domain.org"
+	if dnsSearch, ok := os.LookupEnv("QEMU_DNS_SEARCH"); ok {
+		domains := parseDNSSearchDomains(dnsSearch)
+		if len(domains) > 0 {
+			log.Infof("qemu: QEMU_DNS_SEARCH set to %v, adding %d domains to SLIRP network config", domains, len(domains))
+			netdevArgs += buildDNSSearchNetdevArgs(domains)
+		} else {
+			log.Warnf("qemu: QEMU_DNS_SEARCH contains invalid characters or is empty, ignoring: %s", dnsSearch)
```
smoser: rather than mostly silently doing something unexpected, can we just return an error?

make parseDNSSearchDomains return an error also rigth?

smoser (commented):
I'm sorry if this seems pedantic.  I'd just rather not open up arbitrary function inside a environment variable intended to have a specific use.

pkg/container/qemu_runner.go
egibs: Purely from a flexibility perspective. I'll enforce comma-delimited strings.

egibs: Updated in [`9f4a8c2` (this PR)](https://github.com/chainguard-dev/melange/pull/2481/commits/9f4a8c24a66dba308a9aa569a0423fc92148e0c1).

pkg/container/qemu_runner.go
egibs: Updated in [`9f4a8c2` (this PR)](https://github.com/chainguard-dev/melange/pull/2481/commits/9f4a8c24a66dba308a9aa569a0423fc92148e0c1).

smoser (approved):

PR #2489, opened 2026-04-17
fix(qemu): fix CPU/Memory resource precedence

egibs:
Currently, passing in `--cpu`/`--memory` externally will not apply due to the resource defaults we introduced here. Given the order in which we're calling things, these defaults overwrite what we pass in elsewhere, so we end up with `2 CPU` or `4Gi` no matter what.

This PR will fix the order for which resource configuration takes precedence.


egibs:
The `guac` and other failures are unrelated.

antitree (approved):

PR #2503, opened 2026-04-24
Make the JSON schema compatible with all package YAMLs

Taffer:
Also, regenerate test data.

This is probably looser than we want; next step is to enforce situations where one or the other is required, etc. This doesn't fix complaints about data types either.


Taffer:
I'm just going to give up touching `melange`. :disappointed:

Taffer:
This is too painful to fix from the Go code; I've got a shell script that'll (re)generate the scheme based on what's actually found in stereo.

PR #2511, opened 2026-05-04
chore(config): also create and upload source tarballs of gitlab repositories

sil2100:
We want our SBOMs to always have a downloadable downloadLocation for sources. For github and gitlab repositories we tried using the deterministic archive url, but it was reported that some of our users have problems downloading those tarballs from gitlab. Let's switch gitlab to use the tarball uploader and host those on our chainguard tarballs bucket.


sil2100:
The subversion build failures are unrelated.

pkg/config/config.go
```diff
@@ -659,19 +659,11 @@ func getGitSBOMPackage(repo, tag, expectedCommit string, idComponents []string,
...
 
-	case repoURL.Host == "gitlab.com":
-		repoType = purl.TypeGitlab
-		downloadLocation = fmt.Sprintf("%s://gitlab.com/%s/%s/-/archive/%s/%s.tar.gz", repoURL.Scheme, namespace, name, ref, ref)
-
-	case strings.HasPrefix(repoURL.Host, "gitlab") || hint == "gitlab":
-		repoType = purl.TypeGeneric
-		downloadLocation = fmt.Sprintf("%s://%s/%s/%s/-/archive/%s/%s.tar.gz", repoURL.Scheme, repoURL.Host, namespace, name, ref, ref)
-
 	default:
```
Taffer, line 667: Is GitLab handled here now? Otherwise we might need to add `case repoURL.Host == "gitlab.com"` back in if/when we move off GitHub. (And then again when we eventually move to self-hosted Forgejo... 🔮)

javacruft (approved):

PR #2532, opened 2026-05-17
fix(docker): use per-invocation unique image tag to avoid concurrent race

dustinkirkland:
## Summary

The docker runner tags every apko-built test/build image as the shared name `docker.io/library/melange:latest`. When multiple concurrent `melange test` or `melange build` invocations on the same host race to retag, one's "delete existing + tag new" sequence collides with another's:

```
failed to create an image docker.io/library/melange:latest with
target sha256:...: AlreadyExists: image already exists
```

This makes anything beyond ~6× concurrent local testing unreliable on a single host, and scales to ~100% failure rate with rates of 100x concurrency.

## Fix

Replace the shared `melange:latest` tag with a per-invocation unique tag of the form `melange:<8 hex chars>` derived from `crypto/rand`. 32 bits of randomness gives ~1-in-65k collision odds at 256 concurrent invocations (birthday bound) — comfortably beyond any realistic single-host concurrency.

The matching `RemoveImage` in `Build.Close` cleans up by the ref returned from `LoadImage`, so unique tags are cleaned up automatically with no other code changes.

## Reproduction (before the fix)

In a stereo checkout, with the docker runner:

```bash
for pkg in pkg1 pkg2 pkg3 pkg4 pkg5 pkg6; do
    MELANGE_RUNNER=docker make test/$pkg &
done
wait
# Several workers fail with "AlreadyExists: image docker.io/library/melange:latest"
```

After this fix, 8× concurrent local runs complete cleanly in ~20s wall.

## Validation

- Built locally and ran 8× concurrent `melange test` invocations of unrelated packages — all passed in ~23s wall time
- In a 807-package sample run at concurrency=6 against unpatched melange we measured ~3% docker-tag race false-negatives; at concurrency=16 it was ~21%. With this patch, **zero races at concurrency=8 across a real 8074-package archive sweep currently in flight**
- Validated cleanup: no orphaned `melange:*` tags remain after a successful run

## Note on independent caller-side cleanup

In addition to the per-invocation tag, callers that fan out many concurrent test runs may also want a periodic `docker image prune -a -f --filter until=5m` between batches — `apko_oci.LoadImage` tags the underlying image as `apko.local/cache:<hash>` *in addition to* the melange tag, and that cache-keyed tag persists across runs. That's a caller-side concern, not a melange bug, but worth flagging because it bit us during long-running test sweeps. See [chainguard-dev/stereo#91893](https://github.com/chainguard-dev/stereo/pull/91893) for the orchestrator that drove this discovery.

## Test plan
- [x] Manual: run 8+ concurrent `melange test` invocations of different YAMLs on a single host
- [x] Verify cleanup: confirm no orphaned `melange:*` tags remain after a successful run (`docker images | grep melange`)
- [x] Verify failure cleanup: confirm tags are removed even when test fails

### DIFF
--- a/pkg/container/docker/docker_runner.go
+++ b/pkg/container/docker/docker_runner.go
@@ -18,9 +18,12 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/rand"
+	"encoding/hex"
 	"fmt"
 	"io"
 	"os"
+	"strconv"
 
 	"go.opentelemetry.io/otel"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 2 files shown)


pkg/container/docker/docker_runner.go
```diff
@@ -441,13 +444,38 @@ func (d *dockerLoader) LoadImage(ctx context.Context, layer v1.Layer, arch apko_
...
 	if err != nil {
 		return "", err
 	}
 	return ref.String(), nil
 }
 
+// uniqueImageTag returns an image tag of the form "melange:<pid>-<16 hex chars>"
+// that is statistically unique across concurrent invocations on the same host.
+//
+// The PID prefix gives 100% uniqueness across separate melange processes on the
```
sil2100, line 466: I like the PID addition for debugging purposes, so I'd highlight that as a nice additional rationale!

sil2100 (approved):
Looks good! I was a bit worried thinking about the possible accumulation of image tags and requirement to perform a periodic `prune`. I think this is fine though. From elastic-build POV that doesn't matter, and for local builds it should only ever matter with a lot of interrupted builds or something stress-testing as mentioned in the description. Let's merge this.

pkg/container/docker/docker_runner.go
dustinkirkland: Thanks Lukasz!

PR #2535, opened 2026-05-19
chore(e2e tests): pin 7zip build to gcc-15

stevebeattie:
Wolfi recently moved to gcc-16 by default, but that causes 7zip to FTBFS. Pin to using gcc-15 as the default compiler.

Ref: https://linear.app/chainguard/issue/PSEC-1015/


stevebeattie:
This was fixed with a better approach in https://github.com/chainguard-dev/melange/pull/2524

PR #2545, opened 2026-05-26
DNM: Fix packagePurpose

Taffer:
Per https://github.com/chainguard-dev/customer-issues/issues/3564,

> Per SPDX 2.3 spec, the valid enum values use underscores: OPERATING_SYSTEM. The hyphenated OPERATING-SYSTEM is not in the allowed set.

```sh
NexusIQ CLI error (v2.11.1):
Line: 30, Column: 50, Path: $.packages[0].primaryPackagePurpose,
Error: does not have a value in the enumeration ["OTHER", "INSTALL", "ARCHIVE",
"FIRMWARE", "APPLICATION", "FRAMEWORK", "LIBRARY", "CONTAINER", "SOURCE",
"DEVICE", "OPERATING_SYSTEM", "FILE"]Bug 3: Unparseable SPDX content
File: /var/lib/db/sbom/triton-inference-server-local-cache-26.04-r0.spdx.json
```

Ran `make generate test test-e2e`, all was happy.


sil2100:
eh, this still is problematic: https://spdx.github.io/spdx-spec/v2.3/package-information/#724-primary-package-purpose-field

Taffer:
> eh, this still is problematic: https://spdx.github.io/spdx-spec/v2.3/package-information/#724-primary-package-purpose-field

Awesome.

xnox:
@Taffer @sil2100 - pretty sure we might want to change this in mono/apokaas and maybe in cyclonedx instead?

Has customer tried using cyclonedx?

Possibly can make it an apkoaas featureflag to generate underscore or dash; based on a customer desire.

Possibly can emit two stanzas one with dash and one with underscore. Most implementations will ignore one; or the other.

Taffer:
It looks like the spec uses `OPERATING-SYSTEM` and NexusIQ CLI is wanting `OPERATING_SYSTEM`, possibly because of the published schema. According to @sil2100, they were supposed to fix the schema to have `OPERATING-SYSTEM`. 🤷

PR #2555, opened 2026-06-02
git-checkout: guard branch+expected-commit against stale/diverged commits

stevebeattie:
## Problem

Some upstreams publish releases by advancing a **version-named branch** rather
than cutting tags — the branch *is* the release marker (e.g. `stable/cinc-vX.Y`,
`rNN.NN`, `MAJOR.MINOR.x`). Packages tracking those upstreams use `git-checkout`
with `branch:` + `expected-commit:` instead of `tag:`.

In that mode the pipeline's no-tag (branch) path is a silent-correctness
footgun. When both `branch` and `expected-commit` are set, `depth` defaults to
`-1` (a full clone of all branches/objects), and the pipeline does
`git reset --hard $expected-commit` to whatever object the clone contains —
with no check that the commit is actually on the named branch. So a stale or
foreign `expected-commit` (e.g. the version was bumped, moving a templated
branch, but the commit wasn't updated) silently builds the wrong source
instead of failing.

## Fix

Two complementary safeguards in the no-tag path:

1. **Default-on ancestry guard** — before the reset, require `expected-commit`
   to be reachable from the branch tip (`git merge-base --is-ancestor`). A
   commit that isn't on the branch now fails loudly. Pinning an *older* commit
   that is genuinely on the branch still works.
2. **Opt-in `expected-commit-is-branch-tip: true`** — require `expected-commit`
   to equal the branch tip exactly, for branches used as immutable release
   pointers (the branch-as-release-marker case above). Mutually exclusive with
   `tag`; keeps the clone shallow since no reset is needed.

## Tests

Added e2e positive cases for both behaviors in `git-checkout-build.yaml`; the
existing "branch with old expected" case already covers the older-on-branch
path of the ancestry guard.

## Notes

A scan of downstream package configs found several packages silently building
the wrong source under this footgun (branch bumped, commit left stale); this
guard surfaces them as loud failures.

v2: fix missing documentation for added git-checkout pipeline option.

Refs: [PSEC-1093]

### DIFF
--- a/pkg/build/pipelines/git-checkout.yaml
+++ b/pkg/build/pipelines/git-checkout.yaml
@@ -300,7 +300,7 @@ pipeline:
           if [ "$depth" = "unset" ]; then
             depth=1
             if [ -n "$branch" ] && [ -n "$expcommit" ] &&
-                [ "$expected_commit_is_branch_tip" != "true" ]; then
+                [ "$expected_commit_is_branch_tip" = "false" ]; then
               # if we're just checking out a specific commit on a branch
               # then we need to get history, otherwise it will break
               # if the expected commit is not tip of the branch.
--- a/pkg/build/pipelines/git-checkout.yaml
+++ b/pkg/build/pipelines/git-checkout.yaml
@@ -30,6 +30,16 @@ inputs:
   expected-commit:
     description: |
       The expected commit hash
+  expected-commit-is-branch-tip:
+    description: |
+      When "true", require that expected-commit exactly matches the tip of
+      the checked-out branch instead of allowing any older commit on the
+      branch.  Use this when a branch is used as an immutable release pointer
+      (e.g. version-named release branches) so that a stale expected-commit --
+      for example after bumping the version but forgetting to update the
+      commit -- fails loudly instead of silently checking out old code.
+      Only valid together with 'branch'; mutually exclusive with 'tag'.
+    default: "false"
   recurse-submodules:
     description: |
       Indicates whether --recurse-submodules should be passed to git clone.
@@ -232,11 +242,13 @@ pipeline:
           local cherry_pick="$8" sparse_paths="$9"
           local max_retries="${10:-3}" initial_backoff="${11:-2}" max_backoff="${12:-60}"
           local shallow_submodules="${13:-false}" submodule_jobs="${14:-1}"
+          local expected_commit_is_branch_tip="${15:-false}"
           msg "repo='$repo' dest='$dest' depth='$depth' branch='$branch'" \
               "tag='$tag' expcommit='$expcommit' recurse='$recurse'" \
               "sparse_paths='$sparse_paths' max_retries='$max_retries'" \
               "initial_backoff='$initial_backoff' max_backoff='$max_backoff'" \
-              "shallow_submodules='$shallow_submodules' submodule_jobs='$submodule_jobs'"
+              "shallow_submodules='$shallow_submodules' submodule_jobs='$submodule_jobs'" \
+              "expected_commit_is_branch_tip='$expected_commit_is_branch_tip'"
 
           case "$recurse" in
               true|false) :;;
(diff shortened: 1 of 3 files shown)


egibs (dismissed):

EyeCantCU (dismissed):
Everything looks great, love this

pkg/build/pipelines/git-checkout.yaml
```diff
@@ -271,10 +299,13 @@ pipeline:
 
           if [ "$depth" = "unset" ]; then
             depth=1
-            if [ -n "$branch" ] && [ -n "$expcommit" ]; then
+            if [ -n "$branch" ] && [ -n "$expcommit" ] &&
+                [ "$expected_commit_is_branch_tip" != "true" ]; then
```
xnox: should this be false? cause we do not need unlimited depth, when branch tip is expected to be the expected commit, no?

pkg/build/pipelines/git-checkout.yaml
```diff
@@ -323,10 +354,30 @@ pipeline:
...
-                      msg "expected commit $expcommit on ${branch:-HEAD}," \
-                          "got $foundcommit, performing reset"
-                      vr git reset --hard "$expcommit"
+                  if [ "$expected_commit_is_branch_tip" = "true" ]; then
+                      # the branch is being used as an immutable release
+                      # pointer, so expected-commit must be its tip exactly.
+                      fail "expected-commit $expcommit is not the tip of branch" \
+                           "${branch:-HEAD} ($foundcommit); the branch advanced" \
+                           "or the version was bumped without updating" \
+                           "expected-commit"
```
xnox, line 363: and then this block will not be needed?

pkg/build/pipelines/git-checkout.yaml
```diff
@@ -323,10 +354,30 @@ pipeline:
...
+                          msg "expected commit $expcommit on ${branch:-HEAD}," \
+                              "got $foundcommit, performing reset"
+                          vr git reset --hard "$expcommit"
+                      else
+                          fail "expected commit $expcommit is not reachable from" \
+                               "${branch:-HEAD} (tip $foundcommit); the branch may" \
+                               "have diverged from the pinned commit -- update" \
+                               "expected-commit"
+                      fi
                   else
```
xnox, line 381: i like all of this, but was expecting just the reset to be guarded.

As it should still be allowed to do full clone, with branch tip expected at the right tip. As some builds need depth to build correctly.

xnox (dismissed):

dannf (commented):
`eco-python-llvm-triton-3.6.0` might be a good regression test - it needs a specific commit w/ no branch specified

pkg/build/pipelines/git-checkout.yaml
stevebeattie: Yes, it will only pass this check if the $expected_commit_is_branch_tip is not set to true (i.e. is false) and thus will get depth set to -1 for a full clone. If $expected_commit_is_branch_tip is set to true, then (unless depth was set by the caller) the depth will retain its setting of 1.

xnox:
```suggestion
                [ "$expected_commit_is_branch_tip" = "false" ]; then
```

pkg/build/pipelines/git-checkout.yaml
stevebeattie: I think the block is still needed:

1. it gives a better message than just the git reset failing for unexplained reasons
2. as you say elsewhere, it should be possible to do a full clone or even a shallow clone of depth greater than one and still require the expected commit to be the tip of the branch; without this check, the git reset could possibly succeed because the stale/incorrect commit got pulled in anyway.

One of the annoying things is that we need to either add functionality for the e2e tests to have negative checks (prove the pipeline bailed out in certain situations) or add better pipeline test infrastructure in the go tests (but we still would want to isolate from the host environment I think for the pipeline tests to be representative).

pkg/build/pipelines/git-checkout.yaml
stevebeattie: Yeah, agreed. Will add an e2e test or two to verify this possible.

stevebeattie:
> `eco-python-llvm-triton-3.6.0` might be a good regression test - it needs a specific commit w/ no branch specified

Thanks for this, I did a successful rebuild of `eco-python-llvm-triton-3.6.0` making sure that the checkout landed on the right commit. I also added a simplified testcase in 2703d2cc6828b2e84cbe7df086fe0e2f3279bfed to make sure we didn't break anything with this change.

I also added three additional tests to the `git-checkout-build.yaml` e2e test to make sure we're handling the depth arguments correctlynwhen `expected-commit-is-branch-tip` is set to true.

egibs (approved):

PR #2562, opened 2026-06-08
fix(renovate): bump git-checkout regardless of tag if there is only one

vishal-chdhry:
Currently, the bump library only bumps a git-checkout if it depends on the version of the package. This works well when there are multiple checkouts, or one checkout where the package depends on version (which is almost every case). But there are cases where there is a single checkout, but it does not depend on the version, and there is an implicit assumption that it will get updated. 

This change bumps the checkout without the dependency check if there is only one checkout.

### Functional Changes

Notes:

### SCA Changes

Notes:

### Linter

### DIFF
--- a/pkg/renovate/bump/bump.go
+++ b/pkg/renovate/bump/bump.go
@@ -149,8 +149,23 @@ func New(ctx context.Context, opts ...Option) renovate.Renovator {
 			RecurseNodes().
 			Filter(yit.WithMapValue("git-checkout"))
 
+		var gitCheckoutNodes []*yaml.Node
 		for gitCheckoutNode, ok := it(); ok; gitCheckoutNode, ok = it() {
-			if err := updateGitCheckout(ctx, rc.Configuration, gitCheckoutNode, bcfg.ExpectedCommit); err != nil {
+			gitCheckoutNodes = append(gitCheckoutNodes, gitCheckoutNode)
+		}
+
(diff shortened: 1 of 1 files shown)


justinvreeland (changes requested):
I have run this on a few packages, including the one that prompted this change and I don't see a difference between this and the existing melange bump. Can you provide a bit more detail about what the expected change is here.

vishal-chdhry:
@justinvreeland this is not used by melange, this is used by other services to update melange, the name is confusing, but the cli does not use this library

vishal-chdhry:
I figured out how to test this with `melange` cli, if the git checkout `tag` does not depend on any package version, it will not update in current melange:
```sh
❯ melange bump ./os/perl-module-scandeps.yaml 1.38 --expected-commit 123adfga
2026/06/09 20:41:53 INFO attempting to bump version to 1.38
2026/06/09 20:41:53 INFO Skipping git-checkout node as tag is not derived from package.version

❯ ./melange bump ./os/perl-module-scandeps.yaml 1.38 --expected-commit 123adfga
2026/06/09 20:37:59 INFO attempting to bump version to 1.38
2026/06/09 20:37:59 INFO processing git-checkout node
2026/06/09 20:37:59 INFO   expected-commit: 123adfga
```

justinvreeland (approved):

PR #2564, opened 2026-06-10
feat(qemu): add QEMU_NET_CIDR to override SLIRP internal network

jmeridth:
### Functional Changes

Notes:

Opt-in environment variable; default behavior unchanged. Existing builds (which do not set `QEMU_NET_CIDR`) are unaffected. The SLIRP netdev string is only modified when the env var is set.

### SCA Changes

Notes:

N/A. No SCA-related code changed.

### Linter

Notes:

N/A. No linter changes.

---

## What

Add an opt-in `QEMU_NET_CIDR` environment variable that overrides SLIRP's default internal network (`10.0.2.0/24`) by appending `net=<cidr>` to the SLIRP netdev args. Validates the input as a sane IPv4 CIDR with prefix length between 8 and 30.

## Why

SLIRP user-mode networking treats its default `10.0.2.0/24` network as part of its own NAT space. When the host needs to reach addresses on a private network that fall within the `10.0.0.0/8` range, SLIRP may not forward those connections to the host's network stack, causing the guest VM to time out.

Setting `QEMU_NET_CIDR` to a non-conflicting range (e.g., `192.168.76.0/24`) moves SLIRP's internal NAT off the `10.x.x.x` space so `10.x.x.x` addresses route through the host normally.

This was confirmed in practice: a host process can reach an internal `10.x.x.x` address (TCP/TLS connection completes), but the same connection from inside the QEMU sandbox times out after ~80 seconds. Switching SLIRP's internal CIDR to `192.168.x.x` resolves the conflict.

## Notes

- Opt-in via environment variable; default behavior unchanged.
- Follows the existing `QEMU_DNS_SEARCH` pattern for env-driven netdev configuration.
- Input validation rejects IPv6, prefixes outside `[8, 30]`, and any injection attempts via embedded commas or whitespace.

## Testing

- 11 new unit tests in `TestParseAndValidateNetCIDR` covering valid CIDRs, normalization, whitespace handling, and rejection of malformed/malicious inputs.
- Existing `TestParseDNSSearchDomains` and `TestBuildDNSSearchNetdevArgs` continue to pass.
- `go build ./...` and `go vet ./pkg/container/` pass.

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -798,6 +798,22 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 	baseargs = append(baseargs, serialArgs...)
 	// use -netdev + -device instead of -nic, as this is better supported by microvm machine type
 	netdevArgs := "user,id=id1,hostfwd=tcp:" + cfg.SSHAddress + "-:22,hostfwd=tcp:" + cfg.SSHControlAddress + "-:2223"
+	// QEMU_NET_CIDR overrides SLIRP's default internal network (10.0.2.0/24).
+	// This is necessary when the host needs to reach VPC-internal addresses
+	// that fall within the 10.0.0.0/8 range, since SLIRP treats its default
+	// network as part of its own NAT space and may not correctly forward
+	// connections to other 10.x.x.x addresses on the host's network.
+	// The value must be a valid IPv4 CIDR. SLIRP automatically assigns the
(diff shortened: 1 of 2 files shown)


jmeridth:
Commit signing fixed with https://github.com/chainguard-dev/melange/pull/2565

CI will be fixed with https://github.com/chainguard-dev/melange/pull/2566

justinvreeland (approved):

PR #2568, opened 2026-06-11
fix(qemu): skip empty QEMU_NET_CIDR instead of erroring

jmeridth:
### Functional Changes

Notes:

Fixes a regression from #2564. When `QEMU_NET_CIDR` is set to an empty string (e.g., via a build system that always sets the env var with a default empty value), `os.LookupEnv` returns `ok=true` and the validator rejects it as "empty CIDR", failing the build with `unable to start pod: invalid QEMU_NET_CIDR value "": empty CIDR`. This one-character fix adds `&& netCIDR != ""` to skip silently when empty.

### SCA Changes

Notes:

N/A. No SCA-related code changed.

### Linter

Notes:

N/A. No linter changes.

---

## What

Add `&& netCIDR != ""` to the `QEMU_NET_CIDR` env var check so empty values are silently skipped instead of causing a fatal error.

## Why

`os.LookupEnv` returns `ok=true` for empty string values. Build systems that unconditionally set `QEMU_NET_CIDR=""` as a default cause all QEMU builds to fail.

## Testing

- Existing `TestParseAndValidateNetCIDR/empty` test case passes (validator correctly rejects empty).
- The fix is at the call site: empty values never reach the validator.
- All existing tests pass.

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -806,7 +806,7 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 	// The value must be a valid IPv4 CIDR. SLIRP automatically assigns the
 	// gateway, DNS, and DHCP range based on the supplied network.
 	// Example: QEMU_NET_CIDR="192.168.76.0/24"
-	if netCIDR, ok := os.LookupEnv("QEMU_NET_CIDR"); ok {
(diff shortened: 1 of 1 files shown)


jmeridth:
I'm merging without approval. Elastic builds are failing because of this.

PR #2570, opened 2026-06-12
fix(qemu): pass SLIRP DNS address via kernel cmdline when QEMU_NET_CIDR is set

jmeridth:
## What

When QEMU_NET_CIDR overrides the SLIRP internal network, derive the correct DNS server address (network base + 3, per SLIRP convention) and pass it to the guest via the `dns=` kernel command line parameter.

## Why

Guest init scripts typically write `/etc/resolv.conf` using the `dns=` kernel param with a fallback to `10.0.2.3` (SLIRP's default). When QEMU_NET_CIDR moves SLIRP to a different subnet (e.g., `192.168.76.0/24`), `10.0.2.3` no longer exists on that network and all DNS resolution inside the QEMU VM fails. This breaks git clones, submodule fetches, and any other network operation requiring name resolution.

## Notes

- When QEMU_NET_CIDR is not set or empty, no `dns=` is added to the kernel cmdline. Zero behavior change for existing builds.
- The `dns=` parameter follows QEMU/Linux kernel cmdline conventions and is already consumed by common guest init scripts.
- `slirpDNSAddr` uses proper uint32 arithmetic to derive the address and validates the result is within the network.

## Testing

- [x] `go test ./pkg/container/ -v` -- all existing tests pass, no regressions
- [x] `go vet ./pkg/container/` -- clean
- [x] New `TestSlirpDNSAddr` covers: /24, /16, /8 networks, default SLIRP network, invalid input, empty input

### DIFF
--- a/pkg/container/qemu_runner.go
+++ b/pkg/container/qemu_runner.go
@@ -25,6 +25,7 @@ import (
 	"crypto/sha256"
 	_ "embed"
 	"encoding/base64"
+	"encoding/binary"
 	"encoding/pem"
 	"errors"
 	"fmt"
@@ -818,7 +819,7 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
 		}
 		slirpDNS = dnsIP
 		log.Infof("qemu: QEMU_NET_CIDR set to %s, overriding SLIRP default network (dns=%s)", cidr, slirpDNS)
-		netdevArgs += ",net=" + cidr
+		netdevArgs += ",net=" + cidr + ",dns=" + slirpDNS
 	}
 	// QEMU_DNS_SEARCH allows configuring DNS search domains inside the guest VM.
 	// This is useful for builds that need to resolve short hostnames via search
@@ -2595,43 +2596,33 @@ func parseAndValidateNetCIDR(input string) (string, error) {
 	if bits != 32 {
 		return "", fmt.Errorf("CIDR must be IPv4 (got %d-bit mask)", bits)
 	}
-	if ones < 8 || ones > 30 {
-		return "", fmt.Errorf("CIDR prefix length must be between 8 and 30 (got /%d)", ones)
+	if ones < 8 || ones > 28 {
+		return "", fmt.Errorf("CIDR prefix length must be between 8 and 28 (got /%d)", ones)
 	}
 
 	return ipnet.String(), nil
 }
 
-// slirpDNSAddr returns the DNS server address that SLIRP assigns for a given
-// network CIDR. SLIRP uses a fixed offset from the network base: +2 for the
-// gateway and +3 for the DNS server. For example, 192.168.76.0/24 yields
+// slirpDNSAddr returns the DNS server address to use for the SLIRP network.
+// The address is network base + 3, placed at offset .3 within the subnet.
+// The caller must also pass this address via the QEMU netdev dns= option so
+// SLIRP serves it, and via the kernel cmdline dns= parameter so the guest
+// init writes it to /etc/resolv.conf. For example, 192.168.76.0/24 yields
 // 192.168.76.3. The input must be a valid IPv4 CIDR (as returned by
 // parseAndValidateNetCIDR).
 func slirpDNSAddr(cidr string) (string, error) {
-	ip, ipnet, err := net.ParseCIDR(cidr)
+	_, ipnet, err := net.ParseCIDR(cidr)
 	if err != nil {
 		return "", fmt.Errorf("parse CIDR: %w", err)
 	}
-	ipv4 := ip.To4()
-	if ipv4 == nil {
+	base := ipnet.IP.To4()
+	if base == nil {
 		return "", fmt.Errorf("not IPv4")
 	}
 
-	// Use the network address (mask away host bits) and add 3.
-	base := ipnet.IP.To4()
-	if base == nil {
-		return "", fmt.Errorf("network address is not IPv4")
-	}
-
-	// Convert to uint32 for safe arithmetic, then add 3.
-	n := uint32(base[0])<<24 | uint32(base[1])<<16 | uint32(base[2])<<8 | uint32(base[3])
-	n += 3
-	dns := net.IPv4(byte(n>>24), byte(n>>16), byte(n>>8), byte(n))
-
-	// Verify the result is still within the network.
-	if !ipnet.Contains(dns) {
-		return "", fmt.Errorf("DNS address %s is outside network %s", dns, ipnet)
-	}
+	n := binary.BigEndian.Uint32(base) + 3
+	dns := make(net.IP, 4)
+	binary.BigEndian.PutUint32(dns, n)
 
 	return dns.String(), nil
 }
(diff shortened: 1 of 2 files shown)


pkg/container/qemu_runner.go
```diff
@@ -2589,6 +2602,40 @@ func parseAndValidateNetCIDR(input string) (string, error) {
...
+
+	// Use the network address (mask away host bits) and add 3.
+	base := ipnet.IP.To4()
+	if base == nil {
+		return "", fmt.Errorf("network address is not IPv4")
+	}
+
+	// Convert to uint32 for safe arithmetic, then add 3.
+	n := uint32(base[0])<<24 | uint32(base[1])<<16 | uint32(base[2])<<8 | uint32(base[3])
+	n += 3
```
egibs: This `base + 3` derivation only matches QEMU for /23–/30. For /8–/22 QEMU assigns `net | (0x0203 & ~mask)` (the third octet picks up `+2`), so the value written to the guest's `resolv.conf` won't match the resolver SLIRP is actually serving.

pkg/container/qemu_runner.go
```diff
@@ -806,12 +806,18 @@ func createMicroVM(ctx context.Context, cfg *Config) error {
...
 			return fmt.Errorf("invalid QEMU_NET_CIDR value %q: %w", netCIDR, err)
 		}
-		log.Infof("qemu: QEMU_NET_CIDR set to %s, overriding SLIRP default network", cidr)
+		dnsIP, err := slirpDNSAddr(cidr)
+		if err != nil {
+			return fmt.Errorf("deriving SLIRP DNS address from QEMU_NET_CIDR %q: %w", cidr, err)
+		}
+		slirpDNS = dnsIP
+		log.Infof("qemu: QEMU_NET_CIDR set to %s, overriding SLIRP default network (dns=%s)", cidr, slirpDNS)
 		netdevArgs += ",net=" + cidr
```
egibs: Consider appending `",dns=" + slirpDNS` here as well. Right now melange sets only `net=` and then independently guesses the DNS address QEMU will choose so setting `dns=` explicitly makes melange authoritative and guarantees that `netdev` and `cmdline` agree for every prefix length.

pkg/container/qemu_runner_test.go
```diff
@@ -576,6 +576,70 @@ func TestParseAndValidateNetCIDR(t *testing.T) {
...
+			cidr:     "192.168.76.0/24",
+			expected: "192.168.76.3",
+		},
+		{
+			name:     "default SLIRP network",
+			cidr:     "10.0.2.0/24",
+			expected: "10.0.2.3",
+		},
+		{
+			name:     "/16 network",
```
egibs, line 612: These expected values encode a bug rather than catching one. QEMU assigns `172.16.2.3` for `172.16.0.0/16` and `10.0.2.3` for `10.0.0.0/8`, not `172.16.0.3` / `10.0.0.3`. 

As written the test passes while the real behavior diverges, so it gives false confidence. Worth a case at the /22 boundary (`192.168.0.0/22` → `192.168.2.3`) and the /23 edge case (`192.168.0.0/23` → `192.168.0.3`) as well.

pkg/container/qemu_runner.go
```diff
@@ -2589,6 +2602,40 @@ func parseAndValidateNetCIDR(input string) (string, error) {
...
+		return "", fmt.Errorf("network address is not IPv4")
+	}
+
+	// Convert to uint32 for safe arithmetic, then add 3.
+	n := uint32(base[0])<<24 | uint32(base[1])<<16 | uint32(base[2])<<8 | uint32(base[3])
+	n += 3
+	dns := net.IPv4(byte(n>>24), byte(n>>16), byte(n>>8), byte(n))
+
+	// Verify the result is still within the network.
+	if !ipnet.Contains(dns) {
```
egibs: For any prefix ≤ /30 (which `parseAndValidateNetCIDR` enforces), `base+3` is always within the network and for /30 it's the broadcast address, which `Contains` still reports as inside.

pkg/container/qemu_runner.go
egibs: The /30 lower bound is loose for SLIRP since gateway/DNS/DHCP sit at offsets .2/.3/.15–.31, so a /30 has no room for the DHCP range and `slirpDNSAddr` returns the /30 broadcast (.3) as the DNS. 

Anything narrower than ~/28 won't actually work. It might be worth tightening the minimum (e.g. /29 or /28) so misconfigurations fail fast rather than attempting to use broken networking.

pkg/container/qemu_runner.go
```diff
@@ -2589,6 +2602,40 @@ func parseAndValidateNetCIDR(input string) (string, error) {
...
+// network CIDR. SLIRP uses a fixed offset from the network base: +2 for the
+// gateway and +3 for the DNS server. For example, 192.168.76.0/24 yields
+// 192.168.76.3. The input must be a valid IPv4 CIDR (as returned by
+// parseAndValidateNetCIDR).
+func slirpDNSAddr(cidr string) (string, error) {
+	ip, ipnet, err := net.ParseCIDR(cidr)
+	if err != nil {
+		return "", fmt.Errorf("parse CIDR: %w", err)
+	}
+	ipv4 := ip.To4()
```
egibs: `ip`/`ipv4` are only used for a nil-check that duplicates the `base := ipnet.IP.To4()` check just below, on input that's already validated as IPv4 upstream so this can probably be dropped.

Also, `binary.BigEndian.Uint32(base)` or `PutUint32` would express the pack/unpack more legibly than the manual shifts.

pkg/container/qemu_runner.go
jmeridth: @egibs Addressed in https://github.com/chainguard-dev/melange/commit/91c291ac -- instead of trying to match SLIRP internal algorithm, we now set dns= explicitly on the netdev args so melange is authoritative about the DNS address.

pkg/container/qemu_runner.go
jmeridth: @egibs Addressed in https://github.com/chainguard-dev/melange/commit/91c291ac -- netdevArgs now includes ,dns= alongside ,net= so melange is authoritative and netdev + cmdline always agree.

pkg/container/qemu_runner_test.go
jmeridth: @egibs Addressed in https://github.com/chainguard-dev/melange/commit/91c291ac -- added /22, /23, and /28 boundary test cases. Since we now set dns= explicitly on the netdev, the test expectations reflect our chosen address (net+3) rather than QEMU defaults.

pkg/container/qemu_runner.go
jmeridth: @egibs Addressed in https://github.com/chainguard-dev/melange/commit/91c291ac -- removed the dead Contains check.

pkg/container/qemu_runner.go
jmeridth: @egibs Addressed in https://github.com/chainguard-dev/melange/commit/91c291ac -- tightened the minimum prefix from /30 to /28. SLIRP needs room for gateway (.2), DNS (.3), and DHCP range (.15+), so anything narrower than /28 cannot work.

pkg/container/qemu_runner.go
jmeridth: @egibs Addressed in https://github.com/chainguard-dev/melange/commit/91c291ac -- removed the redundant ip/ipv4 check and switched to binary.BigEndian.Uint32/PutUint32.

egibs (approved):

PR #2573, opened 2026-06-18
feat: pipelines: autoconf: Add targets input for make and make-install

dannf:
# What
Allow overriding the `Makefile` target used for build and/or install.

# Why
`install` isn't always the appropriate install target. I'm packaging an `openssl` variant that wants to use the `install_sw` and `install_ssldirs` targets. I added it to both `make` and `make-install` for symmetry.

# How is this validated?
Several existing `melange` tests use these pipelines, so CI has regression tested them. I've locally tested the feature in a new config (https://github.com/chainguard-dev/stereo/pull/121759)

# Note
This supports multiple space-separated options. You can of course just call the pipeline multiple times instead, but giving `make` multiple targets
at once may allow it to do more in parallel.


EyeCantCU (approved):
I like this, but I do wonder if we shouldn't just consolidate both

PR #2574, opened 2026-06-19
feat: pipelines: split/dev: Add prefix input

dannf:
# What
Support splitting dev files from package installations under prefixes other than `/usr`.

# Why
We frequently need to package custom builds of libraries to be consumed by builds of python libraries. To avoid conflicts with existing Wolfi packages, we use a prefix other than `/usr`.

# Validation
The melange e2e tests include packages that use `split/dev`, so we have some coverage there already. In addition, I selected 32 random packages from `stereo` that use `split/dev` and I did before/after local builds, then compared the APK contents using the following script:

```bash
#!/usr/bin/bash
find stereo stereo.tmp -name '*.apk' | while read apk; do
  tar tfz $apk 2>/dev/null | sort -u > $apk.lst
done
find stereo -name '*.apk.lst' | while read f; do
  diff -u $f "$(echo $f | sed 's/^stereo/stereo.tmp/')"
done
```
1 package (xdpyinfo) ftbfs for both tests. The output of the above script for the other 31 showed no differences.

I confirmed that one of the packages (py3-torch-tensorrt-cuda-13.2) uses the `split/dev` "package" input, just to make sure that case was covered.

### DIFF
--- a/pkg/build/pipelines/split/dev.yaml
+++ b/pkg/build/pipelines/split/dev.yaml
@@ -9,6 +9,10 @@ inputs:
     description: |
       The package to split development files from
     required: false
+  prefix:
+    description: |
+      The package's installation prefix
+    default: "/usr"
 
 pipeline:
   - runs: |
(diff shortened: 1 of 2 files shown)


pkg/build/pipelines/split/dev.yaml
```diff
@@ -24,15 +28,23 @@ pipeline:
...
-        usr/bin/*-config usr/bin/*_config usr/share/vala/vapi \
-        usr/share/gir-[0-9]* usr/share/qt*/mkspecs \
-        usr/lib/qt*/mkspecs \
-        usr/lib/cmake usr/share/cmake \
-        usr/lib/glade/modules usr/share/glade/catalogs \
+      prefix="${{inputs.prefix}}"
+      # trim leading /
+      while [ "${prefix#/}" != "$prefix" ]; do
+        prefix="${prefix#/}"
+      done
```
EyeCantCU, line 35: This while loop looks redundant

EyeCantCU (commented):
One comment, otherwise looks great

pkg/build/pipelines/split/dev.yaml
dannf: It handles the case of multiple leading `/`. I thought about ditching it because that seems like an unlikely corner case, but I could see it *maybe* happening if `prefix: ` was set to a variable. I could clarify the comment to explain it better, but also happy to kill it if you prefer.

EyeCantCU: Makes sense to me

EyeCantCU (approved):

PR #2577, opened 2026-06-22
fix(build): make MutateWith deterministic for forwarded inputs

toabctl:
## Problem

Nesting `uses:` and forwarding a same-named input down — `A` uses `B`, `B` uses `C` with `{ foo: ${{inputs.foo}} }`, where `C` also declares input `foo` — substitutes **nondeterministically** (depends on Go's randomized map iteration). Two intermittent symptoms:

1. **Compile crash:** `${{inputs.foo}}` is sometimes left unresolved; in a `runs:` block `stripComments` then fails with `invalid parameter name`.
2. **Silent wrong value:** a literal override (`{ foo: secret }`) sometimes resolves to the inherited parent value instead of `secret`, with no error.

(Hit in practice by 2-level test bring-up pipelines forwarding `admin-password` / `tls-cert-dir`.)

## Root cause

`compilePipeline` merges the parent's resolved substitution map into the child's `with`, so the merged map holds two entries that collide on one key in `MutateWith`:

```
"${{inputs.foo}}" -> "<parent value>"   (inherited)
"foo"             -> "${{inputs.foo}}"  (child's forward)
```

The single insert loop writes both to `nw["${{inputs.foo}}"]`; `for range` over a map is randomized, so the winner is nondeterministic. When the self-reference wins it resolves to itself and stays unresolved.

## Fix

Make the merge deterministic and ordered: seed the inherited `${{...}}` entries, then apply this pipeline's own inputs (names sorted) resolved against the parent scope — so a forwarded self-reference takes the parent value and a literal override always wins.

## Tests

Each repeats to defeat random map order; all fail on `main`, pass here:
- `Test_MutateWith_ForwardedSameNamedInput` — self-ref, override, default-wins, no-collision, sibling-ref, multi-key.
- `Test_MutateWith_DeepForwarding` — 6-level forwarding chain.
- `Test_CompilePipelines_NestedForwardedInput` — end-to-end 3-level nested `uses:` reproducing the `stripComments` failure.

## Fix site (for reviewers)

The collision originates where `compilePipeline` merges the parent map into the child `with`; I fixed it in `MutateWith`. Alternatives: resolve forwarded `with` values against the parent before merging, or don't let an inherited `${{inputs.X}}` shadow a declared child input. Happy to move it.

### DIFF
--- a/pkg/build/pipeline.go
+++ b/pkg/build/pipeline.go
@@ -40,14 +40,25 @@ const WorkDir = "/home/build"
 func (sm *SubstitutionMap) MutateWith(with map[string]string) (map[string]string, error) {
 	nw := maps.Clone(sm.Substitutions)
 
+	// Seed the parent's inherited "${{...}}" entries first, then this pipeline's
+	// own plain inputs, resolving each against the parent so a forwarded self-ref
+	// (`foo: ${{inputs.foo}}`) takes the parent value and the input then wins.
 	for k, v := range with {
-		// already mutated?
 		if strings.HasPrefix(k, "${{") {
 			nw[k] = v
-		} else {
-			nk := fmt.Sprintf("${{inputs.%s}}", k)
-			nw[nk] = v
 		}
+	}
+	for k, v := range with {
+		if strings.HasPrefix(k, "${{") {
+			continue
+		}
+		nk := fmt.Sprintf("${{inputs.%s}}", k)
+		// Resolve against the parent scope; keep raw if not yet in scope (a
+		// sibling input) — the mutation loop below resolves it.
+		if rv, err := util.MutateStringFromMap(nw, v); err == nil {
+			v = rv
+		}
+		nw[nk] = v
 	}
 
 	// do the actual mutations
--- a/pkg/build/pipeline_test.go
+++ b/pkg/build/pipeline_test.go
@@ -15,6 +15,7 @@
 package build
 
 import (
+	"maps"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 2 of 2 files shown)


addyess:
Took a closer look at the `sort.Strings(inputKeys)` block and wanted to share what I found, though I'm not fully confident this is the whole picture — would love a sanity check.

I tried checking out the branch and **removing the sort entirely** (swapped the sorted-slice + second filter loop back to a plain `for k := range with`), then ran the new tests. Since they already repeat internally (1000× / 200×) to shake out random map order, I also bumped `-count=50`. Everything still passed, and the full `pkg/build` suite stayed green.

That made me wonder whether the sort is actually contributing to the determinism, or whether the real fix is the other two changes:

- **Distinct input names can't collide** — each plain `k` writes a unique `${{inputs.k}}` target, so processing order among them doesn't change the result.
- **Sibling/cross refs look like they're normalized by the final `for k, v := range nw` mutation loop** regardless of the order the plain loop wrote them, since that loop re-resolves every entry.

The concrete case I assumed the sort was guarding is a sibling reference, where ordering *feels* like it should matter:

```yaml
with:
  a: ${{inputs.b}}   # references a sibling input
  b: bee
```

…but this is already the `"sibling reference resolves regardless of order"` row in `Test_MutateWith_ForwardedSameNamedInput`, and it still passes with the sort removed — which is part of why I suspect the final mutation loop, not the sort, is what makes it deterministic. So I think that test already locks the behavior in, and the sort can go without losing coverage. (If anything, it might be worth a comment on that test row noting it's the order-independence guarantee.)

If that's right, the determinism seems to come from (a) seeding the inherited `${{...}}` entries first and (b) resolving each forwarded value against the parent scope before writing it — i.e. this part:

```go
if rv, err := util.MutateStringFromMap(nw, v); err == nil {
    v = rv
}
nw[nk] = v
```

…rather than the sort. But it's very possible I'm missing a case the sort is guarding — if you've got a repro in mind where order *does* change the output, that'd be great to know and I'll happily eat my words.

Separately (and definitely out of scope for this PR): that final mutation loop is itself a single random-order, single-level-resolution pass, so deeper indirection chains *within one* `MutateWith` call — e.g. `a: ${{inputs.b}}`, `b: ${{inputs.c}}`, `c: see` — might still be order-sensitive, and the sort wouldn't fix that either. The forwarding tests sidestep it because each nesting level is its own `MutateWith` call. Just flagging in case it's relevant later; happy to write a failing test for it if useful.

toabctl:
Good catch — you're right, removed the sort (force-pushed); it wasn't load-bearing.

The fix is the two loops: seed inherited `${{...}}` first, then resolve each input against the parent before writing — so a forwarded `foo: ${{inputs.foo}}` takes the parent value and a literal override wins. Reported cases have a single/independent plain key, so order doesn't matter; sibling refs are made order-independent by the final mutation loop (added a comment), not the sort.

Verified after removal: `-count=50` green, golangci-lint clean, functional test still fails on `main` / passes here.

Agreed the deep `a→b→c` chain within one `MutateWith` is a separate pre-existing issue, out of scope — happy to follow up, and a failing test from you would be a great start.

addyess (approved):

PR #2607, opened 2026-08-03
feat(pipelines): auto-inject pipeline-declared capabilities; xcover requests CAP_SYS_ADMIN

AnmolVirdi:
### Functional Changes

## Summary
  
  This PR adds the ability for a pipeline to declare the Linux capabilities it requires, so that those capabilities are automatically merged into the runner whenever the pipeline is used. It also wires this mechanism into the built-in `xcover/profile` pipeline, which now automatically requests `CAP_SYS_ADMIN` because xcover attaches BPF uprobes to the profiled interpreter.

  ## Motivation

  Previously, capabilities could only be declared at the top-level `capabilities` block of a melange manifest, which meant every package author using xcover had to remember to manually add `CAP_SYS_ADMIN` or the profiler would fail at runtime. This change moves that requirement into the pipeline definition itself, following the same pattern already used by `needs.packages`, so the capability is granted transparently and cannot be forgotten.

  ## Changes

  - Added a `Capabilities` field to the `Needs` struct in `pkg/config/config.go`, allowing a pipeline to declare capabilities via `needs.capabilities.add` / `needs.capabilities.drop`.
  - Extended the `Compiled` struct and `gatherDeps` in `pkg/build/compile.go` to accumulate capabilities declared by pipelines as they are walked.
  - Added a `mergeCapabilities` helper and wired it into every build and test compile path so gathered capabilities are merged into the runner configuration.
  - Declared `needs.capabilities.add: [CAP_SYS_ADMIN]` on `pkg/build/pipelines/xcover/profile.yaml`.
  - Regenerated `pkg/config/schema.json` and updated `pkg/config/schema.cue` to reflect the new field.

  ## Behavior

  Manifest-declared capabilities and pipeline-declared capabilities are combined as a deduplicated union rather than one overwriting the other. If a manifest declares `CAP_NET_ADMIN` and a test uses `xcover/profile`, the runner receives both `CAP_NET_ADMIN` and `CAP_SYS_ADMIN`, and duplicates across the two sources are collapsed to a single entry. Because `Configuration.Capabilities` is shared by both the build and test runners, the union applies to whichever runner executes the pipeline; bubblewrap cannot scope a capability to an individual pipeline step.
  
  ## Testing

  Added `TestCompileCapabilities` in `pkg/build/compile_test.go`, which covers both the `Test.Compile` and `Build.Compile` paths and asserts that manifest and pipeline capabilities are merged into a deduplicated union without overwriting each other. `go build ./...` succeeds, and the full `pkg/build` and `pkg/config` test suites pass.


toabctl (commented):
Nice — the mechanism is the right shape and mirrors `needs.packages`. Two asks:

**Don't merge test-pipeline caps into the build container.** `Build.Compile` merges caps from `test:` pipelines into `Configuration.Capabilities`, which feeds the build runner (`build.go:1048`) — but `melange build` never executes test pipelines, so `uses: xcover/profile` under `test:` grants `CAP_SYS_ADMIN` to the container running upstream `configure`/`make` for nothing. Fix: keep the `c.Capabilities` merge, delete the two `tc.Capabilities` merges. `Test.Compile` is already correct — xcover keeps working under `melange test`.

**Warn on add/drop conflicts.** If one source adds a cap and another drops it, melange emits both and each runner arbitrates differently (under bubblewrap the drop silently wins — flag order; docker/qemu decide downstream). Please warn during compile when the merged `Add`/`Drop` sets intersect.

Smaller:
- Log gathered caps in `gatherDeps` (Info, with pipeline id) — packages are logged, sandbox widening should be too.
- Validate cap names at compile time; a typo currently fails at `bwrap --cap-add` runtime.
- `docs/BUILD-PROCESS.md` describes `needs:` as packages-only.
- Nit: `mergeCapabilities` skips dedup when `src` is empty, contradicting its comment.

maxgio92 (commented):
The mechanism looks right and mirrors needs.packages nicely; my main concern is scoping, details inline. These are meant to add to toabctl's review, not repeat it.

Just a detail (non-blocking): the new Needs block in schema.json matches generator output, but a full go generate at this head also emits the apko runtime_keyring definitions, so a future clean regen will produce an unrelated-looking diff.

pkg/build/compile.go
```diff
@@ -111,6 +111,9 @@ func (t *Test) Compile(ctx context.Context) error {
 
 		// Sort and remove duplicates.
 		te.Packages = slices.Compact(slices.Sorted(slices.Values(te.Packages)))
+
+		// Merge any capabilities this subpackage test needs into the runner.
+		mergeCapabilities(&t.Configuration.Capabilities, test.Capabilities)
```
maxgio92: This merges each subpackage test's caps into the shared t.Configuration.Capabilities, but buildWorkspaceConfig applies that to every test container, including the fresh per-subpackage ones created to keep tests isolated (test.go:352-371). One subpackage using xcover/profile would grant CAP_SYS_ADMIN to all sibling test containers. Could the caps stay on the per-test Compiled and be threaded into the corresponding buildWorkspaceConfig call (container.Config already has a Capabilities field), the way te.Packages is scoped today?

pkg/build/compile_test.go
```diff
@@ -121,6 +121,74 @@ func TestCompileTest(t *testing.T) {
 	}
 }
 
+func TestCompileCapabilities(t *testing.T) {
```
maxgio92, line 127: This asserts the caps land on the global Configuration.Capabilities, so it would keep passing with the scoping issues unfixed. Worth adding a case asserting that caps declared only under test: stay out of the build configuration, and one that loads a pipeline via uses: from a testdata dir, so the yaml round-trip of needs.capabilities and the pipeline.Needs = nil interplay in gatherDeps get exercised (here the Needs structs are injected directly).

pkg/config/config.go
```diff
@@ -580,6 +580,9 @@ func (p Package) FullCopyright() string {
 type Needs struct {
 	// A list of packages needed by this pipeline
 	Packages []string
+	// Optional: Linux capabilities needed by this pipeline. These are merged
+	// into the runner's capabilities configuration whenever the pipeline is used.
+	Capabilities Capabilities `json:"capabilities,omitzero" yaml:"capabilities,omitempty"`
```
maxgio92: Question: could this be add-only until a drop use case exists? "This pipeline needs CAP_X dropped" reads as hardening rather than a dependency, and since drops union globally, one pipeline's drop strips the capability from every other step sharing the container (under bubblewrap the drop wins over an add). xcover only needs add, and add-only would also dissolve the add/drop conflict question.

pkg/config/schema.cue
```diff
@@ -281,6 +281,11 @@
 #Needs: close({
 	// A list of packages needed by this pipeline
 	Packages!: [...string]
```
maxgio92: Just a detail (non-blocking): Packages stays required inside needs, so a capabilities-only pipeline has to write packages: []. This PR is what first makes a packages-less needs block meaningful; relaxing it could be a follow-up (the Go field would need omitzero to drop out of required).

pkg/build/compile.go
```diff
@@ -370,6 +386,17 @@ func (c *Compiled) compilePipeline(ctx context.Context, sm *SubstitutionMap, pip
 	return nil
 }
 
+// mergeCapabilities merges the capabilities gathered from pipelines into the
```
maxgio92: Just a detail (non-blocking): the five "Merge any capabilities..." call-site comments restate the call below them; one is probably enough.

AnmolVirdi:
@toabctl, thanks for the suggestions. I've address all concerns:
- Build container no longer gets `test-pipeline` caps: I dropped the two `tc.Capabilities` merges in `Build.Compile` and kept the build-pipeline one, so `Test.Compile` still applies them under melange test, plus a new regression test guards against leakage.
- `mergeCapabilities` now warns at compile time whenever the merged Add and Drop sets intersect.
- `gatherDeps` logs gathered caps at Info with the pipeline id, since widening the sandbox deserves more visibility than the package lines.
- `docs/BUILD-PROCESS.md` now documents needs.capabilities in both the example and the build-process steps, including the build vs test runner distinction.
- The dedup nit is fixed: `mergeCapabilities` dedups unconditionally now and the comment matches.

On compile-time cap-name validation, I left it out of this PR on purpose: there's no capability list in our deps to check against, so it would mean hardcoding and maintaining the kernel set (which drifts as caps like `CAP_BPF` and `CAP_PERFMON` get added), and validating `needs.capabilities` while leaving the existing top-level capabilities: block unchecked would be inconsistent. Happy to do a follow-up that validates all sources against a shared list if you'd like.

maxgio92:
@AnmolVirdi thanks for b61c13e, the build-side fix and the negative test look good. One scoping gap remains under melange test: subpackage test caps still merge into the shared t.Configuration.Capabilities, so one subpackage's CAP_SYS_ADMIN reaches every sibling test container. Scoping them per-test Compiled into the corresponding buildWorkspaceConfig call (container.Config already has a Capabilities field) would close it, the way te.Packages is scoped today.

pkg/config/config.go
maxgio92: Closing as won't-fix: the compile-time conflict warning covers the practical risk.

pkg/build/compile.go
maxgio92: Closing as won't-fix.

pkg/build/compile.go
AnmolVirdi: caps now land on the per-test `Test.Capabilities` and are threaded into the matching `buildWorkspaceConfig` call, so a subpackage's `xcover/profile` no longer widens sibling test containers.

pkg/config/schema.cue
AnmolVirdi: Done here rather than as a follow-up: `Packages` now has `omitzero` and drops out of required, so a capabilities-only needs: block no longer needs `packages: []`.

AnmolVirdi:
Thanks for the follow-up review folks. I've update the PR. 

@toabctl
- All five addressed: test-pipeline caps no longer reach the build runner (and are now scoped per-test rather than shared), compile warns on add/drop intersections, `gatherDeps` logs gathered caps at Info with the pipeline id, `CAP_*` names are validated at compile time against `linux/capability.h`, `docs/BUILD-PROCESS.md` documents `needs.capabilities`, and the `dedup-on-empty` comment is fixed.

@maxgio92
- caps are scoped to each test's runner instead of the shared `Configuration.Capabilities`, `needs.capabilities` is add-only, `Packages` is no longer required inside `needs`, and I kept the unrelated apko `runtime_keyring` defs out of `schema.json` to avoid widening the diff.

pkg/build/compile_test.go
maxgio92: Verified: the scoping cases and the uses: round-trip test are in compile_test.go now. This covers what I asked for.

maxgio92 (dismissed):
Thanks Anmol! The per-test capability scoping is exactly right, and the new compile_test.go cases pin it well. 🚀

pkg/config/config.go
```diff
@@ -585,8 +585,52 @@ func (p Package) FullCopyright() string {
...
 type Needs struct {
-	// A list of packages needed by this pipeline
-	Packages []string
+	// Optional: A list of packages needed by this pipeline
+	Packages []string `json:",omitzero"`
+	// Optional: Linux capabilities needed by this pipeline. They are added to
+	// the runner of the test or build that uses the pipeline. Only additions
+	// are supported: a pipeline declares what it needs, it cannot drop a
+	// capability from the steps that share its container.
+	Capabilities NeedsCapabilities `json:"capabilities,omitzero" yaml:"capabilities,omitempty"`
```
toabctl, line 594: This field is dropped during parsing: `replaceNeeds` (`config.go:1408`) returns `&Needs{Packages: replaceAll(r, in.Packages)}` and never copies `Capabilities`, so inline `needs.capabilities` in a manifest never survives `ParseConfiguration`.

```go
return &Needs{
	Packages:     replaceAll(r, in.Packages),
	Capabilities: in.Capabilities,
}
```

(Use `NeedsCapabilities{Add: replaceAll(r, in.Capabilities.Add)}` instead if you want `${{vars.*}}` to be expandable in capability names — I'd say no, they're fixed literals, but your call.)

pkg/config/config.go
```diff
@@ -945,6 +989,12 @@ type Test struct {
 	// Optional: Additional Environment the test needs to run
 	Environment apko_types.ImageConfiguration `json:"environment" yaml:"environment,omitempty"`
 
+	// Optional: Linux capabilities to apply to this test's runner. Capabilities
+	// required by the test pipelines are gathered here during compilation, so
+	// they are scoped to this test's container rather than shared across every
+	// test in the configuration.
+	Capabilities Capabilities `json:"capabilities,omitzero" yaml:"capabilities,omitempty"`
```
toabctl, line 996: Same root cause as the `Needs.Capabilities` comment above: `replaceTest` (`config.go:1495`) rebuilds `&Test{Environment: ..., Pipeline: ...}` and drops `Capabilities`, so a hand-written `test.capabilities` is silently discarded. Reachable from both `test:` and `subpackages[].test:`.

Add `Capabilities: in.Capabilities` there. It works today only because `Test.Compile` writes the field after parsing.

pkg/build/compile.go
```diff
@@ -188,11 +194,20 @@ func (b *Build) Compile(ctx context.Context, opts ...CompileOption) error {
 
 		// Sort and remove duplicates.
 		te.Packages = slices.Compact(slices.Sorted(slices.Values(te.Packages)))
+
+		// Capabilities from test pipelines are intentionally not applied here:
```
toabctl: The reasoning here is right for `melange build`, but dropping the gathered caps entirely means they can't be recovered from a compiled configuration. `gatherDeps` sets `pipeline.Needs = nil` as it walks, so after `melange compile config.yaml > compiled.yaml`, running `melange test compiled.yaml` gets no `CAP_SYS_ADMIN` and no diagnostic — same for the `.melange.yaml` embedded in the APK (`pkg/build/package.go:231`). Contrast `needs.packages`, which is folded into `test.environment.contents.packages` and therefore round-trips.

Two options: fold them into `Test.Capabilities` here (mirroring what `Test.Compile` does), or leave `Needs` intact on test pipelines so the requirement is still declared in the output.

toabctl (changes requested):
Nice feature — declaring capabilities next to the pipeline that needs them is the right shape, and I like that test caps are scoped per-test container rather than unioned into the build runner.

I verified the shipped path works end to end: with `uses: xcover/profile`, `CAP_SYS_ADMIN` reaches both the build runner (`Build.Compile`) and each test runner including subpackage tests (`Test.Compile`), after going through `ParseConfiguration`. So the xcover fix itself is good.

The problem is the *other* two entry points the PR documents. `ParseConfiguration` rebuilds pipelines and tests field-by-field (`config.go:1785`, `config.go:1799`, unconditionally — not gated on whether any substitutions exist), and `replaceNeeds`/`replaceTest` don't copy the new fields. Parsing this manifest:

```yaml
capabilities:
  add: [CAP_TOP_LEVEL]
pipeline:
  - needs:
      packages: [wget]
      capabilities:
        add: [CAP_SYS_ADMIN]
    runs: echo hi
test:
  capabilities:
    add: [CAP_TEST_LEVEL]
  pipeline:
    - needs:
        capabilities:
          add: [CAP_TEST_PIPELINE]
      runs: echo test
```

gives:

```
top-level capabilities:      {Add:[CAP_TOP_LEVEL] Drop:[]}   <- survives
pipeline[0].Needs:           &{Packages:[wget] Capabilities:{Add:[]}}
test.Capabilities:           {Add:[] Drop:[]}
test.pipeline[0].Needs:      &{Packages:[] Capabilities:{Add:[]}}
```

`needs.packages` survives, every capability is gone. Same for `subpackages[].pipeline[].needs.capabilities` and `subpackages[].test.capabilities`. No error, no warning — `KnownFields(true)` accepts the keys because the fields exist, they're just dropped afterwards.

Two consequences:

1. The inline form that `docs/BUILD-PROCESS.md` introduces the feature with is a no-op, as is `test.capabilities` from `schema.json`/`schema.cue`.
2. The doc line "Names are checked while the pipeline is compiled, so a misspelled `CAP_*` fails the build rather than the container" doesn't hold for the inline form — `needs.capabilities.add: [CAP_TOTALLY_NOT_REAL]` parses clean and `Compile` returns `nil`, because the value is discarded before `knownRunnerCapabilities` validation ever sees it.

No regression risk (both fields are new in this PR), so this is "documented field is dead on arrival" rather than "something broke." Still worth fixing here — shipping the docs without the fix is worse than shipping neither.

The reason CI is green on this is that `TestCompileCapabilities` builds `config.Configuration` literals in Go and never goes through `ParseConfiguration`, so it only exercises the one path that happens to work. A parse-based case would have caught all of it.

Separately, worth considering: these `replaceX` functions returning a freshly-constructed struct means every future field added to `Needs`/`Test`/`Pipeline` is silently dropped until someone remembers to update them. `out := *in` followed by overriding just the fields that need substitution fails safe instead.

pkg/config/config.go
AnmolVirdi: Fixed: `replaceNeeds` now does `out := *in` and only overrides `Packages`, so `Capabilities` carries through; kept the names as fixed literals rather than substitutable, per your call.

pkg/config/config.go
AnmolVirdi: Fixed the same way: `replaceTest` (and `replacePipeline`) now copy the input and override only the substituted fields, so `test.capabilities` survives parsing instead of relying on `Test.Compile`.

pkg/build/compile.go
AnmolVirdi: Took option one: `Build.Compile` now folds the gathered caps into `Test.Capabilities` for both the main and subpackage tests, so `melange test` on a compiled config or the embedded `.melange.yaml` still gets them, while the build runner stays unchanged.

AnmolVirdi:
@toabctl thanks, all three are fixed in a80853bd.

Added parse-based coverage: `Test_capabilitiesSurviveParsing` in `pkg/config` pins all four inline sites, and a new `TestCompileCapabilities` subtest goes manifest -> `ParseConfiguration` -> `Compile`, including a misspelled inline `CAP_*` now failing the build as documented.

On your last point: `replaceNeeds`, `replaceTest` and `replacePipeline` all use `out := *in` plus overrides now, so a field added later carries through instead of being silently dropped.

AnmolVirdi:
The check failure `acid-sandbox` is not linked to this. Fix is being tracked here:
- https://github.com/chainguard-dev/mono/pull/56584

toabctl (approved):

AnmolVirdi:
Thankss @toabctl @maxgio92!

PR #2608, opened 2026-08-03
feat(build): cache `uses:` pipeline definitions; allow script normalization to be skipped

egibs:
Compiling reads and parses a pipeline definition behind every `uses:` step. This is fine for a single package [definition] but wasteful in scenarios which compile an entire repository of package definitions solely to work out its dependency graph.

Definitions can number in the hundreds or thousands and each is read and parsed once per package. `usesDocument` now keeps the parsed YAML document, keyed by the pipeline directories and the `uses` name, with the file read behind the same cache so a also hit costs no syscalls.

The document is cached as opposed to the compiled pipeline. Downstream consumers decode it onto their own pipeline, which then applies the keys the definition sets and leaves the rest alone, exactly as `yaml.Unmarshal` onto an existing value does. A step's own `with`, `if` and `workdir` need to survive loading the definition it names since a cached copy cannot distinguish between "absent" and "present and zero". 

The second change updates `Compile` to support options, and `WithDependenciesOnly()` skips `stripComments`. Dropping comments from a `runs:` body means shell-parsing and re-printing it, which is the single most expensive part of compiling a step. A caller that compiles only to learn what `.environment.contents.packages` resolves to discards that body entirely. 

Substitution, validation, conditionals and the resolved dependency list are unchanged, so a definition that fails to compile still fails the same way. Nothing in this repository passes the option and it exists explicitly for `wolfictl`'s `dag.NewPackages` and any tooling built on it, which is a one-line change there plus a version bump. The `uses` cache needs no caller change at all.

Measured over ~10,000 package definitions across several directories:

| | compile | total runtime | vs baseline |
| -- | -- | -- | -- |
| baseline | 18.54s | 22.54s | — |
| `uses` cache | 10.45s | 14.62s | 1.8× |
| `uses` cache + `WithDependenciesOnly` | 2.55s | 6.54s | 7.3× |

Both changes individually reduce runtime compared to the baseline timing. On a single directory, the cache alone takes 6.55s to 4.25s, the option alone takes it to 3.45s, and together they reach 1.40s. `Compile` is no longer the dominant term and at 2.55s it now sits below `config.ParseConfiguration`.

A local test compiled all ~10,000 definitions and verified each resulting configuration. The `uses` cache produces zero differences against the baseline, so nothing a package records changes, and `WithDependenciesOnly` produces zero differences in resolved dependency lists against a full compile, which is all its callers read. 

Cached documents are shared across goroutines, so `TestUsesDocumentConcurrent` decodes one entry from 32 at once, and the whole corpus was compiled at `GOMAXPROCS=$(nproc)` under `-race` with no races. A load failure is not cached, so an unreadable definition is retried rather than remembered as empty; an empty definition is cached.

### DIFF
--- a/pkg/build/compile.go
+++ b/pkg/build/compile.go
@@ -25,7 +25,6 @@ import (
 	"strings"
 
 	"github.com/chainguard-dev/clog"
-	"gopkg.in/yaml.v3"
 	"mvdan.cc/sh/v3/syntax"
 
 	"chainguard.dev/melange/pkg/cond"
@@ -34,6 +33,27 @@ import (
 )
 
 const unidentifiablePipeline = "???"
+
+// CompileOption adjusts what a Compile produces.
+type CompileOption func(*Compiled)
+
+// WithDependenciesOnly narrows compilation to what a caller resolving build
+// dependencies needs, which is the `uses:` tree and the packages it pulls in.
+//
+// The compiled pipelines are not runnable: each step's `runs:` keeps the
+// comments and formatting it was written with, because normalizing it is the
+// single most expensive part of compiling and a caller that only reads
+// .environment.contents.packages discards it. Everything else — substitution,
+// validation, conditionals and the resolved dependency list — is unchanged, so
+// a definition that fails to compile still fails the same way.
+//
+// Do not use this to produce a configuration that will be built or recorded:
+// use a plain Compile for that.
+func WithDependenciesOnly() CompileOption {
+	return func(c *Compiled) {
+		c.dependenciesOnly = true
+	}
+}
 
 func (t *Test) Compile(ctx context.Context) error {
 	cfg := t.Configuration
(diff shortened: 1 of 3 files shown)


pkg/build/compile.go
```diff
@@ -35,6 +34,27 @@ import (
...
+
+// WithDependenciesOnly narrows compilation to what a caller resolving build
+// dependencies needs, which is the `uses:` tree and the packages it pulls in.
+//
+// The compiled pipelines are not runnable: each step's `runs:` keeps the
+// comments and formatting it was written with, because normalizing it is the
+// single most expensive part of compiling and a caller that only reads
+// .environment.contents.packages discards it. Everything else — substitution,
+// validation, conditionals and the resolved dependency list — is unchanged, so
+// a definition that fails to compile still fails the same way.
```
jmeridth, line 48: :robot: (low, non-blocking) This says "a definition that fails to compile still fails the same way," but that is not quite true for `runs:` shell-syntax errors. `stripComments` shell-parses and re-prints the `runs:` body, and it is the only shell-syntax validation in the compile path, so skipping it means a definition with a malformed `runs:` script (e.g. `if [ true; then fi`) compiles cleanly here but errors under a full `Compile`. The resolved dependency list is unaffected, so this is fine for the dep-graph use case, but a consumer might read this comment and assume the dep pass catches the same errors `melange build` does. Worth scoping the equivalence claim to substitution/validation/conditionals/deps and calling out that `runs:` shell-syntax errors are not surfaced here.

jmeridth (approved):
:robot: (low, non-blocking) Reviewed with three models (Opus 4.8, Sonnet 5, Fable 5). The cache design holds up: the shared parsed document is immutable and each consumer decodes onto its own pipeline (no aliasing or cross-package poisoning), the `sync.Map` backing store is concurrency-safe (verified under `-race` against a 64-goroutine parallel compile, the exact `dag.NewPackages` workload this targets), the `(dirs, uses)` key resolves the correct file, and `WithDependenciesOnly` leaves the resolved dependency list byte-for-byte identical. One non-blocking doc-accuracy note inline: the error-equivalence claim does not hold for `runs:` shell-syntax errors, which are not surfaced when `stripComments` is skipped. Approving; the comment can be a follow-up.

PR #2634, opened 2026-08-31
SCA: Fix versioned depends, cover pkg-config and ld.so, and enable them by default

xnox:

Eight commits, best reviewed one at a time:

1. `pkg/build/package_test.go` — a test showing bug A
2. `pkg/build/package.go` — the fix for bug A
3. config / sca / docs / schemas — a per-package option for versioned depends
4. `pkg/sca/sca.go` — fix for bug B, an inverted check that suppressed *all* cross-package
   versioned depends
5. `.github/workflows/wolfi-presubmit.yaml` — build `jq` in the presubmit to demonstrate it
6. `pkg/sca/sca.go` — version-depend on the dynamic linker like any other shlib
7. `pkg/sca/sca.go` — generate versioned `depends:` for pkg-config packages too
8. config / sca / ci / docs — generate them **by default**, with a `no-*` opt out

Commits 2 and 4 are what make the feature produce anything at all: before this PR, turning it on
emitted either nothing or unconstrained depends. Commits 6–8 then widen the scope and settle on
the interface — commit 8 supersedes the opt-in added in commit 3, so if you are reading commit by
commit, note that the `versioned-shlib-deps` option introduced there does not survive to the tip.

### Bug A: the constraint was stripped

`removeSelfProvidedDeps()` strips the `>=version` from `so-ver:` dependencies so it can look the
dependency name up in the map of dependencies the package provides itself. It reassigned the loop
variable to do that, then appended the truncated value, so the constraint never reached
`.PKGINFO`:

```
in:  so:libc.so.6, so-ver:libc.so.6>=2.42-r0, so:libz.so.1, so-ver:libz.so.1>=1.3.1-r6
out: so:libc.so.6, so-ver:libc.so.6,          so:libz.so.1, so-ver:libz.so.1
```

An unversioned `so-ver:` depend is satisfied by any package providing the shared library, so the
generated dependency was a no-op. No existing test used `so-ver:` dependencies, so nothing caught
it. The fix uses a separate variable for the lookup key; the filtering behaviour is unchanged.

### Bug B: versioned depends were never emitted at all

`determineShlibVersion()` only emits a versioned depend if the providing package publishes a
matching versioned `provides:`, which it checks by parsing the provided version. The predicate
returned `err != nil` — so it accepted a provided version only when parsing **failed**, the
opposite of the comment directly above it:

```go
_, err := apk.ParseVersion(providedVersion)
// If we're able to parse the version,
// then it's a valid one.
return err != nil       // ← accepts only unparseable versions
```

Every version in the wolfi index parses cleanly, so the check never matched and the function fell
through to returning an empty version. No versioned depend was generated for any library provided
by another package, even with the feature switched on — only libraries resolved via the
`@CURRENT@` path could get one, since that returns earlier.

### Widening the scope: pkg-config and the dynamic linker

**pkg-config (commit 7).** Melange already stamps every pkg-config `provides:` with the version of
the package that ships the `.pc` file, but the matching `depends:` was unversioned: a `-dev`
package built against `pc:openssl=4.0.2-r1` recorded only `pc:openssl`, so it installed happily
against an older `openssl-dev` than the headers it compiled against. That is the same build-time
coupling versioned shlib depends exist to capture, so `.pc` `Requires:` now resolve the same way
and emit `pc:openssl>=4.0.2-r1` alongside the unversioned depend. `determineShlibVersion()` is
generalized into `determineProviderVersion()` taking a `providerQuery`; the two namespaces differ
only in shape (a shlib resolves through `so:` but is versioned by `so-ver:`, while pkg-config has
one namespace where `pc:openssl=<pkgver>` does both).

**The dynamic linker (commit 6).** `determineShlibVersion()` returned early for `ld-linux-*.so`,
on the grounds that versioning it would mean rebuilding everything that links glibc whenever glibc
moves. That is what a versioned depend is *for* — the linker is part of glibc's ABI surface like
every other library it ships. The exclusion is gone. Because the linker reaches a binary through
`PT_INTERP` rather than `DT_NEEDED`, the interpreter branch of `generateSharedObjectNameDeps()`
needed the same treatment; the musl rewrite there now runs on the bare basename via `CutPrefix`,
so the resolved name is what gets version-resolved. The `provides:` side needed no change — glibc
has been publishing `so-ver:ld-linux-*.so=<version>` all along.

### The interface, and the default (commit 8)

Versioned depends were gated behind a `MELANGE_VERSIONED_SHLIB_DEPENDS` environment variable, with
the tri-state `versioned-shlib-deps` option from commit 3 able to override it either way. That was
three ways to spell one decision, one of them invisible in the build definition — a package's
`.PKGINFO` depended on how the build was invoked rather than on what the package says about itself.

Commit 8 drops the environment variable and the tri-state option. Versioned depends are generated
by default, and a package declines them the way it declines every other piece of SCA:

```yaml
options:
  no-versioned-shlib-deps: true
```

`no-versioned-shlib-deps` already existed and already meant exactly this, so there is no new option
to learn — it stops being a modifier on a feature flag and becomes the whole interface, in the same
`no-*: true` shape as `no-provides`, `no-depends` and `no-commands`.

**This flips the default, and that is the reviewable decision in this PR.** Please read the
Functional Changes notes below before approving.

While documenting it I corrected two errors in the `no-versioned-shlib-deps` docs: the option
governs the shared libraries a package links *against*, not the ones it ships — the versioned
`so-ver:` `provides` are emitted regardless — and it now also covers pkg-config depends.

### Functional Changes

Notes: **not run, and this is the checkbox that matters most for this PR.** Earlier revisions were
inert unless the feature flag was set; that is no longer true. Every package built by a melange
carrying commit 8 starts emitting `so-ver:...>=` and `pc:...>=` constraints where it previously
emitted unconstrained ones.

To size that: `no-versioned-shlib-deps` and `MELANGE_VERSIONED_SHLIB_DEPENDS` appear in **0 files**
across `stereo` (9,171 package YAMLs over `os`, `enterprise-packages` and `extra-packages`) and the
standalone `enterprise-packages` repo (2,551 YAMLs). Nothing opts out today, so nothing is shielded
from the flip — the blast radius is every package in both repos, not a subset.

The new constraint is a superset of what an unversioned depend asserts, so any APK satisfying it
would have satisfied the old one too; the risk is not unsatisfiable depends in a coherent index but
tighter rebuild coupling, and `apk` solver behaviour at scale. A full Wolfi build, or at minimum an
`apkregress` run over a representative slice, is the right gate here and I have not done it.

### SCA Changes

Notes: not done at APK scale — verified at the unit level plus two real packages in CI.

`go build ./...` and `go test ./...` are green at the tip. Commit 1 fails exactly as documented
(`so-ver:libc.so.6` vs `so-ver:libc.so.6>=2.42-r0`) and commit 2 turns `pkg/build` green; each was
checked in a detached worktree, since an uncommitted fix in the working tree would have made
commit 1 look green.

Tests at the tip:

- `TestVersionedDepsEnabled` — default-on, opt-out, explicit `false`, and an unrelated `no-*` option
- `Test_versionedShlibDepsOption` — round-trips `no-versioned-shlib-deps` through `ParseConfiguration`
- `Test_removeSelfProvidedDeps_KeepsSoVerConstraint` / `_KeepsPkgConfigConstraint` — guard the constraint
- `TestDetermineShlibVersion` — parameterized over library and provider, including a case per
  architecture for the dynamic linker provided by glibc
- `TestDeterminePkgConfigVersion`, `TestVersionedPkgConfigDeps`, `TestVendoredPkgConfig` — the
  pkg-config path, including a `.pc` file whose `Requires:` is vendored by the package itself

The APK-level signal is in CI: the presubmit builds `jq` and `subversion` and **fails the job if no
`so-ver:` depends are produced** (and, for `subversion`, no versioned `pc:` depends), so the
demonstration cannot quietly become a no-op. `jq` covers the shlib and linker paths;
`subversion-dev` is the one that carries `pc:apr-1`, `pc:serf-1`, `pc:sqlite3` and co. Since
commit 8 both packages exercise this by being built at all, so the workflow no longer threads an
environment variable through `sudo` via `env(1)` or into the QEMU step. `actionlint` is clean.

### Linter

Notes: N/A, no new linter check.

### Note on the schemas

`pkg/config/schema.json` and `schema.cue` carry only the `no-versioned-shlib-deps` description
change, hand-applied rather than regenerated. Committing the full `go generate ./internal/gen-schema`
output would have added ~145 lines of unrelated churn: the checked-in `schema.cue` predates cue
v0.17's comment rewrapping, and `schema.json` is stale against current apko (`format`,
`runtime_keyring`, `RuntimeKeyringEntry` are missing). That drift is pre-existing, is not enforced
by CI, and deserves its own regeneration commit.

### Known limitation

Subpackages do not inherit the origin package's options — `pkgFromSub()` passes `sub.Options`
through verbatim and `SCABuildInterface.Options()` returns a zero `PackageOption{}` when it is nil.
Under the old opt-in that meant repeating the option to *enable* the feature. Under commit 8 the
polarity is worse: `no-versioned-shlib-deps: true` on the origin package does **not** opt out its
subpackages, so a package that needs to decline versioned depends has to repeat the option on every
subpackage, and is silently opted back in if it forgets one.

I left that alone deliberately, because fixing it means choosing between changing inheritance for
`no-provides` / `no-depends` / `no-commands` too, or making exactly one option inherit. If reviewers
want the narrow version, the containable spot is `SCABuildInterface.Options()`, which can already
see the origin's options. If the default flip lands, this is worth doing before it ships widely.


a-crate (commented):
Mostly LGTM

> with ld-linux-x86-64.so.2 staying unversioned by design

I'm not sure about this.

1. We generate so-ver provides for this today. [:link:](https://apk.dag.dev/https/packages.wolfi.dev/os/x86_64/ld-linux-2.44-2.44-r4.apk@sha1:29c4fe1113a19ea2a3a2590a050dcf44fc3ddf70/.PKGINFO) This contradicts [this comment](https://github.com/chainguard-dev/melange/blob/93b6e111db504997870489e06ecbe978e6b3239b/pkg/sca/sca.go#L364).
2. I think that comment is wrong, if we generate `>=` so-ver dependencies there's no reason we can't do the same for ld-linux that we're doing for libc. It's roughly the same blast radius.
3. Tests should assert this, and I think they will catch the bug from point 1.

xnox:
> Mostly LGTM
> 
> > with ld-linux-x86-64.so.2 staying unversioned by design
> 
> I'm not sure about this.
> 
> 1. We generate so-ver provides for this today. [🔗](https://apk.dag.dev/https/packages.wolfi.dev/os/x86_64/ld-linux-2.44-2.44-r4.apk@sha1:29c4fe1113a19ea2a3a2590a050dcf44fc3ddf70/.PKGINFO) This contradicts [this comment](https://github.com/chainguard-dev/melange/blob/93b6e111db504997870489e06ecbe978e6b3239b/pkg/sca/sca.go#L364).
> 2. I think that comment is wrong, if we generate `>=` so-ver dependencies there's no reason we can't do the same for ld-linux that we're doing for libc. It's roughly the same blast radius.
> 3. Tests should assert this, and I think they will catch the bug from point 1.

agree!

PR #2641, opened 2026-09-02
feat(pipelines): add php/composer-install pipeline

### LINKED ISSUE #242: Add php/composer-install pipeline
```
pipeline:
  - uses: php/composer-install
    with:
      package: blah
      version: 1.2.3
```

runs

```
echo '{"require":{"blah":"1.2.3"}}` > composer.json
php composer install
```

(...or something, I don't know anything about PHP or composer. Help me @erikaheidi)

https://getcomposer.org/doc/03-cli.md#install-i

jflacker:
Closes #242.

Adds a php/composer-install pipeline, mirroring the existing npm/install and ruby/install pipelines: it takes a Composer package name and version and runs composer require to install it. Installed packages land in a single shared vendor tree at /usr/share/php/vendor, the same way python/install and ruby/install target one shared runtime location rather than a directory per package. This is the one real design decision here (Composer has no equivalent of Ruby's Gem.default_dir or Python's site-packages), documented in the pipeline itself along with its limitation: composer-installed packages across multiple apks share one autoload tree, so it assumes no conflicting versions of the same dependency between them.

The pipeline is self-sufficient for its own runtime dependency (php), matching how npm/install lists nodejs directly in its own needs.packages. Includes examples/composer-install.yaml and e2e-tests/php-composer-install-build-test.yaml, both building and testing a real, dependency-free package (psr/log).

Tested by actually extracting the built apk and inspecting its contents, not just reading the YAML. That caught a real issue: composer require necessarily creates composer.json/composer.lock in its working directory, and since this pipeline installs into a shared runtime location rather than a real project directory, those build-time manifest files were shipping inside the final apk with no purpose at runtime. Fixed by removing them after install; confirmed the autoloader still works afterward and re-extracted the apk to confirm they're gone.

Also verified failure handling directly: a composer require for a nonexistent version correctly fails the build (exit code 2) rather than being masked by the cleanup step, since melange wraps every runs: script in set -eo pipefail.

Other issues caught during iteration: composer require does not accept --no-dev (that's an install-only flag), the test environment needed busybox for /bin/sh, and php needed to be in the pipeline's own needs.packages rather than relying on the consumer to add it (it happened to resolve transitively through composer's own apk dependency, but explicit matches the established convention).

### DIFF
--- a/e2e-tests/php-composer-install-build-test.yaml
+++ b/e2e-tests/php-composer-install-build-test.yaml
@@ -1,0 +1,34 @@
+# Builds a real, dependency-free Composer package (psr/log, the PSR-3
+# logger interface) using the php/composer-install pipeline, then verifies
+# the installed autoloader can actually load and instantiate a class from
+# it.
+package:
+  name: php-psr-log
+  version: 3.0.2
+  epoch: 0
+  description: "psr/log, the PSR-3 logger interface package"
+  dependencies:
+    runtime:
+      - php
+
+environment:
+  contents:
+    packages:
+      - composer
+      - php
+
+pipeline:
+  - uses: php/composer-install
+    with:
+      package: psr/log
+      version: ${{package.version}}
+
+test:
+  environment:
+    contents:
+      packages:
+        - busybox
+        - php
+  pipeline:
+    - runs: |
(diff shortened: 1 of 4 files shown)


xnox:
Looks nice! Let me check if we have someone that can review this!

Issue #1002, opened 2024-02-08
arch-all pipelines


bgermann:
It would be nice especially for the fetching phase if one could declare arch-all pipelines.
With the current implementation, a melange pipeline is run once on every architecture that the build is run for.
That means downloading the same files over and over again.
An obvious optimisation would be the following: Having a pipeline flag (i.e., arch-all) that declares it to be independent of the build architecture that is run preferrably in the native build environment.
If the melange pipeline starts with any number of such pipelines, they could be run once and hitting the first arch-dependent pipeline the builders can be spawned. If an arch-all pipeline is run after that, it can obviously also be executed within the builder.

build.arch is a variable that should be forbidden from being used in the proposed pipelines.

PR #2645, opened 2026-09-03
feat: add noarch package architecture support

lioneloh:
## Summary

Add support for building and indexing noarch packages.

## How it works

- Detect whether `noarch` is set on package.
- Spawn one runner using host architecture.
- Runner builds package and subpackages once, then copies them into each architecture-specific output directory.
- Packages retain `noarch` architecture in `PKGINFO` and SBOM.
- Runner indexes packages separately per architecture, replacing `noarch` with corresponding concrete architecture in each `APKINDEX`.
- Linter rejects executables linked to a specific architecture.

This reproduces existing Alpine `lddtree` behavior:

https://pkgs.alpinelinux.org/package/edge/main/x86/lddtree

## Validation

- `go test ./pkg/build ./pkg/index ./pkg/config ./pkg/linter ./pkg/cli`
- `git diff --check origin/main...HEAD`

Uploaded logs from restic and gzip build from wolfi os ditribution
[melange_restic.txt](https://github.com/user-attachments/files/31788744/melange_restic.txt)
[melange_gzip.txt](https://github.com/user-attachments/files/31788745/melange_gzip.txt)

## AI assistance

PR prepared with assistance from Claude Sonnet 5 and GPT-5.6-Luna.


lioneloh:
could resolve https://github.com/chainguard-dev/melange/issues/1002

mosabua:
I approved the workflow to run and added some engineers to review in more detail. I also pinged them on our internal slack

xnox:
Can you please explain a little bit why do you need this functionality; and what do you want to achieve with it; or how do you see people going to use this?

We already have qemu runner; and it does allow to boot VM for a target arch to build for example arm64 packages on x86_64 and vice versa. And if the contents is arch independent, even with Qemu emulation it should be quick enough with existing builds.

A lot of noarch packages, often package actually arch specific data - and I worry by introducing this mechanism there is no clear way to test these things on each of the target data; nor to have correct per-arch SCA and thus per-arch resolution. We have pondered to allow noarch packages in the past - but never had a good answer to these questions.

lioneloh:
When building big java projects (spark, trino, ...)  it avoids to build multiple times the same set of jars and have a single build that we could use for multiple archs. Especially when iterating over CI builds and the build is time consuming (building on native host architecture is also lot faster than with qemu).

Moreover from a SCA perspective, we would ensure that we have no binaries dependencies to a specific architecture when building package.
For example for java, only the jdk, which is a separate package, would be architecture specific.

I agree we do not absolute need this feature as we other workaround (qemu, binfmt, ...) but they are not the fastest neither the most cost effective.

xnox:
> When building big java projects (spark, trino, ...) it avoids to build multiple times the same set of jars and have a single build that we could use for multiple archs. Especially when iterating over CI builds and the build is time consuming (building on native host architecture is also lot faster than with qemu).
> 
> Moreover from a SCA perspective, we would ensure that we have no binaries dependencies to a specific architecture when building package. For example for java, only the jdk, which is a separate package, would be architecture specific.
> 
> I agree we do not absolute need this feature as we other workaround (qemu, binfmt, ...) but they are not the fastest neither the most cost effective.

Nice. For Chainguard, we cannot take such risks when we build things such as java. As too many projects eventually have subtle optional jni paths that go untested and then crash in deployed projects. And we do use bare-metal arm64 & x86_64 instances with hardware accelerated VM capabilities in google cloud for fast native builds on both arches. Hence for production builds we do recommend to do native builds - at minimum using arm64 & x86_64 github actions runners.

If all you do is change the arch tag in .PKGINFO inside the apk..... you could just right a command to override/replace arch in the .apk as post-processing step with a regular build. Or like use apk-tools / abuild tooling (from alpinelinux gitlab) to unpack and repack an apk.

So if you can build native apk as native host arch; and rewrite it to other arches; why any changes in melange needed at all? or like do you want to tweak "melange index" command to in addition to "--arch" which limits which arch is being added to an index, to like have '--force-arch' to allow adding like arm64 .apk to an x86_64 .apk? (or any other commands like melange scan --force-arch -> to like rebuild .PKGINFO sca for a different arch) and any other subcommands?

It feels a bit like too much code to created ability to create noarch packages; but then not actually reuse them and synthesize multiple arch-specific packages; when stock melange can be used to build first arch -> and synthesize the rest statically without any arbitrary code execution.

Long term maintaince will likely be on me; and features that are unused internally at chainguard - often break and are difficult to keep working cause we are unlikely to use this.

lioneloh:
> > When building big java projects (spark, trino, ...) it avoids to build multiple times the same set of jars and have a single build that we could use for multiple archs. Especially when iterating over CI builds and the build is time consuming (building on native host architecture is also lot faster than with qemu).
> > Moreover from a SCA perspective, we would ensure that we have no binaries dependencies to a specific architecture when building package. For example for java, only the jdk, which is a separate package, would be architecture specific.
> > I agree we do not absolute need this feature as we other workaround (qemu, binfmt, ...) but they are not the fastest neither the most cost effective.
> 
> Nice. For Chainguard, we cannot take such risks when we build things such as java. As too many projects eventually have subtle optional jni paths that go untested and then crash in deployed projects. And we do use bare-metal arm64 & x86_64 instances with hardware accelerated VM capabilities in google cloud for fast native builds on both arches. Hence for production builds we do recommend to do native builds - at minimum using arm64 & x86_64 github actions runners.
> 
> If all you do is change the arch tag in .PKGINFO inside the apk..... you could just right a command to override/replace arch in the .apk as post-processing step with a regular build. Or like use apk-tools / abuild tooling (from alpinelinux gitlab) to unpack and repack an apk.
> 
> So if you can build native apk as native host arch; and rewrite it to other arches; why any changes in melange needed at all? or like do you want to tweak "melange index" command to in addition to "--arch" which limits which arch is being added to an index, to like have '--force-arch' to allow adding like arm64 .apk to an x86_64 .apk? (or any other commands like melange scan --force-arch -> to like rebuild .PKGINFO sca for a different arch) and any other subcommands?
> 
> It feels a bit like too much code to created ability to create noarch packages; but then not actually reuse them and synthesize multiple arch-specific packages; when stock melange can be used to build first arch -> and synthesize the rest statically without any arbitrary code execution.
> 
> Long term maintaince will likely be on me; and features that are unused internally at chainguard - often break and are difficult to keep working cause we are unlikely to use this.

I totally understand your point of view. Let me rephrase, based on your feedback, two new features to ease this noarch package process :

- a flag for 'melange build' to build the package based on host platform arch if set. It will check the allowed 'target-architecture' if set in melange build and fails if host is not allowed. For example '--arch host' or '--host-arch' (feel free to suggest)
This flag will simplify the noarch build (no checking before on wich arch we are, if arch is correct based on melange file, ...). It will allow simpler build for local development as well.
- a '--force-arch' for 'melange index' to override arch in '.APKINFO' and index the same package to another arch.

I still have an issue. How do my CI guess this package is a noarch kind package based on melange file ? Is there any tag or yaml snippet i could add in melange file that would allow me to set different build parameter for this package without disturbing the current beahvior of melange (to avoid any specific code implementation) ?
Do you have any idea ?

I really appreciate this technical discussion. Thanks for your feedback.

xnox:
> How do my CI guess this package is a noarch kind package based on melange file

This type of annotation/marker/field we currently do not have in our schema. Previously I work extensively on delivering similar functionality in Debian & Ubuntu, with most details captured in this specification https://wiki.ubuntu.com/MultiarchSpec#Binary_package_control_fields which has since been extended to supports all combinations of build/host/target up to and including requiring to build certain arch:all multiarch:allowed packages only on ppc64le or only on s390x. The use case there is related to your, albeit inverse - to create qemu VMs on x86_64 & arm64 to emulate ppc64le/s390x one needs arch:all data-files which are compiled firmware blobs.... that can only be compiled on ppc64le/s390x hosts. Because setting up cross-toolchains and doing cross-builds is wasteful and difficult to test. But the net effect is restricted build-hosts + unrestricted target arches.

We do have a freeform schema of annotations these days that are strictly booleans, I will ponder what we can document or allow there. Because yes, you do need declarative way to identify which things are capable of less restrictive builds - or specifically less restrictive publication.

xnox:
This did come up when trying to figure out if and how we could support wasm "architecture" too.

I will ponder this a bit more, and maybe suggest updates to the build-file spec only; and see if that would be acceptable. I hope this can be achieved with less code in the build; and just additional logic of where to spit out the .apk and which indexes to add them to.

xnox:
We may also need something like https://wiki.debian.org/BuildProfileSpec too. Which I really missed when trying to create bootstrap repos (see archived https://github.com/orgs/wolfi-dev/repositories?q=bootstrap ) as we had to maintain copies of recipes because we couldn't specify optional build-deps; optional packages; optional targets like what one can in debian to have single recipe for bootstrap; docs; nodocs; noarch; etc.

lioneloh:
I propose to close this merge request and start a new issue to fully specify the solution that would provide the two new fields (mutiarch and build-depends).
It seems this two new fields could be used to have the multiarch feature asked but also much more functionality.
In this new issue we would focus only on a specification markdown for each field and agree on it before implementation.
If ok i will proceed to MR closing and issue creation.

PR #2657, opened 2026-09-14
split/manpages: add paths and preserve-path inputs, and fix merging two man roots

dustinkirkland:
## What this adds

Two inputs to `split/manpages`, plus a bug fix that fell out of writing the test.

### `paths`

Additional manual page directories to split, whitespace separated and relative to the package root. The three standard roots are still always searched; these are searched as well.

Some software installs its manuals outside those roots, and today those pages cannot be split at all — they ship in the main package where neither a doc subpackage nor a `man` invocation will find them:

| software | where its pages land |
|---|---|
| LLVM toolchains | `usr/lib/llvm-N/share/man` |
| JDKs | `usr/lib/jvm/<jdk>/man` |
| applications installed as a venv | `opt/<app>/share/man` |

### `preserve-path`

Keeps each page at the path it was installed to, instead of relocating it under `usr/share/man`.

Relocating remains the default, and is right for most packages: it puts pages where `man(1)` looks. But it collapses every source directory onto a single destination, which makes it unusable wherever several versions of the same software are meant to coexist. Concretely, from the archive I was working in:

- Every `llvm-N` package ships a page called **`scan-build.1`** — eight packages, one name.
- **`openjdk-17` and `openjdk-21` share 27 page names** between them (`java.1`, `javac.1`, `jar.1`, …); `openjdk-8` ships 40 and `openjdk-11` 25.

Relocating those into `usr/share/man` would make their doc subpackages conflict with one another instead of installing side by side. With `preserve-path: "true"` the pages stay in their versioned tree, which is also where the software's own tooling expects them.

An unrecognised value is rejected rather than treated as false, so a caller who asks to preserve paths cannot silently get a relocation instead.

## The bug fix

The loop moved each root with `mv "$mandir"/* "$destdir/"`, into the same destination for all three roots. Two roots can each contain a `man1/`, and the second `mv` then fails:

```
mv: can't rename '.../usr/local/share/man/man1': File exists
```

So **a package shipping manual pages in both `usr/share/man` and `usr/local/share/man` could not be split at all.** No existing test populated more than one standard root, so this had never surfaced. The contents are now merged into the destination rather than moved wholesale.

I did not go looking for this — the new e2e test populates several roots at once and hit it immediately.

## Testing

`e2e-tests/split-manpages-build-test.yaml` exercises all three modes in a single build, each as its own subpackage with assertions:

| subpackage | checks |
|---|---|
| `split-manpages-relocate-test` | pages from all three standard roots end up under `usr/share/man`, the non-standard roots are **not** touched, and the sources are removed from the parent |
| `split-manpages-extra-path-test` | a root named via `paths` is relocated, and is **not** left at its original path |
| `split-manpages-preserve-path-test` | the page stays at its installed path, is **not** also present under `usr/share/man`, and is removed from the parent |

Verified locally under the qemu runner — `make test-e2e-pipelines E2E_TESTS=split-manpages-build-test.yaml` passes, and the resulting apks contain:

```
split-manpages-relocate-test       usr/share/man/man1/local.1
                                   usr/share/man/man1/standard.1
                                   usr/share/man/man8/usrman.8
split-manpages-extra-path-test     usr/share/man/man1/opt.1
split-manpages-preserve-path-test  usr/lib/llvm-99/share/man/man1/versioned.1
split-manpages-build-test          (no manual pages — all were taken)
```

`relocate-test` holding both `standard.1` and `local.1` in the same `man1/` is the merge that was previously impossible.

`go test ./pkg/build/...` passes.

One note on the test's style: the assertions are `if [ ... ]; then ... fi` rather than `[ ... ] && { ... }`. Pipelines run under `set -e`, and an `&&` compound whose test is false returns non-zero — which fails the script on exactly the path where the assertion passes. My first draft had that bug.

## Backward compatibility

Both inputs default to the current behaviour: `paths` is empty and `preserve-path` is `"false"`, so an existing `uses: split/manpages` with no `with:` block behaves as before — except that it now also succeeds in the two-roots case that previously errored.

## Why

Manual pages belong in a `-doc` subpackage. In the archive I am working through, ~10 packages need a path-preserving split (the LLVM toolchains and the JDKs) and could otherwise only be handled by hand-rolling the same shell into each manifest — which is what prompted this rather than doing it ten times.


Taffer (dismissed):
LGTM

dannf (commented):
Inline comments below. I ran the new script and the old one under busybox 1.37 `sh`/`cp`/`mv` for each of the behaviour claims.

One item has no line to attach to. `split/alldocs.yaml` still has the original three-root `mv "$mandir"/* .../usr/share/man/` loop. This PR fixes a failure in `split/manpages` that happens when two roots both have `man1/`. That failure is still present in `alldocs`. Reproduced with busybox: `usr/share/man/man1/a.1` plus `usr/local/share/man/man1/b.1` fails on the second root with `mv: can't rename '.../man1': Directory not empty`. Worth porting the same merge in this PR, or having `alldocs` reuse the `manpages` logic.

On the `paths` name: `split/lib` already has a `paths` input that is additive to its defaults. The name is consistent within `split/`. The format is not consistent. `split/lib` documents newline separation for its `paths` input and reads it with `while read -r`. A newline-separated value breaks this script. See the comment on the `for` line.

pkg/build/pipelines/split/manpages.yaml
```diff
@@ -21,14 +45,34 @@ pipeline:
...
+      case "${{inputs.preserve-path}}" in
+        true|false) ;;
+        *)
+          echo "ERROR: preserve-path must be \"true\" or \"false\", got \"${{inputs.preserve-path}}\"" && exit 1
+          ;;
+      esac
+
+      # Held relative to the package root, so one value names both the
+      # source directory and, when preserving, the destination.
+      for mandir in usr/share/man usr/local/share/man usr/man ${{inputs.paths}}; do
```
dannf: A newline anywhere in `paths` is a shell syntax error. `${{inputs.paths}}` is spliced into the script text before the shell parses it. A YAML block scalar (`paths: |`, one directory per line, the form `split/lib` documents for its own `paths` input) turns this into:

```
for mandir in usr/share/man usr/local/share/man usr/man usr/lib/llvm-18/share/man
; do
```

busybox sh reports `syntax error: unexpected ";" (expecting "do")`. Even a single entry written as a block scalar fails, because yaml.v3 keeps the trailing newline. The description says "whitespace separated", which suggests newlines are fine.

Assigning to a variable first avoids this. Expansion of a variable is word-split on newlines too:

```sh
paths="${{inputs.paths}}"
for mandir in usr/share/man usr/local/share/man usr/man $paths; do
```

That also lets the description say "newline or whitespace separated", matching `split/lib`.

pkg/build/pipelines/split/manpages.yaml
```diff
@@ -21,14 +45,34 @@ pipeline:
...
+        [ -d "$PACKAGE_DIR/$mandir" ] || continue
```
dannf: Three input-handling gaps on `paths` entries, all reproduced:

1. `paths: ../bar` with the default `preserve-path` passes this check. It copies the whole of `melange-out/bar` into `foo-doc/usr/share/man/`, then runs `rm -rf` on `melange-out/bar`. Exit code is 0.
2. A glob such as `usr/lib/llvm-*/share/man` is expanded on the `for` line relative to the shell CWD. The CWD is `/home/build` (the source tree), not `$PACKAGE_DIR`. The glob either expands against the wrong tree or stays literal, and this check then skips it. The versioned root is left unsplit. Exit code is 0.
3. A typo (`usr/lib/lvm-17/share/man`) is skipped here with no message. A package whose pages live only at the custom path then gets an empty `-doc` subpackage.

Suggest rejecting entries that are absolute or contain `..`, and failing, or at least warning, when an explicitly listed path does not exist. `|| continue` makes sense only for the three implicit defaults.

pkg/build/pipelines/split/manpages.yaml
```diff
@@ -21,14 +45,34 @@ pipeline:
...
 
-        if [ -d "$mandir" ]; then
-          mkdir -p "${{targets.contextdir}}/usr/share/man"
-          mv "$mandir"/* "${{targets.contextdir}}/usr/share/man/"
-          rmdir --parents --ignore-fail-on-non-empty "$mandir"
+        if [ "${{inputs.preserve-path}}" = "true" ]; then
```
dannf: `preserve-path` applies to the three standard roots too, not only to `paths` entries. With `preserve-path: "true"`, pages under `usr/local/share/man` stay at `usr/local/...` in the `-doc` subpackage. The `usrlocal` linter flags this by default (`Warn`, "This package should be a -compat package"). Also, `usr/man` pages stay outside the directories `man` searches by default. Before this PR both were always relocated.

The JDK case in the description is a realistic trigger: a build with `prefix=/usr/local` and pages under `usr/lib/jvm/<jdk>/man`, with `preserve-path` set to protect the versioned tree.

Scoping `preserve-path` to the `paths` entries, with the standard roots always relocating, would avoid that. It would also give a way to split a versioned root without touching `usr/share/man`. Purely additive `paths` cannot express that today.

pkg/build/pipelines/split/manpages.yaml
```diff
@@ -21,14 +45,34 @@ pipeline:
...
+        else
+          destdir="${{targets.contextdir}}/usr/share/man"
         fi
+
+        # Merge into the destination rather than moving wholesale. Two
+        # roots can each contain a man1/, and `mv src/man1 dest/` fails
+        # with "File exists" once dest/man1 is there -- so a package
+        # shipping pages in both usr/share/man and usr/local/share/man
+        # could not be split at all.
+        mkdir -p "$destdir"
```
dannf: An empty man root now produces a subpackage containing only directory entries. `[ -d ]` passes, `mkdir -p` creates the destination, and `cp -a dir/.` succeeds with nothing to copy. The old `mv "$mandir"/*` failed on the unmatched glob under `set -e`. That failure surfaced the broken install step.

melange does emit such an apk. `pkg/tarball/write.go` writes `TypeDir` headers for every directory, and `build.go` calls `Emit` for every subpackage unconditionally. So if `make install` creates `usr/man/` and installs nothing, the result is a `foo-doc` apk containing `usr/`, `usr/share/`, `usr/share/man/` and no files.

A non-empty check before the copy, e.g. `[ -n "$(ls -A "$PACKAGE_DIR/$mandir")" ]`, would keep the old failure visible.

pkg/build/pipelines/split/manpages.yaml
```diff
@@ -21,14 +45,34 @@ pipeline:
...
+        cp -a "$PACKAGE_DIR/$mandir"/. "$destdir"/
```
dannf: In relocate mode this silently overwrites a same-named page from an earlier root, and busybox `cp` writes through existing destination inodes. Reproduced with busybox 1.37:

- `usr/share/man/man1/foo.1` (A) then `usr/local/share/man/man1/foo.1` (B): the result is B. A is gone. Exit code is 0. The old `mv` failed loudly here, with `Directory not empty` on `man1/`.
- Destination `a.1` is hardlinked to `alias.1` (gzip'd alias pages are often stored this way). The source has a different `a.1`. After the copy, `alias.1` content changes too, because both names still share one inode.
- Destination `a.1 -> b.1` is a symlink alias. The source has a regular `a.1`. After the copy, `b.1` content is replaced and `a.1` is still a symlink to it. The old `mv` replaced the symlink entry and left `b.1` alone.

The hardlink and symlink corruption are new. `mv` never opened destination inodes. A per-entry merge that checks `[ -e "$dest/$page" ]` and then fails, or at least warns, would keep the collision visible.

dannf: Separately from the overwrite question: `cp -a` + `rm -rf` turns a rename into a full byte copy plus a second tree walk. `targets.destdir`, `targets.outdir` and `targets.contextdir` are all under `/home/build/melange-out` on one mount, so the old `mv` was a single `rename(2)`, not a copy.

The copy also runs in two cases that do not need a merge. One is the `preserve-path` branch, where the destination is always fresh. The other is the first root in relocate mode, where `mkdir -p` just created the destination.

A rename-based merge keeps the semantics. Use `mv` for the whole root when `$destdir` does not exist yet. Otherwise merge per entry. A one-level loop is not enough, since locale subtrees (`fr/man1`) sit two levels down. A small recursive function using `local` works under busybox sh and dash. That merge still overwrites same-named files, so it pairs with the collision check from the other comment on this line.

pkg/build/pipelines/split/manpages.yaml
```diff
@@ -21,14 +45,34 @@ pipeline:
...
+        rm -rf "$PACKAGE_DIR/$mandir"
```
dannf: When the root's final path component is a symlink to a directory, `[ -d ]` follows it. `cp -a dir/.` then copies the target's pages into the subpackage, and `rm -rf` removes only the link. The real pages stay in the parent package.

Reproduced: `usr/man -> lib/foo/man` (target not in the list) leaves `usr/lib/foo/man/man1/a.1` in the parent and also ships `usr/share/man/man1/a.1` in `-doc`. Exit code is 0. `usr/share/man -> ../man` with `preserve-path: "true"` materialises the pages twice inside `-doc`. The old code aborted at `rmdir` on the symlink, so a build error became silent duplication.

A `[ -L "$PACKAGE_DIR/$mandir" ]` check before the copy (skip, resolve to the target, or fail with a message) would cover it.

pkg/build/pipelines/split/manpages.yaml
```diff
@@ -21,14 +45,34 @@ pipeline:
...
+        rmdir --parents --ignore-fail-on-non-empty "$(dirname "$PACKAGE_DIR/$mandir")" 2>/dev/null || true
```
dannf: The `2>/dev/null || true` looks redundant. `--ignore-fail-on-non-empty` already covers the expected failure. `dirname` strips trailing slashes, so the argument always names a parent that still exists (checked with `paths: opt/x/share/man/` under busybox). The walk up through `$PACKAGE_DIR` when the package had only man pages is unchanged from the old line.

The suppression does change one thing. A real error no longer fails the step under `set -e`. Two examples: permission denied on a parent directory, or an odd entry like `usr/share/man/.`, where dirname resolves to the directory that was just removed. The old line and `split/alldocs.yaml` run the same `rmdir` without it.

e2e-tests/split-manpages-build-test.yaml
```diff
@@ -0,0 +1,115 @@
+# Test split/manpages in its three modes: the default relocating
+# behaviour, an extra non-standard root via `paths`, and `preserve-path`
+# keeping pages where they were installed.
+#
+# preserve-path exists for software that ships several coexisting versions.
+# Every llvm-N package ships a page called scan-build.1 and openjdk-17 and
+# openjdk-21 share 27 page names, so relocating those into usr/share/man
+# makes their doc subpackages collide rather than install side by side.
+#
+# Note the assertions use `if [ ... ]; then ... fi` rather than
```
dannf: Three notes on the test:

1. The premise here is not right. `pkg/build/pipeline.go` wraps each `runs:` fragment as `set -e -o pipefail ... <fragment> exit 0`. Under that wrapper a failing test in an `&&` list is not fatal, because `set -e` ignores non-final elements of an AND list, and the trailing `exit 0` covers the status of the last command. `e2e-tests/capabilities-add-drop-build.yaml` already uses `grep ... && { ...; exit 0; }` this way. `test -f X || { echo ...; exit 1; }` one-liners, as in `split-lib-build-test.yaml`, would replace the nine four-line blocks and make this comment unnecessary.
2. The `-build-test` suffix makes `run-tests` do `melange build && melange test`. There are no `test:` blocks here, so the test op runs nothing. `split-lib-build-test.yaml` asserts via `test:` against the installed apk. Either add `test:` blocks or name this `-build.yaml` like the other build-only checks.
3. None of the new edge cases are covered: a multi-line `paths` value, a same-named page in two roots, `preserve-path: "true"` while a standard root is still present, a symlinked root, and the `case` rejection of a bad `preserve-path` value. A multi-line `paths` value would have caught the syntax error. The preserve-path subpackage only runs after the first subpackage has consumed all three standard roots, so that interaction is never exercised.

pkg/build/pipelines/split/manpages.yaml
```diff
@@ -21,14 +38,155 @@ pipeline:
...
+      paths="${{inputs.paths}}"
+
+      # Pathname expansion off while splitting `paths` into words. A glob is
+      # not resolvable here anyway -- it would expand against the build
+      # directory rather than the package root, and then silently match
+      # nothing -- so keep the entry literal and reject it by name below.
+      set -f
+      extra=""
+      for root in $paths; do
+        case "$root" in
```
dannf, line 153: The new checks catch absolute entries, `..`, and globs. They do not catch `.` (or `./`).

With `preserve-path: "true"`, the destination for `.` is `${{targets.contextdir}}/.`. This path exists because melange creates the subpackage directory before any pipeline runs (build.go:684). So `split_root` takes the `merge_tree` branch and moves every file in the parent package into the `-doc` subpackage. The script does exit 1, but only because the trailing `rmdir` on `foo/.` returns EINVAL, after the move has happened.

Reproduced under busybox 1.37 with the compiled script and `paths: "."`, `preserve-path: "true"`. The parent ends with zero files. `foo-doc/usr/bin/foo` and `foo-doc/usr/lib/x/libx.so` appear. The output is `rmdir: '.../foo/.': Invalid argument`.

Relocate mode is safe. `mv foo/. .../usr/share/man` fails with EBUSY before anything moves.

Adding `.|./*|*/.|*/./*` to the `case` would close the gap. A component-precise pattern for `..` (`..|../*|*/..|*/../*`) would also stop rejecting legitimate names like `opt/foo..bar/man`.

pkg/build/split_manpages_pipeline_test.go
```diff
@@ -0,0 +1,588 @@
...
+	pkgDir = filepath.Join(outDir, "foo")
+	ctxDir = filepath.Join(outDir, "foo-doc")
+	for _, d := range []string{pkgDir, ctxDir} {
+		if err := os.MkdirAll(d, 0o755); err != nil {
+			t.Fatalf("MkdirAll(%s): %v", d, err)
+		}
+	}
+	// The subpackage directory is created by melange before the pipeline
+	// runs, but an empty one must not look like an existing destination.
+	if err := os.Remove(ctxDir); err != nil {
```
dannf: The comment says an empty subpackage directory "must not look like an existing destination". But `$dest` is always `contextdir/usr/share/man` or `contextdir/$root`, never `contextdir` itself. So an empty `contextdir` is not a destination in any case the script handles.

Melange creates `contextdir` before pipelines run (build.go:684). Removing it here makes the fixture differ from production.

That difference matters for the `paths: "."` case on the yaml side. In production, `contextdir` exists, `split_root` takes the `merge_tree` branch, and the parent package ends up empty. With this fixture, `contextdir` is absent, `split_root` takes the wholesale `mv` branch instead, and `mv foo/.` fails with EBUSY before anything moves. A test for that case written against this fixture would pass for the wrong reason. Dropping the `os.Remove` and its comment keeps the fixture matching what melange does.

dannf (commented):
Follow-up on the rework. The earlier items are addressed; two new ones below, both around the `paths` validation and its test fixture.

pkg/build/pipelines/split/manpages.yaml
dustinkirkland: Fixed as you suggested. `paths="${{inputs.paths}}"` is assigned before the loop, so a value with newlines is a string rather than script text, and the description now says newline-separated to match `split/lib`.

Word splitting also happens under `set -f` now, for the glob problem in your next comment.

Covered by `TestSplitManpagesSucceeds/newline-separated_paths_are_both_split`, which compiles the real yaml so the splice itself is under test. The e2e preserve-path subpackage also passes `paths` as a block scalar, so the trailing newline is exercised in a real build.

pkg/build/pipelines/split/manpages.yaml
dustinkirkland: All three confirmed and all three now rejected by name.

`..` was the worst of the set. Reproduced your case exactly: `melange-out/bar` deleted, its non-man files copied into `foo-doc/usr/share/man/`, exit 0.

Entries that are absolute, contain `..`, or contain a glob are now errors. A glob additionally stays literal because the split happens under `set -f`, so the message names the pattern the packager wrote rather than reporting a missing path.

Your `|| continue` distinction is what the code does now: the three standard roots skip when absent, an explicitly named path that does not exist is an error.

Tests: `a_paths_entry_containing_..`, `an_absolute_paths_entry`, `a_glob_in_a_paths_entry`, `a_paths_entry_that_does_not_exist`. Every failure case also asserts that a sibling package directory still exists afterwards.

pkg/build/pipelines/split/manpages.yaml
dustinkirkland: Adopted, and I think your scoping is the better design rather than just the safer one. `preserve-path` now applies only to the directories named in `paths`; the standard roots always relocate.

Confirmed `usrlocal` at `pkg/linter/linter.go:107`, `defaultBehavior: Warn`, and the explain string you quoted. The `usr/man` half matters too, since it is outside man(1)'s default search path.

The point that decided it is your last one: additive `paths` could not express splitting a versioned root without also taking `usr/share/man`, and now it can.

`TestSplitManpagesSucceeds/preserve-path_keeps_paths_entries_and_still_relocates_standard_roots` covers it, and so does the e2e subpackage, whose apk contains `usr/lib/llvm-99/share/man/man1/versioned.1` alongside `usr/share/man/man1/{standard,local}.1` and `usr/share/man/man8/usrman.8`, with no page left under `usr/local` or `usr/man`.

pkg/build/pipelines/split/manpages.yaml
dustinkirkland: Confirmed and restored as a failure. Measured both: the old `mv "$mandir"/*` exits 1 with `mv: cannot stat '.../*'`, mine exited 0 having produced 0 files and 4 directories.

`split_root` now rejects a root that exists and contains no files, with a message pointing at the install step. No currently-green package can be hitting this, since the old code failed on it too.

Test: `TestSplitManpagesFailsLoudly/a_man_root_that_exists_but_is_empty`.

pkg/build/pipelines/split/manpages.yaml
dustinkirkland: All three reproduced. One refinement: the hard-linked case is busybox-specific. GNU `cp -a` breaks the link (`a.1=REPLACED alias.1=ORIGINAL`), busybox `cp -a` writes through the shared inode (`alias.1=REPLACED`). The symlink case reproduces with both. Since this pipeline declares `needs: busybox`, your result is the one that matters, but it is implementation-dependent rather than universal.

Fixed structurally rather than by checking: there is no copy any more, every move is a `rename(2)`, so a destination inode is never opened. The collision itself is now an error, which is the distinction that makes the original two-root bug fixable without hiding this one -- directories merge, files collide.

Re-ran your two cases against the new code. Both exit 1 with `a.1 is present in more than one man root`, and `alias.1` is still `ORIGINAL`, `b.1` still `BTARGET`. A non-colliding merge into a directory holding a symlink alias leaves the symlink a symlink.

Tests: `the_same_page_in_two_roots`, `a_colliding_page_that_is_a_hard-linked_alias`, `a_colliding_page_that_is_a_symlinked_alias`, and `alias_pages_survive_a_merge_that_does_not_collide`.

dustinkirkland: Adopted, though for the correctness reason rather than the performance one. A man tree is small enough that the extra walk would not matter on its own; what makes rename the right call is that it never opens a destination inode, so the alias corruption in your other comment on this line becomes impossible instead of guarded against.

Both paths you identified are there. `mv` moves the whole root in one rename when `$destdir` does not exist, which covers every `preserve-path` split and the first root in relocate mode. `merge_tree` runs only when there is something to merge with.

`merge_tree` recurses, as you said it had to -- a one-level loop misses `fr/man1`. It uses `local`, checked under both busybox sh and dash. Covered by `locale_subtrees_two_levels_down_are_merged`, and the e2e now installs a `fr/man1` in two roots so the recursive merge appears in a real apk.

pkg/build/pipelines/split/manpages.yaml
dustinkirkland: Reproduced exactly as described: pages left in the parent at `usr/lib/foo/man/man1/a.1` and shipped again in `-doc`, exit 0. Now an error.

One subtlety worth recording: the `[ -L ]` check sits *after* `[ -d ]`, not before. A compat `usr/man -> share/man` is dangling by the time the loop reaches it, because `usr/share/man` was already moved out -- so `-d` is false and it is skipped, exactly as before. Checking `-L` first would have failed every build that ships such a symlink, which is a regression the old code did not have.

Tests: `a_man_root_that_is_a_symlink_to_a_directory`, which also asserts the target's pages are still in the parent, and `dangling_compat_symlink_root_is_skipped` for the case that must keep working.

pkg/build/pipelines/split/manpages.yaml
dustinkirkland: Removed. You are right on both counts: `--ignore-fail-on-non-empty` already covers the expected failure, and `dirname` always names a parent that still exists, so the suppression bought nothing and cost the real error classes you list.

The `rmdir` now matches the old line and `split/alldocs`.

e2e-tests/split-manpages-build-test.yaml
dustinkirkland: All three correct, including the premise -- thanks for catching that.

**1.** My note was wrong. Verified under busybox sh: with melange's wrapper (`set -e -o pipefail ... <fragment> ... exit 0`) a false test in an AND list is not fatal, and the trailing `exit 0` covers the final statement too. What I actually hit was that compound being the last statement in my own harness, which has no `exit 0` -- I generalised a local artifact into a claim about melange and then wrote it into the file. The note is gone and the nine blocks are `test ... || exit 1` one-liners, as in `split-lib-build-test.yaml`.

**2.** Added `test:` blocks to all three subpackages and to the parent, so assertions now run against the installed apk instead of the staging directory. Kept the `-build-test` name since the test op now does something.

**3.** Each subpackage builds its own fixture under `targets.outdir` and passes `package:`, following `split-lib-build-test.yaml`. That is what makes your third point fixable: the preserve-path subpackage now installs a versioned root *and* all three standard roots, so the interaction is exercised for the first time -- previously, as you say, the first subpackage had consumed all three. `paths` is given as a block scalar there, so the newline is covered in a real build.

The cases that must fail the build -- collision, empty root, symlinked root, bad `preserve-path`, and the `paths` validation -- are in `TestSplitManpagesFailsLoudly`, since this harness has no expected-failure convention.

One thing I got wrong on the way: directory-absence assertions are unusable here. `/usr/man`, `/usr/local` and `/opt` all ship in the base test environment, so the assertions name individual pages instead.

dustinkirkland:
Thanks @dannf -- this was a genuinely useful review. Every one of the nine findings reproduced, and the one refinement I found is noted inline (the hard-linked alias corruption is busybox-specific; GNU `cp -a` breaks the link instead).

The unifying problem was worth stating plainly: every crudity in the `mv "$mandir"/*` loop I replaced was fail-loud, and I converted five of them into silent wrong output. The PR claimed to fix merging two man roots, but what it actually did was stop the failure happening rather than make the merge correct. The distinction that resolves it is **directories merge, files collide** -- two roots each carrying a `man1/` is the case this pipeline needs to support, the same page in both roots is a packaging bug.

Addressed in two commits, `aa5fb10` and `0262fbc`, with all 101 checks passing. All nine inline threads have replies and are resolved.

### The two items with no line to attach to

**`split/alldocs`.** Ported, and the same helpers are now shared with `split/manpages`. Its `usr/share/doc` / `usr/local/share/doc` loop has the identical bug, which is why I did not take the `uses: split/manpages` option you offered -- nesting would have fixed the man loop and left the doc loop broken. Both loops now go through `split_root`. Covered by `TestSplitAllDocsMergesRoots`, including `two_doc_roots_sharing_a_package_directory_are_merged` and the doc-file collision. Kept the two copies of the helpers with a comment on each saying to keep them in step; if you would rather see the duplication removed, I am happy to look at what it would take for one pipeline to source shell from another, but that felt out of scope here.

**The `paths` name and format.** You are right that the name is consistent within `split/` and the format was not. `split/lib` documents newline separation for its `paths` and reads it with `while read -r`; mine said "whitespace separated" and spliced the value straight into a `for` line. The description now says newline-separated, the value is assigned to a variable first, and word splitting happens under `set -f` so a glob cannot expand against the build directory. Kept the `paths` name for the consistency you point out.

### On testing

The failure cases could not go in `e2e-tests/`, since `run-tests` has no expected-failure convention -- every yaml there has to build cleanly, and half of what this pipeline now does is fail loudly. They are in `pkg/build/split_manpages_pipeline_test.go`, which compiles the real pipeline yaml through `compilePipeline` and runs the script, following the precedent in `patch_pipeline_test.go`. That matters for the newline bug specifically: it lived in the substitution rather than in a value, so a test that only exercised the shell would have missed it.

All 24 subtests fail against the version of the pipeline that had the corresponding bug -- 14 against this PR's previous head, and a different 21 against `upstream/main`, which is the check that the merge behaviour itself is pinned.

The e2e test is reworked per your third note and now asserts against installed apks rather than the staging directory.

Re-requesting your review, and thanks again for reproducing each case -- it made every one of these unambiguous to act on.

pkg/build/pipelines/split/manpages.yaml
dustinkirkland: Confirmed and fixed in `18d764c`. Reproduced exactly what you describe: with `paths: "."` and `preserve-path: "true"` the parent ends with zero files, `foo-doc/usr/bin/foo` and `foo-doc/usr/lib/x/libx.so` appear, and the only reason the build stops is `rmdir: '.../foo/.': Invalid argument` after the move. Relocate mode is safe for the reason you give.

Both patterns are now component-precise, as you suggested:

```sh
..|../*|*/..|*/../*)
.|./*|*/.|*/./*)
```

Checked the second point too: `*..*` was rejecting `opt/foo..bar/man`, and the component form accepts it while still catching `..`, `../x`, `a/b/..` and `usr/../../etc`.

Tests: `a_paths_entry_of_._under_preserve-path`, which also asserts the parent keeps `usr/bin/foo` and `usr/lib/x/libx.so`; `a_paths_entry_with_a_leading_./`; `a_paths_entry_with_a_trailing_/.`; and `a_paths_entry_whose_directory_name_contains_two_dots` for the false rejection. All four fail with the new checks removed.

pkg/build/split_manpages_pipeline_test.go
dustinkirkland: You are right that the premise was wrong -- `$dest` is always a path beneath `contextdir`, never `contextdir` itself, except for the `"."` entry in your other comment, which is exactly the case that needed a truthful fixture. Dropped the `os.Remove` and rewrote the comment to say why it stays.

One thing I could not reproduce: on GNU coreutils the `"."` case empties the package on *both* fixtures, because GNU `mv` moves the contents rather than refusing outright, so the wholesale branch is not protective there either. So the specific "passes for the wrong reason" divergence looks busybox-`mv`-specific. It does not change the conclusion -- the fixture should match what melange does, and the test is only meaningful if it exercises the branch production takes -- so the change is in as you asked.

The `"."` test now fails with `parent package lost "usr/bin/foo"; has []` when the validation is removed, which is the failure mode you described.

dustinkirkland:
Two more findings from @dannf addressed in `18d764c`, both reproduced first:

- **A `paths` entry of `.` bypassed the validation.** Under `preserve-path` its destination is the subpackage directory, which melange creates before any pipeline runs, so the merge branch moved *every* file in the package into the `-doc` subpackage. The build stopped only on the trailing `rmdir` returning EINVAL on `.`, well after the move. Both dot patterns are now matched per component. The same change fixes a false rejection: `*..*` was refusing legitimate names like `opt/foo..bar/man`.
- **The test fixture was removing the subpackage directory** on a premise that does not hold, which made the `.` case take a branch production does not take. Dropped, so the fixture matches melange.

101/101 checks pass and all 11 review threads are resolved. Four new subtests, 28 in total, and all four of the new ones fail with the corresponding check removed.

dannf (approved):
All findings from both reviews are addressed. Verified on 18d764cf: the split Go tests pass, the e2e test passes under the qemu runner, and the component-precise path checks behave as intended under busybox sh.

PR #2661, opened 2026-09-18
fix(git-checkout): retry the tag fetch, not just the clone

toabctl:
## Problem

The `tag` path in the `git-checkout` pipeline performs two network operations:

1. the `git clone`, wrapped in `retry_with_backoff`
2. a subsequent bare `git fetch` of the tag refspec — **not** wrapped

Because the script runs under `set -e`, a single transient failure on step 2 killed the build immediately, regardless of how high `max-retries` was set. A build with `max-retries=8` would retry the clone eight times, succeed, and then die on the very next request.

```
[git checkout] Retry attempt 1/8 after 60s backoff...
[git checkout] cloned from https://gitlab.com/gitlab-org/gitlab.git
[git checkout] execute: git fetch --quiet origin --depth=1 --no-tags +refs/tags/v19.1.7-ee:...
fatal: remote error: GitLab is currently unable to handle this request due to load (ID a3cebbcd8806c8ee-ORD)
```

## Why this is worth fixing

This turned out to be the dominant failure mode, not an edge case. Over a 30-day window of build logs for `gitlab.com/gitlab-org/gitlab`, 223 `git-checkout` runs hit upstream HTTP 503 load-shedding:

| Died on | Runs |
|---|---|
| **tag fetch (unretried)** | **73** |
| clone (retries exhausted) | 7 |
| post-clone | 5 |

194 runs hit the error at least once, 109 recovered, 85 died — and **73 of those 85 deaths (86%) were on this one unretried line**. The clone retry was doing its job: 164 runs hit the same error during clone and only 7 died there.

## Change

Wrap the tag fetch in `retry_with_backoff` using the same parameters as the clone, and `fail` with an explicit message when retries are exhausted, instead of falling through to `git checkout` of a ref that was never fetched.

`vr` is passed *through* `retry_with_backoff` rather than around it, so every attempt still logs its `execute:` line and `retry_with_backoff` itself needed no change.

## Testing

Extracted the rendered script and ran it under `busybox sh` (and syntax-checked under `dash`) against a local repo with an annotated tag, using a `git` shim to inject failures:

- happy path — exit 0, log output identical to before the change
- fetch fails twice, then succeeds — `Retry attempt 1/4`, `2/4`, then completes, exit 0
- retries exhausted — `FAIL git fetch of tag v1.0 failed after 1 retries`, exit 1, and it does **not** fall through to `git checkout`

`shellcheck` passes via pre-commit.

## Note

The `git fetch` inside `process_cherry_picks` is the same class of gap and is still unretried. Left out to keep this change minimal; happy to follow up.


Taffer (approved):
LGTM, I wonder if I can approve PRs in this repo... 🤔

